PR #70, opened 2017-04-20
Record phone numbers to the database with each transaction

zeppelin:


joshmh:
Thanks, @zeppelin. We're including this in our upcoming version here:

https://github.com/lamassu/lamassu-server/blob/tx-based/migrations/011-transactions-reload-2.js#L39

zeppelin:
@joshmh Yes, however we really need this feature and won't be able to upgrade to the new version easily, even if it was introduced today.

If you're willing to merge: do you want me to update the column type from `varchar` to `text` to match your version?

naconner:
@zeppelin Is there something that prevents deploying these changes on your current server instance?

debreczeni:
@naconner  https://youtu.be/3_eR0IVSOhY

debreczeni:
@joshmh Do you have a realistic ETA for the new version? This branch seems to add too many new features in one go, and it also has been diverged from master more than a year ago... Which doesn't make for a straightforward upgrade.
Also we can't see any tests for this piece of software. Software that is handling cash and programmable money... A software that already has so many bugs... Bugs that are not being fixed, but instead new features being introduced on a completely off branch. Or in fact in a big git repository with unclear versioning and branching structure. I don't mean to be rude but to be brutally honest we can't see this project going anywhere good and assuring :( 
So that is why we keep adding quick fixes - but most importantly to reduce our support time costs. Because the lamassu system keeps our support and management staff a lot more busy then it should. The reason we still have the hardware is because it is the only BTM in our country, but we loose faith every day. From what there is not much more left...
If you guys don't feel to merge this it's OK. We can keep adding quick fixes to eliminate the most painful points and spend time on projects that has a brighter future...

manchfu:
@debreczeni The new version is alpha testing, you can take it for a spin if you'd like. We are still maintaining the old version for critical bugs according to their priority. We are not adding features to the old version at this point, as that would delay the release of our new version.

zeppelin:
> We are not adding features to the old version at this point

@manchfu Shall I close the PR then?

PR #75, opened 2017-08-14
V5

goga-m:


migrations/036-add_customers_table.js
```diff
@@ -0,0 +1,32 @@
+var db = require('./db')
+
+exports.up = function(next) {
+  const sql =
+    [
+    'CREATE TABLE customers ( ' +
```
joshmh: Use ES2015 template literals here and in compliance_authorizations: https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Template_literals

Example: https://github.com/lamassu/lamassu-server/blob/v5/migrations/035-log_bank_notes.js

joshmh: Also, our practice is to not capitalize anything, all lowercase. But that's not a big deal.

migrations/036-add_customers_table.js
```diff
@@ -0,0 +1,32 @@
...
+    'id uuid PRIMARY KEY,' +
+    'phone text unique NOT NULL,' +
+    'phone_at timestamptz NOT NULL,' +
```
joshmh: I realized that since cash_in_txs has a foreign key to this table, we always need to link to a record in this table. Since it's possible to have a cash_in_tx where we don't know the user's phone number, we should make these fields optional.

joshmh: We might consider creating an "anonymous" record for every transaction that's recorded where we don't have any info. In some cases that will be most or all of the txs, so no point having a blank record here for every transaction.

PR #85, opened 2017-10-30
Update Zcash and Litecoin

CrypticaScriptura:
Update Litecoin, Zcash, and add `-disabledeprecation=1.0.12` flag to zcashd command.


naconner:
@joshmh Removed sed and added `disabledeprecation` flag to zcash.js. Looks good?

PR #90, opened 2018-01-04
Add Coinbase BCH ticker

naconner:


naconner:
@joshmh Shall we merge?

PR #91, opened 2018-01-26
Update node wallet versions

naconner:
This version of Bitcoin ABC includes CashAddr support (default).


naconner:
Wait until revamped BCH support to merge.

PR #92, opened 2018-02-01
Update lamassu-update, add to bin

naconner:
Without --allow-root, when running the terminal streams an endless:

```
gyp WARN EACCES user "root" does not have permission to access the dev dir...
```


joshmh:
Shouldn't the flag be ``--unsafe-perm``?

naconner:
`--allow-root` avoided it for me, though both are reported to work: https://github.com/nodejs/node-gyp/issues/454

PR #93, opened 2018-02-01
Trade on USD Kraken if fiat is other than USD/EUR

naconner:
If a machine currency is other than USD or EUR, Kraken trades will fail because pair is not defined in `common/kraken.js` (and not offered by Kraken).

Instead, trade on USD market.

### DIFF
--- a/lib/plugins/exchange/kraken/kraken.js
+++ b/lib/plugins/exchange/kraken/kraken.js
@@ -19,7 +19,11 @@ function trade (account, type, cryptoAtoms, fiatCode, cryptoCode) {
   const kraken = new Kraken(account.apiKey, account.privateKey, {timeout: 30000})
   const amount = common.toUnit(cryptoAtoms, cryptoCode)
   const amountStr = amount.toFixed(6)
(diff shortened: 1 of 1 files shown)


joshmh:
Have you talked to operators to see if this is their desired behavior? How do we know if they want EUR or USD?

naconner:
Unless we create an admin-based selector for USD vs EUR market, this brings it in line with expectations we can set in knowledgebase documentation.  The USD market has greater depth, and Kraken is a U.S.-based exchange. (Our Bitstamp trader had been USD-market only, even though EUR pairs existed, and we set this expectation without issue.)

joshmh:
Kraken has much higher EUR liquidity than USD. Should we still choose USD for Kraken?

naconner:
Ah, my bad. Using EUR makes sense then, also since more currencies exist around the Eurozone than the U.S., and they may have SEPA transfers and better rates for EUR conversion.

naconner:
@joshmh Now defaulting to EUR. All set?

PR #96, opened 2018-02-05
Ticker file for Indian exchange Coinome

tarunpall:
Indian Bitcoin prices are usually 10-20% higher than those outside India.  www.coinome.com is a popular Indian exchange, therefore, adding their API.


naconner:
Thanks @tarunpall! Safe to assume this would be useful only if the server is set to INR, correct?

We'll be looking at ways to communicate that certain tickers support only specific currencies, thus better to save this until then if Coinome is INR-only.

PR #110, opened 2018-04-05
Update ZEC & BCH for hardforks, ETH release

naconner:
ZEC will hardfork upgrade on 25 June. v1.1.0 is compatible with new consensus changes.

BCH will hardfork upgrade on 15 May. v0.17.0 is compatible with new consensus changes.

ETH update to address network DoS attacks.


naconner:
@joshmh Done.

PR #111, opened 2018-04-05
Use forex conversion on Bitstamp

naconner:
Employs the method we use for Kraken when machine currency is other than EUR or USD.


joshmh:
Let's do the PR to staging.

naconner:
@joshmh Done.

PR #118, opened 2018-05-29
Terms and conditions screen

lidio601:


joshmh:
lamassu-config.json looks good.

Please add a migration file for the database change. See the latest files in the migrations directory for examples. Then you can run bin/lamassu-migrate to run and test.

migrations/1527814550220-add-tx-terms-accepted.js
```diff
@@ -0,0 +1,14 @@
+var db = require('./db')
+
+exports.up = function (next) {
+  const sql = [
+    'alter table cash_in_txs add column terms_accepted boolean default null',
+    'alter table cash_out_txs add column terms_accepted boolean default null'
```
joshmh: These should be **not null** and default to false.

lib/routes.js
```diff
@@ -62,6 +63,17 @@ function poll (req, res, next) {
         }
       }
 
+      let terms = false;
```
joshmh: We can do this without a mutable variable. Make a function that creates the **terms** object, and do:

```
const terms = config.termsScreenActive ? createTerms(config) : null
```

PR #121, opened 2018-06-08
Expired transaction after bugfix

lidio601:


install
```diff
@@ -160,7 +160,10 @@ cat <<EOF > $CONFIG_DIR/lamassu.json
   "lamassuServerPath": "$NODE_MODULES/lamassu-server",
   "migrateStatePath": "$MIGRATE_STATE_PATH",
   "blockchainDir": "$BLOCKCHAIN_DIR",
-  "ofacDataDir": "$OFAC_DATA_DIR"
+  "ofacDataDir": "$OFAC_DATA_DIR",
+  "coinAtmRadar": {
+    "url": null
```
joshmh: Might as well fill this in with the real coinAtmRadar URL.

joshmh: https://coinatmradar.info/api/lamassu/

lib/coinatmradar/coinatmradar.js
```diff
@@ -103,8 +103,14 @@ function getMachines (info) {
...
 function sendRadar (data) {
+  const url = _.get(options, ['coinAtmRadar', 'url']);
+
+  if (!url) {
+    return Promise.reject(new Error('options.coinAtmRadar.url is null!'));
+  }
+
   const config = {
-    url: options.coinAtmRadar.url,
+    url: url,
```
joshmh: Change to: `url,` instead of `url: url,`

lib/route-helpers.js
```diff
@@ -49,7 +49,8 @@ function toCashOutTx (row) {
 function fetchPhoneTx (phone) {
   const sql = `select * from cash_out_txs
     where phone=$1 and dispense=$2
-    and (extract(epoch from (coalesce(confirmed_at, now()) - created))) * 1000 < $3`
+    and (extract(epoch from (now() - created))) * 1000 < $3
+    and confirmed_at is not null`
```
joshmh: Looks like you found the bug, but I don't think it should be testing `confirmed_at is not null` because I think we need to pull these records even before there's a confirmation.

lib/wallet.js
```diff
@@ -138,6 +138,7 @@ function authorizeZeroConf (settings, tx, machineId) {
   const plugin = cryptoConfig.zeroConf
   const zeroConfLimit = machineConfig.zeroConfLimit
 
+  // what if tx.fiat === undefined?
```
joshmh: I'd rather not have this kind of thing in comments (even thought there are some of these in the existing code). Better to specifically handle it.

lidio601: Yes,
sorry I meant to add a comment here.
What should be the expected behavior here in that case?

joshmh: I believe tx.fiat should never been `undefined` at this point, so you can throw an error if that happens.

lidio601:
This should be ready

lib/coinatmradar/coinatmradar.js
```diff
@@ -103,8 +103,14 @@ function getMachines (info) {
 }
 
 function sendRadar (data) {
+  const url = _.get(options, ['coinAtmRadar', 'url']);
```
joshmh: I'm really concerned about this. Lodash arguments are reversed. This means you didn't test this code, so if I miss this in code review we're going to spend a lot of time debugging, or it could even lead to a bug in production. We need to test all changes and added features, at least with a unit test.

The trick with lodash/fp is that the last argument is almost always the object we want to work on. In this case we want to pull a field out of options, so options would be the last argument. This is handy for fp stuff because we can do:

```
then(_.get('coinAtmRadar.url'))
```

instead of:

```
then(options => _.get('coinAtmRadar.url', options)
```

lidio601: True I didn't tested it and I'm still trying to get the difference between lodash and lodash/fp
Ok, thanks for the clarification.
I just came up with that exception once I was missing the option in my lamassu.json

joshmh: Yeah, it's really confusing, I've made those mistakes many times. That makes it even more important to test and try to use types. We should talk about this more.

PR #122, opened 2018-06-08
Strike bugfix

lidio601:
https://github.com/lodash/lodash/commit/c6f8896826a82d894e0335e78e42d8d8f80bb87a

### DIFF
--- a/install
+++ b/install
@@ -160,7 +160,13 @@ cat <<EOF > $CONFIG_DIR/lamassu.json
   "lamassuServerPath": "$NODE_MODULES/lamassu-server",
   "migrateStatePath": "$MIGRATE_STATE_PATH",
   "blockchainDir": "$BLOCKCHAIN_DIR",
-  "ofacDataDir": "$OFAC_DATA_DIR"
+  "ofacDataDir": "$OFAC_DATA_DIR",
+  "strike": {
+    "baseUrl": "https://api.strike.acinq.co/api/"
+  },
+  "coinAtmRadar": {
+    "url": "https://coinatmradar.info/api/lamassu/"
+  }
 }
 EOF
 
--- a/lib/plugins/layer2/strike/strike.js
+++ b/lib/plugins/layer2/strike/strike.js
@@ -9,7 +9,10 @@ module.exports = {
   cryptoNetwork
 }
 
-axios.defaults.baseURL = options.strike.baseUrl
+axios.defaults.baseURL = _.get('strike.baseUrl', options)
+if (_.isEmpty(axios.defaults.baseURL)) {
+  throw new Error('Missing Strike baseUrl!')
+}
 
 function cryptoNetwork (account, cryptoCode) {
   return Promise.resolve('test')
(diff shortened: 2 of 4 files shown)


install
```diff
@@ -160,7 +160,10 @@ cat <<EOF > $CONFIG_DIR/lamassu.json
   "lamassuServerPath": "$NODE_MODULES/lamassu-server",
   "migrateStatePath": "$MIGRATE_STATE_PATH",
   "blockchainDir": "$BLOCKCHAIN_DIR",
-  "ofacDataDir": "$OFAC_DATA_DIR"
+  "ofacDataDir": "$OFAC_DATA_DIR",
+  "strike": {
+    "baseUrl": null
```
joshmh: Put in the default correct URL: https://api.strike.acinq.co/api/

lib/plugins/layer2/strike/strike.js
```diff
@@ -1,5 +1,5 @@
 const axios = require('axios')
-const _ = require('lodash/fp')
+const _ = require('lodash')
```
joshmh: We always use `lodash/fp` as it allows us to be more functional. It's important to keep this consistent.

lib/plugins/layer2/strike/strike.js
```diff
@@ -9,7 +9,7 @@ module.exports = {
   cryptoNetwork
 }
 
-axios.defaults.baseURL = options.strike.baseUrl
+axios.defaults.baseURL = _.toString(_.get(options, 'strike.baseUrl', ''))
```
joshmh: How does it help to default to blank baseURL? Should probably throw an error, instead.

lib/plugins/layer2/strike/strike.js
```diff
@@ -38,15 +38,18 @@ function createCharge (account, info) {
   return axios({
     method: 'post',
     url: 'v1/charges',
-    auth: {username: account.token, password: ''},
+    auth: {
+      username: _.get(account, 'token', ''),
```
joshmh: Token should never be blank. Throw error instead.

lib/plugins/layer2/strike/strike.js
```diff
@@ -38,15 +38,18 @@ function createCharge (account, info) {
...
+    },
     data
   }).then(_.get('data'))
 }
 
 function newAddress (account, info) {
   return checkCryptoCode(info.cryptoCode)
     .then(() => createCharge(account, info))
-    .then(_.tap(console.log))
+//    .then(_.tap(console.log))
```
joshmh: Just delete this line. Comments should only be descriptive text.

lidio601:
this should be ready

lib/plugins/layer2/strike/strike.js
```diff
@@ -9,7 +9,11 @@ module.exports = {
   cryptoNetwork
 }
 
-axios.defaults.baseURL = options.strike.baseUrl
+const baseUrl = _.attempt(() => options.strike.baseUrl)
```
joshmh: Wouldn't `_.get` be simpler here?

lib/plugins/layer2/strike/strike.js
```diff
@@ -34,19 +38,25 @@ function createCharge (account, info) {
     currency: 'btc',
     description: 'Lamassu cryptomat cash-out'
   }
+  const token = _.attempt(() => account.token)
```
joshmh: Use `_.get`.

lidio601:
updated

PR #130, opened 2018-06-19
Sms integration fix

lidio601:
This is to:

- Fix any configuration error that wasn't sent within the Promise context because it was crashing here https://github.com/lamassu/lamassu-server/compare/staging...lidio601:sms-integration-fix?expand=1#diff-69df051a0a40f600c077ad5f90babb52L10
- Added the change to have an error by the mock-sms integration by sending '666' as phone number 👿

### DIFF
--- a/dev/send-message.js
+++ b/dev/send-message.js
@@ -1,29 +1,30 @@
 require('es6-promise').polyfill()
 
-var config = require('../lib/config')
-var plugins = require('../lib/plugins')
+var config = require('../lib/settings-loader')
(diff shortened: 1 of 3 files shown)


joshmh:
Haha, great idea on the 666 :japanese_ogre:

joshmh:
[If bad Twilio credentials, requires l-m restart](https://trello.com/c/9277wdW9/464-if-bad-twilio-credentials-requires-l-m-restart)

PR #132, opened 2018-07-01
compliance router fix

lidio601:
lib/compliance.js has a different signature

```
function validationPatch (deviceId, config, customer) {
```


joshmh:
@lidio601 Can you add more details about this? What PR changed the signature? Why does this have to be changed now?

lidio601:
Sure,
this is the same fix I did here https://github.com/lamassu/lamassu-server/pull/120 unfortunately I didn't realized the same bug was appearing also here.

The signature changed here https://github.com/lidio601/lamassu-server/commit/99dae10547f72798c02e340877268a7c6b5052bb#diff-6fdd173c94cfbb369224848953c3921eR70

PR #137, opened 2018-07-17
OFAC flow fix

lidio601:
https://trello.com/c/xDhnNOaq/572-blocked-customer-flow-on-cash-out-invokes-cash-in-screens

### DIFF
--- a/lib/compliance.js
+++ b/lib/compliance.js
@@ -59,16 +59,17 @@ function matchOfac (deviceId, customer) {
     })
 }
 
-function validateOfac (deviceId, customer) {
-  if (customer.sanctionsOverride === 'blocked') return false
-  if (customer.sanctionsOverride === 'verified') return true
+function validateOfac (deviceId, config, customer) {
+  if (!config.sanctionsVerificationActive) return Promise.resolve(true)
+  if (customer.sanctionsOverride === 'blocked') return Promise.resolve(false)
+  if (customer.sanctionsOverride === 'verified') return Promise.resolve(true)
 
   return matchOfac(deviceId, customer)
     .then(didMatch => !didMatch)
 }
 
 function validationPatch (deviceId, config, customer) {
-  return validateOfac(deviceId, customer)
+  return validateOfac(deviceId, config, customer)
     .then(ofacValidation => {
       if (_.isNil(customer.sanctions) || customer.sanctions !== ofacValidation) {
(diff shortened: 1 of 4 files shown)


lib/compliance.js
```diff
@@ -59,16 +59,17 @@ function matchOfac (deviceId, customer) {
     })
 }
 
-function validateOfac (deviceId, customer) {
-  if (customer.sanctionsOverride === 'blocked') return false
-  if (customer.sanctionsOverride === 'verified') return true
+function validateOfac (deviceId, config, customer) {
+  if (!config.sanctionsVerificationActive) return Promise.resolve(true)
```
lidio601: Actually we need to return a Promise here otherwise at line 73 there won't be a thenable but a boolean

lib/ofac/index.js
```diff
@@ -42,13 +42,13 @@ function makeCompatible (nameParts) {
 
 function match (nameParts, birthDateString, options) {
   const {debug} = options
+
   if (!structs) {
     const message = 'The OFAC data sources have not been loaded yet.'
-    return Promise.reject(new Error(message))
+    throw new Error(message)
```
lidio601: I think it's better to raise here because before the fix of the `config.sanctionsVerificationActive`, it was behaving in a wrong way.

lidio601: the rejected promise where passed [here](https://github.com/lamassu/lamassu-server/pull/137/files#diff-6fdd173c94cfbb369224848953c3921eR57) as results variable and then was processed here https://github.com/lamassu/lamassu-server/pull/137/files#diff-6fdd173c94cfbb369224848953c3921eR21 somehow.
BTW it wasn't catched outside as an Error

joshmh:
[Blocked customer flow on cash-out invokes cash-in screens](https://trello.com/c/xDhnNOaq/572-blocked-customer-flow-on-cash-out-invokes-cash-in-screens)

lib/compliance.js
joshmh: 👍

lib/ofac/index.js
joshmh: Yeah good point, but I think we should catch errors in `validateOfac`, return `false`, and log the error using `logger.error()`.

lidio601:
@joshmh code updated

PR #139, opened 2018-08-10
config-validator to consider default values from schema

lidio601:
This is to solve https://gist.github.com/naconner/1f1e668ca69039675f581efcb602423d#file-gistfile1-txt-L73

To test you can take the last user_config and remove all the terms related configs from the array.
Then when you start l-a server you'll see the config-validator error.

This change is to make it consider also defaults from the schema in case the config value is missing (like in this case when we were migrating from an old version)


joshmh:
['No such group schema' / 'Invalid configuration' following upgrade to 7.0](https://trello.com/c/MQOLHgPP/583-no-such-group-schema-invalid-configuration-following-upgrade-to-70)

PR #151, opened 2018-08-15
Id card photo image upload

lidio601:
Relies on https://github.com/lamassu/lamassu-machine/pull/134


lib/customers.js
```diff
@@ -332,4 +338,40 @@ function batch () {
...
 
-module.exports = {add, get, batch, getById, update}
+function updatePhotoCard (id, data) {
+  let imageData = _.get('idCardPhotoData', data)
+
+  if (_.isEmpty(imageData)) {
+    return Promise.resolve(data)
+  }
+
+  data = _.omit('idCardPhotoData', data)
```
joshmh: We should have a comment that shows what ``data`` looks like. Also, ``data`` shouldn't be mutated, we should define a new variable for the filtered view of data. More functional.

lib/customers.js
```diff
@@ -332,4 +338,40 @@ function batch () {
     }, customers)))
 }
 
-module.exports = {add, get, batch, getById, update}
+function updatePhotoCard (id, data) {
+  let imageData = _.get('idCardPhotoData', data)
+
+  if (_.isEmpty(imageData)) {
+    return Promise.resolve(data)
+  }
```
joshmh: It's best to start returning a promise at the beginning of the function and have all the returns inside that block instead of using Promise.resolve.

lib/customers.js
```diff
@@ -332,4 +338,40 @@ function batch () {
...
+
+  data = _.omit('idCardPhotoData', data)
+
+  // decode the binary image
+  imageData = Buffer.from(imageData, 'base64')
```
joshmh: Need to check the performance of this. It's the simplest way to do it, which is good, but for huge images it might be a lot faster and use less bandwidth to send the image in binary.

lib/customers.js
```diff
@@ -332,4 +338,40 @@ function batch () {
...
+  // workout the image folder
+  const rpath = _.join(path.sep, _.map(_.wrap(_.join, ''), _.take(3, _.chunk(2, _.split('', hash)))))
+
+  // create the directory tree if needed
+  const dirname = path.join(idPhotoCardBasedir, rpath)
+  _.attempt(() => makeDir.sync(dirname))
+
+  // write image file
+  const filename = path.join(dirname, hash + '.jpg')
+  fs.writeFileSync(filename, imageData)
```
joshmh: Use async ``writeFile``, especially since this is a big file.

lib/routes.js
```diff
@@ -328,7 +329,7 @@ const app = express()
 const localApp = express()
 
 app.use(helmet({noCache: true}))
-app.use(bodyParser.json())
+app.use(bodyParser.json({limit: '2mb'}))
```
joshmh: See above about performance.

lib/customers.js
lidio601: ok

lib/customers.js
lidio601: done

lidio601:
Seems good to go then?
Should I add a TODO note to move from json encoded to multipart form data?

joshmh:
No need for TODO note, but you can do a trello card in priorities.

PR #156, opened 2018-09-01
layer2 default fix

lidio601:
I was condering if it is the case of giving the error anyway (in the logs).

Otherwise I can just fix with

```if (_.defaultTo('no-layer2', plugin) === 'no-layer2') return Promise.resolve()```

### DIFF
--- a/lib/layer2.js
+++ b/lib/layer2.js
@@ -1,8 +1,17 @@
 const configManager = require('./config-manager')
 const ph = require('./plugin-helper')
+const _ = require('lodash/fp')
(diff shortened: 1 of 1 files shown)


joshmh:
Oops, just saw your comment now ;)

joshmh:
[Make 'No layer 2' default, otherwise cash-out non-functional until set after update](https://trello.com/c/HyqEgVXF/593-make-no-layer-2-default-otherwise-cash-out-non-functional-until-set-after-update)

PR #157, opened 2018-09-01
More descriptive SMS alerts

naconner:
Instead of subject, which does not specify the individual machine for which there are alerts, instead use email alert text for SMS.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -31,10 +31,18 @@ Use Postgres.app: https://postgresapp.com/
 
 **psql** is automatically installed. You won't need to set up users.
 
+## NodeJS
+
+```
+curl -L https://git.io/n-install | bash -s -- -y lts
+. ~/.bash_profile
+```
+
 # Installation
 
 ## Install node modules
-Ignore any warnings.
+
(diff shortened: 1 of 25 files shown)


joshmh:
Won't this be too long for an SMS?

naconner:
Soft limit is 160 characters, this is 96 characters:

```
Errors were reported by your Lamassu Machines.

Errors for Test Machine:
Machine down for ~1 minute
```

Twilio will split anything over 160 characters into multiple messages, which shouldn't be minded given the utility of the alerts: https://www.twilio.com/docs/glossary/what-sms-character-limit

joshmh:
@naconner ping.

naconner:
Makes sense. I can try a few things. Issue I had with it in testing was that no changes to notifier.js were reflected in new SMS alerts, despite restarting l-s and the server itself. Only by installing on a fresh droplet from a branch with that change was it finally reflected.

joshmh:
@naconner Only thing that makes sense to me is that there was a prefix issue and your were running different code than you changed.

joshmh:
Closing this for now.

PR #159, opened 2018-09-05
Improve cash out monitor transaction handling

RafaelTaranto:
Made the queries from 'monitorLiveIncoming' and 'monitorStaleIncoming'
return mutually exclusive records from the db.

Both these methods are polled and perform the same actions on the
records. As a result, transaction errors were prone to happen.


joshmh:
[Database error saving dispense status.](https://trello.com/c/scY6g0en/556-database-error-saving-dispense-status)

joshmh:
Looks great, merging!

PR #163, opened 2018-09-11
Improve serialization error handling on cash_out

RafaelTaranto:
Made API calls return 204 on serialization error.

When network is unreliable there is a high chance of generating a
serialization exception. By returnig 204 on them the lamassu-machine can
just continue to work as if nothing happened.

[Trello card](https://trello.com/c/WaVHCVw1/600-db-serialization-bug-read-write-deps#)

### DIFF
--- a/lib/routes.js
+++ b/lib/routes.js
@@ -25,6 +25,7 @@ const argv = require('minimist')(process.argv.slice(2))
 
 const CLOCK_SKEW = 60 * 1000
 const REQUEST_TTL = 3 * 60 * 1000
+const SERIALIZATION_FAILURE_CODE = '40001'
 
 const pids = {}
 const reboots = {}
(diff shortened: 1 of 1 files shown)


lib/routes.js
```diff
@@ -25,6 +25,7 @@ const argv = require('minimist')(process.argv.slice(2))
 
 const CLOCK_SKEW = 60 * 1000
 const REQUEST_TTL = 3 * 60 * 1000
+const SERIALIZATION_FAILURE_CODE = '40001'
```
RafaelTaranto: This is the second time I've added this const, should I create a file with BD errors consts or should we wait to see if they happen more in the code base?

joshmh: Yeah, if you've got it in two separate files it's a good idea to break it out to a common file. Maybe put that in a new PR.

lib/routes.js
```diff
@@ -134,6 +135,7 @@ function postTx (req, res, next) {
       return res.json(tx)
     })
     .catch(err => {
+      if (err.code === SERIALIZATION_FAILURE_CODE) return res.status(204).json({})
```
joshmh: Can you add a comment explaining why it's OK to return a 204 on this error? You can put this in a new PR, since I'm merging this.

RafaelTaranto:
[DB serialization bug (read/write deps)](https://trello.com/c/WaVHCVw1/296-db-serialization-bug-read-write-deps)

PR #164, opened 2018-09-12
Add device_id to cash_out_actions table

RafaelTaranto:
If a costumer originates a cashOut transaction at one device, and then
redeem at a different device, the wrong machine would have bills debited from it.

With device_id being saved at the cash_out_actions table the bills can
be properly reconciled.

**Note**: Possibly incomplete, see more on trello card.

### DIFF
--- a/lib/cash-out/cash-out-actions.js
+++ b/lib/cash-out/cash-out-actions.js
@@ -18,7 +18,7 @@ function logActionById (t, action, _rec, txId) {
 }
 
 function logAction (t, action, _rec, tx) {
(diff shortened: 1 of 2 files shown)


RafaelTaranto:
[Trello Card](https://trello.com/c/fHwZBeE1/450-cashoutactions-deviceid-must-be-stored)

joshmh:
[cash_out_actions: device_id must be stored](https://trello.com/c/fHwZBeE1/450-cashoutactions-deviceid-must-be-stored)

PR #171, opened 2018-09-23
fix outdated symlinks after update from v5

lidio601:


lidio601:
[Move old lamassu-server directory with lamassu-update](https://trello.com/c/gzdEDWk3/616-move-old-lamassu-server-directory-with-lamassu-update)

PR #174, opened 2018-09-25
lamassu-update minor fix

lidio601:
reset the terminal when linking/unlinking new exes


lidio601:
[Move old lamassu-server directory with lamassu-update](https://trello.com/c/gzdEDWk3/616-move-old-lamassu-server-directory-with-lamassu-update)

PR #175, opened 2018-09-27
reworking l-s startup

lidio601:
it was failing to retry when starting with an invalid configuration

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -17,6 +17,8 @@ const devMode = argv.dev || options.http
 const version = require('../package.json').version
 logger.info('Version: %s', version)
 
+let settings
+
 function run () {
   let count = 0
   let handler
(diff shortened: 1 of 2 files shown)


lidio601:
[lamassu-install@stable results in settings-loader.js error](https://trello.com/c/f5uGYsIL/626-lamassu-installstable-results-in-settings-loaderjs-error)

lib/app.js
```diff
@@ -17,6 +17,8 @@ const devMode = argv.dev || options.http
 const version = require('../package.json').version
 logger.info('Version: %s', version)
 
+let settings
```
joshmh: I don't think this has to be a global mutable variable. See more comments, below.

lib/app.js
```diff
@@ -27,18 +29,23 @@ function run () {
     logger.error('[%d] Retrying in 10s...', count)
   }
 
-  const runner = () => runOnce()
-    .then(() => clearInterval(handler))
-    .catch(errorHandler)
+  const runner = () =>
+    settingsLoader.loadLatest()
+      .then((loadedSettings) => {
+        settings = loadedSettings
```
joshmh: I think we can delete this line, change ``loadedSettings`` to ``settings``, and just pass settings to ``loadSanctions`` and ``startServer``.

In order to do this, we need to change the code to something like this:

```javascript
    settingsLoader.loadLatest()
      .then(settings => {
        clearInterval(handler)
        return loadSanctions(settings)
        .then(() => startServer(settings))
        .catch(errorHandler)
      })
```

PR #177, opened 2018-09-28
Disable motd scripts

naconner:
Disable post-SSH logon prompts to take upgrade actions which could break the server.

### DIFF
--- a/bin/lamassu-update
+++ b/bin/lamassu-update
@@ -71,6 +71,15 @@ BACKUP_CRON="@daily $BACKUP_CMD > /dev/null"
 ( (crontab -l 2>/dev/null || echo -n "") | grep -v '@daily.*lamassu-backup-pg'; echo $BACKUP_CRON ) | crontab - >> $LOG_FILE 2>&1
 $BACKUP_CMD >> $LOG_FILE 2>&1
 
+decho "updating motd scripts
+set +e
+chmod -x /etc/update-motd.d/*-release-upgrade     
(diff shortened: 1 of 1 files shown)


naconner:
[Use lamassu-install and lamassu-update to set Ubuntu to not prompt users to run 'do-release-upgrade'](https://trello.com/c/Z58BGVEb/632-use-lamassu-install-and-lamassu-update-to-set-ubuntu-to-not-prompt-users-to-run-do-release-upgrade)

bin/lamassu-update
```diff
@@ -71,6 +71,15 @@ BACKUP_CRON="@daily $BACKUP_CMD > /dev/null"
 ( (crontab -l 2>/dev/null || echo -n "") | grep -v '@daily.*lamassu-backup-pg'; echo $BACKUP_CRON ) | crontab - >> $LOG_FILE 2>&1
 $BACKUP_CMD >> $LOG_FILE 2>&1
 
+decho "updating motd scripts
+set +e
+chmod -x /etc/update-motd.d/*-release-upgrade
```
joshmh: Why chmod instead of just deleting? Also, do we want to add our own motd?

naconner: No reason, though perhaps we want to re-enable update/release notifications if we have future versions that are update-resilient.

We could add our own motd, one that displays current and latest server versions, server process status, etc.

However, would prefer to do this for now and circle back to that, since we need to pin down lamassu-install for #stable soon.

PR #178, opened 2018-09-28
Disable motd scripts

naconner:
Disable post-SSH logon prompts to take upgrade actions which could break the server.

Includes change from https://github.com/lamassu/lamassu-server/pull/175.


naconner:
[Use lamassu-install and lamassu-update to set Ubuntu to not prompt users to run 'do-release-upgrade'](https://trello.com/c/Z58BGVEb/632-use-lamassu-install-and-lamassu-update-to-set-ubuntu-to-not-prompt-users-to-run-do-release-upgrade)

joshmh:
No need to submit PR twice for dev and stable. Closing this one.

PR #179, opened 2018-10-02
7.3 crafty chhnemu | https://trello.com/c/3fMfHclE

pedromiguelmiranda:
card: https://trello.com/c/3fMfHclE

Unfortunately, Inquirer.js API does not provide signatures to change Initial instructions as requested on card. Can we live with this?

### DIFF
--- a/lib/blockchain/install.js
+++ b/lib/blockchain/install.js
@@ -104,7 +104,7 @@ function run () {
       name: c.display,
       value: c.code,
       checked,
-      disabled: c.cryptoCode.toString() === 'ETH'
+      disabled: c.cryptoCode === ETHEREUM_CRYPTO_CODE
         ? 'Installed, use Infura option'
         : checked && 'Installed'
     }
--- a/lib/blockchain/install.js
+++ b/lib/blockchain/install.js
@@ -97,13 +97,17 @@ function plugin (crypto) {
 }
 
 function run () {
+  const ETH_CRYPTO_CODE ='ETH'
   const choices = _.map(c => {
     const checked = isInstalledSoftware(c) && isInstalledVolume(c)
     return {
       name: c.display,
       value: c.code,
       checked,
-      disabled: checked && 'Installed'
+      disabled: z => {
+        if (checked || c.cryptoCode.toString() === ETH_CRYPTO_CODE)
+          return 'Installed, use Infura option'
+      }
     }
   }, cryptos)
 
@@ -117,5 +121,8 @@ function run () {
   })
 
   inquirer.prompt(questions)
-    .then(answers => processCryptos(answers.crypto))
+    .then(answers => {
+      answers.crypto.push(coinUtils.getCryptoCurrency(ETH_CRYPTO_CODE).code)
+      processCryptos(answers.crypto)
+    })
 }
--- a/lib/blockchain/install.js
+++ b/lib/blockchain/install.js
@@ -97,17 +97,16 @@ function plugin (crypto) {
 }
 
 function run () {
-  const ETH_CRYPTO_CODE ='ETH'
+  const ETHEREUM_CRYPTO_CODE ='ETH'
   const choices = _.map(c => {
     const checked = isInstalledSoftware(c) && isInstalledVolume(c)
     return {
       name: c.display,
       value: c.code,
       checked,
-      disabled: z => {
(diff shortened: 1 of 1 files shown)


pedromiguelmiranda:
[Direct operators away from Geth with lamassu-coins script](https://trello.com/c/3fMfHclE/620-direct-operators-away-from-geth-with-lamassu-coins-script)

lib/blockchain/install.js
```diff
@@ -117,5 +121,8 @@ function run () {
   })
 
   inquirer.prompt(questions)
-    .then(answers => processCryptos(answers.crypto))
+    .then(answers => {
+      answers.crypto.push(coinUtils.getCryptoCurrency(ETH_CRYPTO_CODE).code)
```
pedromiguelmiranda: limitation of Inquirer.js lib:
When a checkbox option is marked as disabled, it is not included on answers.

joshmh: Why not just use ``ETH`` instead of ``coinUtils.getCryptoCurrency(ETH_CRYPTO_CODE).code``?

lib/blockchain/install.js
```diff
@@ -97,13 +97,17 @@ function plugin (crypto) {
 }
 
 function run () {
+  const ETH_CRYPTO_CODE ='ETH'
```
joshmh: Since ``ETH`` is in the constant name itself, I don't think this extra redirection is worth it. Could just use the string ``ETH`` when needed.

lib/blockchain/install.js
```diff
@@ -97,13 +97,17 @@ function plugin (crypto) {
...
     const checked = isInstalledSoftware(c) && isInstalledVolume(c)
     return {
       name: c.display,
       value: c.code,
       checked,
-      disabled: checked && 'Installed'
+      disabled: z => {
+        if (checked || c.cryptoCode.toString() === ETH_CRYPTO_CODE)
+          return 'Installed, use Infura option'
+      }
```
joshmh: I think this will show the wrong message if the coin is not ETH and is already checked (installed). Could do this (up above):

```javascript
const disabled = c.cryptoCode.toString() === 'ETH'
  ? 'Installed, use Infura option' 
  : checked && 'Installed'
```

pedromiguelmiranda: good suggestion.

lib/blockchain/install.js
pedromiguelmiranda: Since this constant is used two more times on this specific function, I think it is a good practice to define it once, and use the defined variable wherever I need. Otherwise would need to change it in several places if such change would be required.
On the other hand, I don't think that ethereum will change its acronym from ETH to something else...

lib/blockchain/install.js
pedromiguelmiranda: We could hard code the crypto code here (according to coin-utils.js, the code for ETH is 'code: 'ethereum').
However, I don't think it is a good practice to hard code it: 
- We have a nice set of utilities on coin-utils module,. So, every time we need to know/change some of we limit our scope of action to this specific module.
Imagine if we hard code this string among several modules and, for some reason, we need to change the string?
We would have to scan all the modules for all the hard coded strings.

lib/blockchain/install.js
pedromiguelmiranda: good point.

lib/blockchain/install.js
```diff
@@ -97,17 +97,16 @@ function plugin (crypto) {
...
     const checked = isInstalledSoftware(c) && isInstalledVolume(c)
     return {
       name: c.display,
       value: c.code,
       checked,
-      disabled: z => {
-        if (checked || c.cryptoCode.toString() === ETH_CRYPTO_CODE)
-          return 'Installed, use Infura option'
-      }
+      disabled: c.cryptoCode.toString() === 'ETH'
```
joshmh: Since we're using ``ETHEREUM_CRYPTO_CODE``, use it here.

joshmh: Also, I believe cryptoCode is already a string.

pedromiguelmiranda: copy&paste is the root of all evil.
Well spotted.

joshmh:
@pedromiguelmiranda This PR was already merged, so it can't be merged anymore. You'll have to create a new PR for it.

PR #181, opened 2018-10-04
Drop Geth installation upon coins installation

pedromiguelmiranda:
GUI output is performing as expected, however, coins installation still proceeds for Geth deamon install. Corrected the behaviour.


pedromiguelmiranda:
[Direct operators away from Geth with lamassu-coins script](https://trello.com/c/3fMfHclE/620-direct-operators-away-from-geth-with-lamassu-coins-script)

PR #186, opened 2018-10-11
Expand customer results

naconner:
List 1000 customers instead of 20.


naconner:
[Expand list results in Customers panel](https://trello.com/c/duCEIQaz/660-expand-list-results-in-customers-panel)

PR #189, opened 2018-10-19
Adjust mock-ticker rates, add warning

naconner:
In case of accidental use in production, use more favourable rates and add a warning in the admin.


naconner:
[Warn against mock ticker use, use favourable rates](https://trello.com/c/8uROAE2P/671-warn-against-mock-ticker-use-use-favourable-rates)

naconner:
Let's make for Bullish Bunene.

PR #190, opened 2018-10-22
Change size of operatorId

RafaelTaranto:


RafaelTaranto:
[Enabling CoinATMRadar results in polluted error log with error codes 400 / 500](https://trello.com/c/yQ2JBXO9/674-enabling-coinatmradar-results-in-polluted-error-log-with-error-codes-400-500)

PR #196, opened 2018-10-26
Remove clutter from logger

RafaelTaranto:


RafaelTaranto:
[Ticker API errors result in log dump of code and CloudFlare HTML](https://trello.com/c/i2KsN9OG/687-ticker-api-errors-result-in-log-dump-of-code-and-cloudflare-html)

RafaelTaranto:

lib/logger.js
```diff
@@ -4,8 +4,12 @@ const options = require('./options')
 const logger = new winston.Logger({
   level: options.logLevel,
   transports: [
-    new (winston.transports.Console)({timestamp: true, colorize: true})
-  ]
+    new (winston.transports.Console)({ timestamp: true, colorize: true })
+  ],
+  rewriters: [(_, __, meta) => {
```
joshmh: Here's a way to not require dummy variables:

```suggestion
  rewriters: [(...[,,meta]) => {
```

lib/logger.js
```diff
@@ -4,8 +4,12 @@ const options = require('./options')
...
+    if (meta) return meta.toString()
```
joshmh: Instead of checking truthiness of ``meta``, check if it has ``toString()``.

joshmh (changes requested):

lib/logger.js
```diff
@@ -1,15 +1,15 @@
...
   level: options.logLevel,
   transports: [
     new (winston.transports.Console)({ timestamp: true, colorize: true })
   ],
-  rewriters: [(_, __, meta) => {
-    if (meta) return meta.toString()
-  }]
-
+  rewriters: [
+    (...[,, meta]) => _.hasIn('toString', meta) ? meta.toString() : meta
```
joshmh: Hmm, if meta doesn't have a ``toString`` function, we probably don't want to print it out. Maybe this should just be some fixed message, like "Error, no further information is available".

joshmh (changes requested):

lib/logger.js
RafaelTaranto: Good catch! What a brain fart =p

PR #199, opened 2018-10-31
Migration sql changes

lidio601:


lidio601:
[lamassu-update: Don't die on migration error](https://trello.com/c/7RFv8h6H/639-lamassu-update-dont-die-on-migration-error)

PR #201, opened 2018-10-31
Fix logger clutter message

RafaelTaranto:

### DIFF
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -1,6 +1,6 @@
 const winston = require('winston')
 const options = require('./options')
-const _ = require('lodash')
+const _ = require('lodash/fp')
 
 const logger = new winston.Logger({
   level: options.logLevel,
@@ -8,8 +8,7 @@ const logger = new winston.Logger({
     new (winston.transports.Console)({ timestamp: true, colorize: true })
   ],
   rewriters: [
-    (...[,, meta]) => _.hasIn('toString', meta)
-      ? meta.toString() : 'Error, no further information is available'
+    (...[,, meta]) => meta instanceof Error ? { message: meta.message, stack: meta.stack } : meta
   ]
 })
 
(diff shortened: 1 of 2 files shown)


lib/logger.js
```diff
@@ -1,15 +1,14 @@
...
+const _ = require('lodash/fp')
 
 const logger = new winston.Logger({
   level: options.logLevel,
   transports: [
     new (winston.transports.Console)({ timestamp: true, colorize: true })
   ],
   rewriters: [
-    (...[,, meta]) => _.hasIn('toString', meta)
-      ? meta.toString() : 'Error, no further information is available'
```
RafaelTaranto: 'Error, no further information is available' isn't actually needed, I'm not sure why it made sense on the last review, maybe because my other mistakes were silly ones.

RafaelTaranto:
["Error, no further information is available" appended to every log message](https://trello.com/c/cHvt2jJx/701-error-no-further-information-is-available-appended-to-every-log-message)

lib/logger.js
joshmh: I think we were concerned that we'd get huge error strings again, if winston tried to print an object. Can you confirm that won't happen?

RafaelTaranto: The main cause for huge js dump were errors from axios. I'm sanitizing only instances of `Error` which are the most common objects passed to the logger, any other object will get fully printed. 

The issue with the meta check is that not every message will have a meta. For instance:
```
logger.error('this is a error')
logger.error('this is a error', null)

// meta === {} or null depending on the second parameter
// msg === 'this is a error'
(...[level, msg, meta]) =>...
```

We would be appending `[object Object]` or `Error, no further information is available` to a lot of valid message strings.

joshmh: Ah right.

PR #203, opened 2018-11-01
Update last_online device field on each machine ping

kokipsiho:


lib/plugins.js
```diff
@@ -231,14 +231,20 @@ function plugins (settings, deviceId) {
   }
 
   function recordPing (serialNumber, deviceTime, rec) {
-    const r = {
+    const r_mp = {
```
joshmh: Should be camel-case. Also, if there are multiple variables, longer names are better.

lib/plugins.js
```diff
@@ -231,19 +231,19 @@ function plugins (settings, deviceId) {
...
-    const r_d = {
+    const rDevices = {
       last_online: deviceTime
     }
 
     return Promise.all([
-      db.none(pgp.helpers.insert(r_mp, null, 'machine_pings')),
-      db.none(pgp.helpers.update(r_d, null, 'devices') + `WHERE device_id = '${deviceId}'`)
+      db.none(pgp.helpers.insert(rMachinePings, null, 'machine_pings')),
+      db.none(pgp.helpers.update(rDevices, null, 'devices') + `WHERE device_id = '${deviceId}'`)
```
joshmh: Missed this last time. You should let pgp do the interpolation of ``deviceId`` instead of using JS backticks. Otherwise the quotes won't be escaped properly in certain situations and it could even lead to sql injection attacks.

Also, I don't think we need the ``r`` prefix here. Why not just ``machinePings`` and ``devices``.

PR #206, opened 2018-11-05
Filter only mock options when in dev mode

kokipsiho:


lib/admin/config.js
```diff
@@ -116,6 +116,12 @@ function fetchConfigGroup (code) {
 
       const selectedCryptos = _.defaultTo([], getCryptos(config, machineList))
 
+      if (devMode) {
+        const allowed = ['mock-ticker', 'mock-wallet', 'mock-exchange', 'mock-sms', 'mock-id-verify', 'mock-zero-conf']
+        const isValid = o => _.includes(o.code, allowed)
+        data.accounts = _.filter(isValid, data.accounts)
+      }
+
```
joshmh: I think there was a misunderstanding with the spec:

> l-a-s should only supply the mock-ticker, mock-wallet, mock-exchange, mock-sms, mock-id-verify, and mock-zero-conf options to l-a when ``--dev`` flag is present

Reading it over, I can see how it could be confusing. What I meant was that in devMode all options should be shown, but in non-devMode, all options besides these options should be shown. We want to hide this options in production mode because they are dangerous.

Let's do some unit tests for this to make the behavior easier to see.

kokipsiho: How would you suggest to unit test this?

lib/admin/config.js
```diff
@@ -116,6 +117,12 @@ function fetchConfigGroup (code) {
 
       const selectedCryptos = _.defaultTo([], getCryptos(config, machineList))
 
+      if (!devMode) {
+        const notAllowed = ['mock-ticker', 'mock-wallet', 'mock-exchange', 'mock-sms', 'mock-id-verify', 'mock-zero-conf']
+        const filterOut = o => _.includes(o.code, notAllowed)
+        data.accounts = _.filter(a => !filterOut(a), data.accounts)
```
joshmh: Instead of mutating ``data`` object, apply this filter in ``fetchData``.

lib/admin/config.js
joshmh: Maybe just create a separate function that take data.account and devMode as arguments and unit test that.

test/unit/filter-config-accounts.js
```diff
@@ -0,0 +1,24 @@
...
+    {code: 'bitcoind', display: 'bitcoind', class: 'wallet', cryptos: ['BTC']},
+    {code: 'quadrigacx', display: 'QuadrigaCX', class: 'exchange', cryptos: ['BTC', 'ETH', 'LTC', 'BCH']},
+    {code: 'mock-wallet', display: 'Mock (Caution!)', class: 'wallet', cryptos: ALL_CRYPTOS}
+  ]
+}
+
+test('Do not filter accounts in dev mode', t => {
+  const devMode = true
+  const filteredData = filterAccounts(data, devMode)
+  t.is(filteredData.accounts.length, 4)
```
joshmh: These tests should check the codes of the returned accounts. Just checking the length could miss errors where devMode gets only the blocked accounts instead of the allowed accounts.

PR #207, opened 2018-11-05
Change seed file with a bip39 mnemonic

RafaelTaranto:

### DIFF
--- a/bin/bip39
+++ b/bin/bip39
@@ -2,12 +2,8 @@
 
 'use strict'
 
-const bip39 = require('bip39')
+const mnemonicHelpers = require('../lib/mnemonic-helpers')
 
 const seed = process.argv[2]
 
-const words = bip39.entropyToMnemonic(seed).split(' ')
-
-for (let i = 0; i < words.length; i += 6) {
-  console.log(words.slice(i, i + 6).join(' '))
-}
+console.log(mnemonicHelpers.fromSeed(seed))
--- a/bin/bip39
+++ b/bin/bip39
@@ -1,0 +1,13 @@
+#!/usr/bin/env node
+
+'use strict'
+
+const bip39 = require('bip39')
+
+const seed = process.argv[2]
+
+const words = bip39.entropyToMnemonic(seed).split(' ')
+
+for (let i = 0; i < words.length; i += 6) {
+  console.log(words.slice(i, i + 6).join(' '))
+}
--- a/bin/cert-gen.sh
+++ b/bin/cert-gen.sh
@@ -17,7 +17,7 @@ mkdir -p $CERT_DIR
 mkdir -p $CONFIG_DIR >> $LOG_FILE 2>&1
 
 echo "Generating mnemonic..."
-MNEMONIC_DIR=mnemonics
+MNEMONIC_DIR=$CONFIG_DIR/mnemonics
 MNEMONIC_FILE=$MNEMONIC_DIR/mnemonic.txt
 mkdir -p $MNEMONIC_DIR >> $LOG_FILE 2>&1
 SEED=$(openssl-1.0 rand -hex 32)
--- a/bin/cert-gen.sh
+++ b/bin/cert-gen.sh
@@ -16,12 +16,13 @@ OFAC_DATA_DIR=$CONFIG_DIR/ofac
 mkdir -p $CERT_DIR
 mkdir -p $CONFIG_DIR >> $LOG_FILE 2>&1
 
-echo "Generating seed..."
-SEEDS_DIR=seeds
-SEED_FILE=$SEEDS_DIR/seed.txt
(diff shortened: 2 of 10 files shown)


RafaelTaranto:
[Switch seed file to mnemonic](https://trello.com/c/3w3vmSRD/707-switch-seed-file-to-mnemonic)

bin/lamassu-update
```diff
@@ -4,6 +4,9 @@ set -e
 export LOG_FILE=/tmp/update.$(date +"%Y%m%d").log
 export NPM_BIN=$(npm -g bin)
 
+MNEMONIC_FILE=$HOME/mnemonics/mnemonic.txt
```
joshmh: * Should be based off of lamassu.json
* Should update lamassu.json
* Path should be ``/etc/lamassu/mnemonics/mnemonic.txt``

lib/wallet.js
```diff
@@ -28,8 +29,9 @@ function computeSeed (masterSeed) {
 }
 
 function fetchWallet (settings, cryptoCode) {
-  return fs.readFile(options.seedPath, 'utf8')
-    .then(hex => {
+  return fs.readFile(options.mnemonicPath, 'utf8')
+    .then(mnemonic => {
+      const hex = bip39.mnemonicToEntropy(mnemonic.split('\n').join(' ').trim())
```
joshmh: This should be done in a separate module so that it's consistent in different places in the code. Should return a buffer.

bin/bip39
```diff
@@ -0,0 +1,13 @@
...
+
+const bip39 = require('bip39')
+
+const seed = process.argv[2]
+
+const words = bip39.entropyToMnemonic(seed).split(' ')
+
+for (let i = 0; i < words.length; i += 6) {
+  console.log(words.slice(i, i + 6).join(' '))
+}
```
joshmh: This should be in a separate module so it can be used in both places.

joshmh (changes requested):

bin/cert-gen.sh
```diff
@@ -16,12 +16,13 @@ OFAC_DATA_DIR=$CONFIG_DIR/ofac
...
 mkdir -p $CONFIG_DIR >> $LOG_FILE 2>&1
 
-echo "Generating seed..."
-SEEDS_DIR=seeds
-SEED_FILE=$SEEDS_DIR/seed.txt
-mkdir -p $SEEDS_DIR >> $LOG_FILE 2>&1
-SEED=$(openssl rand -hex 32)
-echo $SEED > $SEED_FILE
+echo "Generating mnemonic..."
+MNEMONIC_DIR=mnemonics
```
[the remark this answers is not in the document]
RafaelTaranto: Oh, previously the dev environment had the seeds file on the root of the project.

bin/lamassu-update
RafaelTaranto: The `lamassu.json` update happens on `lamassu-update-to-mnemonic` script.

> Should be based off of lamassu.json
> Path should be /etc/lamassu/mnemonics/mnemonic.txt

I don't get it, should I create a script to get the path from `lamassu.json` and default the value to `/etc/lamassu/mnemonics/mnemonic.txt` ?

joshmh: In this case, you should run lamassu-update-to-mnemonic every time. lamassu-update-to-mnemonic should gracefully fail when the mnemonic file is present. That way, this script (lamassu-update) doesn't have to deal with mnemonic path logic.

Another issue is that lamassu-update-to-mnemonic is used for both developers and production. In production, we want to write to ``/etc/lamassu/mnemonics`` and for developers we probably want the path to be ``~/.lamassu/mnemonics``. So the script should be able to handle both cases depending on who is calling it.

joshmh (changes requested):

lib/coinatmradar/coinatmradar.js
```diff
@@ -1,11 +1,13 @@
 const axios = require('axios')
 const _ = require('lodash/fp')
 const hkdf = require('futoin-hkdf')
+const bip39 = require('bip39')
```
joshmh: Not necessary. Editor configured for Standard should warn you about this.

```suggestion
```

lib/wallet.js
```diff
@@ -1,11 +1,13 @@
 const _ = require('lodash/fp')
 const mem = require('mem')
 const hkdf = require('futoin-hkdf')
+const bip39 = require('bip39')
```
joshmh: Not necessary.

```suggestion
```

joshmh (changes requested):

lib/coinatmradar/coinatmradar.js
RafaelTaranto: Hum, looks like I'm having some issues with workspace configuration on vscode =p

PR #208, opened 2018-11-08
Add optional repository identifier argument to lamassu-update

chrisrico:
Also since lamassu-update-to-mnemonic isn't put in the `node_modules/bin` folder, `$SCRIPTPATH` doesn't help get you there, so I use `SERVER_ROOT` instead. 

Would it maybe be better to have the repository identifier as an environment variable? Read from the lamassu.json?


joshmh:
@chrisrico This is a good idea, I think we'll be incorporating this in a change we're making now.

PR #209, opened 2018-11-09
Add itBit exchange/ticker

chrisrico:
Still need to an an error handler to extract information about failed calls.

It was unclear how to compile elm.js so I edited it by hand.


chrisrico:
We recently ordered a Sintra Forte and we use ItBit as our exchange provider.

joshmh:
Thanks @chrisrico! We've put it on our priority list, although it still might take some time before it's merged.

chrisrico:
Is there anything I can do to speed up the process?

joshmh:
You can check back in a week. We'll let you know if we have any questions once we look at it more carefully.

chrisrico:
Ok, thanks. Didn't meant to imply there was any rush, we won't have our machine until next month at the earliest, and it probably won't be in operation until early 2019.

joshmh:
Cool, shouldn't be a problem then :)

PR #213, opened 2018-11-16
Echo message on openssl fail 

DatoBula:


DatoBula:
[Openssl fails silently on .sh files](https://trello.com/c/Hz4M8oSW/712-openssl-fails-silently-on-sh-files)

PR #214, opened 2018-11-16
Move lamassu-install files to this repo's lamassu-remote-install folder.

DatoBula:


DatoBula:
[Merge l-i into l-s](https://trello.com/c/wSKQd6Rb/718-merge-l-i-into-l-s)

lamassu-remote-install/install
```diff
@@ -0,0 +1,257 @@
...
+SEED=$(openssl rand -hex 32)
+echo $SEED > $SEED_FILE
+
+decho "Installing latest npm package manager for node..."
+retry 3 npm -g --unsafe-perm install npm@5  >> $LOG_FILE 2>&1
+NODE_MODULES=$(npm -g root)
+NPM_BIN=$(npm -g bin)
+
+decho "Installing lamassu-server..."
+retry 3 npm -g --unsafe-perm install lamassu/lamassu-server#master >> $LOG_FILE 2>&1
```
lidio601: hey @joshmh I was thinking about the branch name: what if we have the branch name as parameter?
it can default to `master` otherwise we can pass in something like:
```
curl -sS https://raw.githubusercontent.com/lamassu/lamassu-server/master/lamassu-remote-install/install | bash -s -- master
```

joshmh: Yeah, I like that idea. @DatoBula any reason not to do this?

lamassu-remote-install/test.js
```diff
@@ -0,0 +1 @@
+console.log(__dirname)
```
joshmh: We can delete this file.

lamassu-remote-install/install
DatoBula: No. I will do this.

PR #216, opened 2018-11-19
lamassu-migrate-config to update ls basepath

lidio601:


lidio601:
[lamassu-update: Results in admin looking for cert in old directory](https://trello.com/c/cVf9vxA8/640-lamassu-update-results-in-admin-looking-for-cert-in-old-directory)

PR #221, opened 2018-11-21
lamassu-update to migrate to new /opt/lamassu-server install path

lidio601:
update.log https://gist.github.com/lidio601/65600ab3677b998d31edc29f2c85cb06


lidio601:
[lamassu-update: overcome error if lamassu-server-old already exists](https://trello.com/c/hvV0BAk4/692-lamassu-update-overcome-error-if-lamassu-server-old-already-exists)

lidio601:
@joshmh @naconner seems working fine.
The only thing I wasn't able to solve is how to export the new PATH to the running terminal.

the issue is that after running `curl -sS https://raw.githubusercontent.com/lamassu/lamassu-server/eec09883efbba8f53d951e229db265e9b528d25d/bin/lamassu-update | bash`, if then you type `which lamassu-server` it won't be able to find the new `/opt/lamassu-server/bin/lamassu-server` until a new shell is opened (once the new /etc/environment is loaded)

PR #224, opened 2018-11-21
Log unsuccessful trade attempts in the db

DatoBula:


DatoBula:
[Log unsuccessful trade attempts in the db](https://trello.com/c/OkJv993t/612-log-unsuccessful-trade-attempts-in-the-db)

lib/plugins.js
```diff
@@ -541,14 +547,19 @@ function plugins (settings, deviceId) {
...
       convertBigNumFields,
       _.mapKeys(_.snakeCase)
     )
 
     const tradeEntry = massage(_tradeEntry)
+
+    if (error) {
+      tradeEntry.error = error.message
+    }
+
```
joshmh: To make this more functional (by not mutating a record), try adding the error in ``massage``. Other than that, looks good!

lib/plugins.js
```diff
@@ -541,14 +547,19 @@ function plugins (settings, deviceId) {
...
     const massage = _.flow(
       _.pick(['cryptoCode', 'cryptoAtoms', 'fiatCode', 'type']),
       convertBigNumFields,
       _.mapKeys(_.snakeCase)
     )
 
     const tradeEntry = massage(_tradeEntry)
+
+    if (error) {
+      tradeEntry.error = error.message
```
joshmh: Should truncate this to some sane max length, like 200 chars.

PR #225, opened 2018-11-22
migrate-options unit test added

lidio601:
I took out the updateOptionBasepath function and added a unit test.
I'll port the same on master


lidio601:
[lamassu-update: Results in admin looking for cert in old directory](https://trello.com/c/cVf9vxA8/640-lamassu-update-results-in-admin-looking-for-cert-in-old-directory)

PR #227, opened 2018-11-23
Log commission percentage on txt table

lidio601:


lidio601:
[Log commission percentage at time of transaction](https://trello.com/c/fOib7cif/732-log-commission-percentage-at-time-of-transaction)

lib/plugins.js
```diff
@@ -30,6 +30,15 @@ const PONG_TTL = '1 week'
 const tradesQueues = {}
 
 function plugins (settings, deviceId) {
+  function getCommissionForTx (tx) {
```
joshmh: I think it would be better to add the commission to the tx when it's created, e.g.: https://github.com/lamassu/lamassu-machine/blob/dev/lib/brain.js#L1722

Then we could use that setting when we do the rate computations, as well.

joshmh (changes requested):

lib/plugins.js
lidio601: uhm, I'm not sure it's a good idea to let l-m send commission percentage.
Also I'm calling this function on ls postTx API which is the same as https://github.com/lamassu/lamassu-machine/blob/dev/lib/brain.js#L1722 but done on l-s side.

lidio601: I was checking on l-m, I think we only output finished rates so we would need to expose this setting through l-s API, to allow l-m set it on tx creation and then send it through postTx API.
Seems kind of complicated to send back and forth this number while we already have it on l-s side.

RafaelTaranto: @joshmh Did this got settled? Other than that LGTM!

joshmh: Not sure. We just have to make sure we're using the same value for the actual commission and price calculations.

RafaelTaranto:
@joshmh this can be closed, replaced by https://github.com/lamassu/lamassu-server/pull/250 and https://github.com/lamassu/lamassu-machine/pull/251

PR #228, opened 2018-11-25
Log ticker price with tx record

lidio601:
Relies on #227


lidio601:
[Log ticker rate at time of transaction](https://trello.com/c/6IpkHUXu/733-log-ticker-rate-at-time-of-transaction)

lib/tx.js
```diff
@@ -16,15 +16,22 @@ function post (tx, pi) {
...
   const transformDate = (v, k) => isDateField(k) ? new Date(v) : v
   const mapValuesWithKey = _.mapValues.convert({'cap': false})
   const transformDates = r => mapValuesWithKey(transformDate, r)
   const logCommission = r => _.assign(r, {
-    commissionPercentage: _.get(direction, pi.getCommissionPercentage(cryptoCode))
+    commissionPercentage: _.get(direction
+      , pi.getCommissionPercentage(cryptoCode))
+  })
+
+  const tickerPrice = await pi.getRawTickerPrice(fiatCode, cryptoCode)
```
RafaelTaranto, line 32: @joshmh  This is receiving proper error handling internally, but we should be careful on what we're adding on this method. This made the code feel a little fragile, but I don't think there's a better place for it to live. 

The code is also using async/await, is that ok?

RafaelTaranto: Also this code is using async/await, is it ok?

joshmh: What part do we have to be careful about? I don't think it needs to be async/await. We should be caching current rates periodically and just use that cached value. We shouldn't poll each time we need the rates in the code.

RafaelTaranto: This `massage` function is a critical one, adding in async function calls without proper error handling can make tx fail for ignorable reasons (like saving ticker price). In this case we're safe because of the error handling inside the ticker.js

This is cached already, I'm just going to swap async await with promises

RafaelTaranto:
@joshmh this can be closed, replaced by https://github.com/lamassu/lamassu-server/pull/250 and https://github.com/lamassu/lamassu-machine/pull/251

PR #229, opened 2018-11-26
Ln strike support added back. Refactorings.

dabulashvili-zz:


DatoBula:
[LN: Add back Strike support](https://trello.com/c/uVEPJtYG/588-ln-add-back-strike-support)

lib/layer2.js
```diff
@@ -11,28 +13,117 @@ function fetch (settings, cryptoCode) {
...
+  const layer2StatusPromise = getLayer2Status(settings, tx)
+
+  return Promise.all([
+    wallet.getStatus(settings, tx),
+    layer2StatusPromise
+  ])
+    .then(([walletStatus, layer2Status]) => {
+      return mergeStatus(walletStatus, layer2Status)
+    })
+    .then((statusRec) => {
```
joshmh: This section applies to all wallets, not just layer2, so this should remain in wallet.js.

lib/plugins.js
```diff
@@ -13,6 +13,7 @@ const T = require('./time')
 const configManager = require('./config-manager')
 const ticker = require('./ticker')
 const wallet = require('./wallet')
+const layer2 = require('./layer2')
```
joshmh: plugins.js shouldn't know about layer2, just wallet.

lib/wallet.js
```diff
@@ -77,17 +75,8 @@ function sendCoins (settings, toAddress, cryptoAtoms, cryptoCode) {
 }
 
 function newAddress (settings, info) {
-  const walletAddressPromise = fetchWallet(settings, info.cryptoCode)
+  return fetchWallet(settings, info.cryptoCode)
```
joshmh: Need to update fetchWallet so that it returns the layer2 plugin, when enabled.

joshmh (changes requested):

lib/wallet.js
DatoBula: Instead of wallet? If I do that, I will have to write second `fetchWallet` and `newAddress` for only wallet without layer2 because `layer2.newAddress` uses `wallet.newAddress`, so wallet fetching will be executed twice. 
Did I get your review?

DatoBula: Now I think i got it, I can make composite of wallet and layer2.

lib/layer2.js
```diff
@@ -1,36 +1,99 @@
 const configManager = require('./config-manager')
 const ph = require('./plugin-helper')
 const _ = require('lodash/fp')
-const logger = require('./logger')
 
-function fetch (settings, cryptoCode) {
+class Layer2Composite {
```
joshmh: We're not using JavaScript classes anywhere in the code, mostly because we don't like the object-oriented paradigm, so let's not add classes in.

I think this is getting too complicated with layer2-composite and wallet-adapter.js. It seems like there a lot of logic in wallet.js and we also don't want to be in a situation where layer2 depends on wallet.js and wallet.js also depends on layer2. Therefore, it seems like the easiest solution is to leave everything in wallet.js, as it was before.

joshmh (changes requested):

PR #234, opened 2018-12-09
Update ZEC, BCH, ETH node versions

naconner:


naconner:
[Update wallet node versions](https://trello.com/c/2R02ikX8/771-update-wallet-node-versions)

PR #235, opened 2018-12-09
Caution against mock-ticker use

naconner:
Rename 'Mock ticker' to 'Mock (Caution!)'


naconner:
[Warn against mock ticker use](https://trello.com/c/8uROAE2P/671-warn-against-mock-ticker-use)

PR #237, opened 2018-12-14
Log miner fee

dabulashvili-zz:


DatoBula:
[Log miner fees in cash_in_txs](https://trello.com/c/Ecayjmum/729-log-miner-fees-in-cashintxs)

PR #239, opened 2018-12-21
Allow the domain to be configurable when generating developer certificates

littlenaut:
The intended domain/IP can be provided as an argument to the bash script.
Example: `$ bash bin/cert-gen.sh 192.168.1.1`

The lack of such argument will default to the previous value: `localhost`.


littlenaut:
[Create ssuboard image](https://trello.com/c/5COrN0uF/755-create-ssuboard-image)

PR #245, opened 2019-01-26
7.2.1, add lamassu-devices

CrypticaScriptura:
For use when relaying device_ids for remote updates.


CrypticaScriptura:
[Tag lamassu-server v7.2.1](https://trello.com/c/dhkeJafj/841-tag-lamassu-server-v721)

PR #246, opened 2019-02-07
Fix cash-out exchange behavior for 'rejected' tx

RafaelTaranto:
A TX on the 'rejected' state should still be watched for authorization.

This fixes the bug where delayed cash-out confirmations would not
trigger a exchange sell command.


RafaelTaranto:
[Exchange auto-trading not performed on 1-conf cash-out](https://trello.com/c/j5GARw8E/837-exchange-auto-trading-not-performed-on-1-conf-cash-out)

PR #248, opened 2019-02-15
Add lamassu-coinatmradar, lamassu-devices, lamassu-operator

CrypticaScriptura:
For use when relaying device_ids for remote updates, and retrieving the operator ID used with CoinATMRadar.

### DIFF
--- a/bin/lamassu-coinatmradar
+++ b/bin/lamassu-coinatmradar
@@ -5,7 +5,7 @@ SEED="$(cat ~/seeds/seed.txt)"
 echo
 echo "Here is the 'External ID' of your paired machine(s), for use under the 'ATM / Teller details' of your CoinATMRadar listing:"
 echo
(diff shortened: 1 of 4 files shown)


bin/lamassu-coinatmradar
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+SEED="$(cat ~/seeds/seed.txt)"
+
+echo
+echo "Here is the 'External ID' of your paired machine(s), for use under the 'ATM / Teller details' of your CoinATMRadar listing:"
+echo
+paste <(su - postgres -c "psql \"lamassu\" -Atc \"select device_id from devices\"" | cut -c -32) <(su - postgres -c "psql \"lamassu\" -Atc \"select name from devices\"") | column -s $'\t' -t
```
joshmh: Shouldn't this be a single JOIN select statement?

joshmh (changes requested):

bin/lamassu-coinatmradar
CrypticaScriptura: Good point. This allowed nice formatting of the result, though just pushed a change with similar formatting using a single select statement.

PR #254, opened 2019-03-05
Compliance: Paper wallet printing

ghost:


ghost:
[Private key printing](https://trello.com/c/A9y5mQE1/808-private-key-printing)

lib/routes.js
```diff
@@ -155,6 +155,15 @@ function postTx (req, res, next) {
     .catch(next)
 }
 
+function createWallet (req, res, next) {
+  const pi = plugins(req.settings, req.deviceId)
+  pi.newAddress(req.tx)
```
joshmh: What about the private key?

joshmh (changes requested):

lib/routes.js
```diff
@@ -367,6 +367,7 @@ app.post('/phone_code', getCustomerWithPhoneCode)
 app.patch('/customer/:id', updateCustomer)
 
 app.post('/tx', postTx)
+app.post('/tx/createWallet', createWallet)
```
joshmh: Where is the ``createWallet`` function?

migrations/1552064796155-tx-paper-wallet.js
```diff
@@ -0,0 +1,15 @@
+'use strict'
+
+const db = require('./db')
+
+exports.up = function (next) {
+  var sql = [
+    'ALTER TABLE cash_in_txs ADD COLUMN is_paper_wallet boolean null DEFAULT false',
```
joshmh, line 7: Is this being used?

joshmh (changes requested):

lib/routes.js
ghost: That one escaped me. Removed.

migrations/1552064796155-tx-paper-wallet.js
ghost: Yes, the [migration script](https://github.com/lamassu/lamassu-server/blob/dev/bin/lamassu-migrate) goes through all the scrips in the migration directory.

joshmh: I mean is the ``is_paper_wallet`` field ever being set?

ghost: Yes, it is. [See here](https://github.com/lamassu/lamassu-machine/blob/449a6cae14d97aa875281cd72372bd040a554699/lib/brain.js#L1775).

PR #255, opened 2019-03-05
Fix deprecated api call on bitgo api

RafaelTaranto:


RafaelTaranto:
[Update BitGo module to API v2](https://trello.com/c/6FaCLmf4/710-update-bitgo-module-to-api-v2)

lib/plugins/wallet/bitgo/bitgo.js
```diff
@@ -94,14 +97,28 @@ function newAddress (account, info) {
...
-      if (requested.gt(rec.balance)) return { status: 'insufficientFunds' }
-      if (requested.gt(rec.confirmedBalance)) return { status: 'authorized' }
-      return { status: 'confirmed' }
+    .then(() => getWallet(account, cryptoCode))
+    .then(wallet => wallet.transfers({ type: 'receive', address: toAddress }))
+    .then(({ transfers }) => {
+      const filterConfirmed = _.filter(it =>
+        it.state === 'confirmed' && it.type === 'receive'
+      )
+      const filterUnconfirmed = _.filter(it =>
```
joshmh: Change this to ``filterPending`` and have it check if state is either unconfirmed or confirmed. Then we won't have to add ``confirmed`` records later on.

joshmh (changes requested):

PR #256, opened 2019-03-05
Change backup folder on updates

RafaelTaranto:


RafaelTaranto:
[lamassu-update fails on presence of lamassu-server-old](https://trello.com/c/UjgSgp1R/905-lamassu-update-fails-on-presence-of-lamassu-server-old)

PR #260, opened 2019-03-07
Fix lamassu-update-to-mnemonic --prod

RafaelTaranto:


RafaelTaranto:
[lamassu-update-to-mnemonic does not generate path or file](https://trello.com/c/w4XeCckj/906-lamassu-update-to-mnemonic-does-not-generate-path-or-file)

PR #263, opened 2019-03-21
Fix BKP_NAME

CrypticaScriptura:
Otherwise, fails with:

```
bash: line 37: BKP_NAME: command not found
```


CrypticaScriptura:
[lamassu-update fails on presence of lamassu-server-old](https://trello.com/c/UjgSgp1R/450-lamassu-update-fails-on-presence-of-lamassu-server-old)

PR #265, opened 2019-03-25
Fetch time till hardLimit clears

RafaelTaranto:


RafaelTaranto:
[Hard limit reached date is hard coded](https://trello.com/c/Zg4fSQIY/139-hard-limit-reached-date-is-hard-coded)

lib/customers.js
```diff
@@ -115,17 +117,32 @@ function getById (id, userToken) {
...
+  let startDate = moment()
+  startDate = startDate.subtract(1, 'days')
+
+  const cashInMoment = moment(cashInDate || startDate)
+  const cashOutMoment = moment(cashOutDate || startDate)
+
+  const cashInDuration = moment.duration(cashInMoment.diff(startDate)).asHours()
+  const cashOutDuration = moment.duration(cashOutMoment.diff(startDate)).asHours()
+
+  return _.ceil(_.max([cashInDuration, cashOutDuration]))
```
joshmh: Do we need to check for negative values?

joshmh (changes requested):

lib/customers.js
RafaelTaranto: Not really because of the sql 'where'.

joshmh: Hmm, maybe but I think an extra check in the logic here will make it safer.

joshmh (changes requested):

lib/customers.js
RafaelTaranto: You're absolutely right =)

PR #267, opened 2019-03-29
Update lamassu-coinatmradar for 64 characters

CrypticaScriptura:
We're passing 64-characters to the CAR API, instead of 32.


CrypticaScriptura:
[Fix Coin ATM Radar API reporting](https://trello.com/c/BMOcLO6V/163-fix-coin-atm-radar-api-reporting)

PR #269, opened 2019-04-02
Send legacy address for BCH on bitgo api

RafaelTaranto:


RafaelTaranto:
[BitGo module not detecting BCH 0-conf deposits](https://trello.com/c/TRGzJTXa/164-bitgo-module-not-detecting-bch-0-conf-deposits)

PR #270, opened 2019-04-03
Add lamassu-revoke and lamassu-users

GenericCPU:
lamassu-users - prints a list of registered users
lamassu-revoke - revokes access from a user based on their name


CrypticaScriptura:
Thanks. Could you add them to package.json's bins so they get symlinked upon install?

CrypticaScriptura:
[Add lamassu-revoke and lamassu-users](https://trello.com/c/DNSqSvcH/169-add-lamassu-revoke-and-lamassu-users)

PR #271, opened 2019-04-16
Fix fee fetching issues

RafaelTaranto:


RafaelTaranto:
[Log miner fees in cash_in_txs](https://trello.com/c/Ecayjmum/80-log-miner-fees-in-cashintxs)

PR #279, opened 2019-05-13
Trim API token's whitespace

CrypticaScriptura:
BitGo's site adds an extraneous space to the API token when copy/pasting to the admin field, causing requests to fail.

Trim whitespace before using. Temporary fix until we have credential field validation in the new admin.


CrypticaScriptura:
[Trim BitGo API token's whitespace](https://trello.com/c/iiPCZdJT/210-trim-bitgo-api-tokens-whitespace)

PR #281, opened 2019-05-22
Add new operator info fields

RafaelTaranto:


RafaelTaranto:
[Private key printing](https://trello.com/c/A9y5mQE1/136-private-key-printing)

PR #284, opened 2019-06-01
Add blacklist script

RafaelTaranto:


RafaelTaranto:
[Script to add addresses to blacklist](https://trello.com/c/BKL2BtF0/212-script-to-add-addresses-to-blacklist)

bin/lamassu-blacklist
```diff
@@ -0,0 +1,68 @@
...
+  input: process.stdin,
+  output: process.stdout
+})
+
+const sql = `select regexp_replace(crypto_code, '$', ' ') as code,regexp_replace(address, '^', ' ') as address from blacklist`
+const insertSql = `insert into blacklist(crypto_code, address) values ($1, $2)`
+
+db.query(sql)
+  .then(many => {
+    console.log('')
```
joshmh: I think you can just use: ``console.log()``.

bin/lamassu-blacklist
```diff
@@ -0,0 +1,68 @@
...
+    console.log('Here is your current list of prohibited addresses:')
```
joshmh:
```suggestion
    console.log('Here is your current list of prohibited addresses:\n')
```

Then you don't need all the empty ``console.log``s.

bin/lamassu-blacklist
```diff
@@ -0,0 +1,68 @@
...
+    console.log('')
+    console.log(many.map(it => `${it.code} | ${it.address}`).join(os.EOL))
+    console.log('')
+    console.log('What address would you like to add to the list?')
+    console.log('')
+    console.log('\x1b[2m%s\x1b[0m', 'Example:', 'bc1q8pu2zlfxg8jf4cyuf844rl3uxmmw8sj9yaa393')
```
joshmh: What do these escape sequences do? How about using https://github.com/chalk/chalk instead?

joshmh (changes requested):

bin/lamassu-blacklist
RafaelTaranto: I though about adding chalk but I don't think it is worth adding a dependency at this point. 

The escape sequences dim the first string so we give a different look to `Example:`.

I can either add chalk or name the escape sequences, up to you.

bin/lamassu-blacklist
RafaelTaranto: I've used os.EOL.

PR #285, opened 2019-06-04
Handle high precision bitpay response numbers

RafaelTaranto:


RafaelTaranto:
[BigNumber error from Kraken and BitPay tickers](https://trello.com/c/O6tCQDt6/232-bignumber-error-from-kraken-and-bitpay-tickers)

CrypticaScriptura:
This resolves the condition noticed earlier today, and published as an interim patch here:

https://support.lamassu.is/hc/en-us/articles/360028823472-Required-ticker-update

PR #286, opened 2019-06-04
Handle high precision bitpay response numbers

RafaelTaranto:


RafaelTaranto:
[BigNumber error from Kraken and BitPay tickers](https://trello.com/c/O6tCQDt6/232-bignumber-error-from-kraken-and-bitpay-tickers)

PR #293, opened 2019-07-15
Turn terms and conditions off if text is empty

RafaelTaranto:


RafaelTaranto:
[If text is missing from T&C box, default to blank text, but don't take l-s down](https://trello.com/c/jTS7YKa0/25-if-text-is-missing-from-tc-box-default-to-blank-text-but-dont-take-l-s-down)

PR #295, opened 2019-07-17
Geth/infura nonce fix

RafaelTaranto:


RafaelTaranto:
[Rapid ETH transactions create duplicate nonce values, with ETH blockchain dropping one](https://trello.com/c/VbY3Gmu1/4-rapid-eth-transactions-create-duplicate-nonce-values-with-eth-blockchain-dropping-one)

PR #299, opened 2019-07-30
bitcoind 0.17/0.18 getbalance compatibility change

error10:
Because of bitcoind deprecating its internal account system, the getbalance RPC call now has a dummy argument as the first argument. The help says it can be omitted or changed to '*', so we change it to '*', both for backward compatibility and because in testing I found that omitting the argument was broken. Using '*' has the same semantics as omitting the argument, if accounts were not used in the previous version of bitcoind.


CrypticaScriptura:
[bitcoind 0.17/0.18 getbalance compatibility change](https://trello.com/c/BCrwcNAJ/60-bitcoind-017-018-getbalance-compatibility-change)

CrypticaScriptura:
Thanks, @error10! :raised_hands:

PR #300, opened 2019-07-31
Prevent address reuse option

RafaelTaranto:


RafaelTaranto:
[Option to reject the reuse of addresses](https://trello.com/c/EoZsnFD9/49-option-to-reject-the-reuse-of-addresses)

PR #301, opened 2019-08-02
Take machineid into account for coinatmradar rates

RafaelTaranto:


RafaelTaranto:
[Coin ATM Radar module reports global commission values, instead of that of individual machines](https://trello.com/c/x8myB5em/9-coin-atm-radar-module-reports-global-commission-values-instead-of-that-of-individual-machines)

RafaelTaranto:
[Display machine version number in corner of initialise and pair screens](https://trello.com/c/D1Io5QN9/33-display-machine-version-number-in-corner-of-initialise-and-pair-screens)

PR #314, opened 2019-09-17
Throttle clock skew and fetch logs

RafaelTaranto:


RafaelTaranto:
['BigNumber', 'clock skew', 'internal server error' rendering all machines down connected to a server](https://trello.com/c/yLrrKD0D/81-bignumber-clock-skew-internal-server-error-rendering-all-machines-down-connected-to-a-server)

lib/routes.js
```diff
@@ -26,12 +26,23 @@ const argv = require('minimist')(process.argv.slice(2))
...
+const THROTTLE_CLOCK_SKEW = 1 * 60 * 1000
 
 const pids = {}
 const reboots = {}
 
 const devMode = argv.dev || options.http
 
+_.mixin({
+  memoizeThrottle: function(wait=0, func) {
+    var mem = _.memoize(function(param) {
```
joshmh:
```suggestion
    const mem = _.memoize(function(param) {
```

lib/routes.js
```diff
@@ -26,12 +26,23 @@ const argv = require('minimist')(process.argv.slice(2))
...
+      return _.throttle(wait, func)
+    });
+    return function(param){mem(param)(param)}
+  }
+});
```
joshmh:
```suggestion
})
```

Make sure the code passes standard.

joshmh (changes requested):

PR #319, opened 2019-09-27
Prevent isZeroConf from being called on cashIn

RafaelTaranto:


RafaelTaranto:
[Error on cash-in only servers with notification enabled](https://trello.com/c/rA9WojCX/88-error-on-cash-in-only-servers-with-notification-enabled)

PR #321, opened 2019-10-04
Add Cardano cryptocurrency

detrandix:


lib/blockchain/cardano.js
```diff
@@ -0,0 +1,47 @@
...
+
+const coinUtils = require('../coin-utils')
+const options = require('../options')
+
+module.exports = { setup }
+
+const es = common.es
+
+// todo: maybe add '#' as second parameter?
+function escapeRegExp(str) {
```
ppershing: hmm, what is this regex supposed to do?

lib/blockchain/cardano.js
```diff
@@ -0,0 +1,47 @@
...
+  return str.replace(/[.*+?^${}()|[\]\\#]/g, '\\$&')
+}
+
+function setup(dataDir) {
+  const coinRec = coinUtils.getCryptoCurrency('ADA')
+
+  // todo: maybe use template tag for escaping?
+  es(
```
ppershing: this definitely needs a comment as to what the command is supposed to do

lib/blockchain/cardano.js
```diff
@@ -0,0 +1,47 @@
...
+    `sudo sed -i 's#state-wallet-testnet/#${escapeRegExp(
+      dataDir
+    )}/#g' /usr/local/bin/cardano-testnet-wallet`
+  )
+
+  common.firewall([coinRec.defaultPort])
+
+  const config = buildConfig(null, null, null)
```
ppershing: could we not use forrawd references? Let's move `buildConfig` before this function

lib/coin-utils.js
```diff
@@ -58,6 +58,15 @@ const CRYPTO_CURRENCIES = [
     daemon: 'bitcoincashd',
     defaultPort: 8335,
     unitScale: 8
+  },
+  {
+    cryptoCode: 'ADA',
+    display: 'Cardano',
+    code: 'cardano',
+    configFile: 'cardano.json',
+    daemon: 'cardano-testnet-wallet',
```
ppershing: hmm, is cardano-sl using "testnet" wallet?

lib/plugins/wallet/cardano/api.js
```diff
@@ -0,0 +1,161 @@
...
+const unitScale = 6
+
+class CardanoAPI {
+  constructor(client, port, hostname = 'localhost') {
+    this.client = client
+    this.port = port
+    this.hostname = hostname
+  }
+
+  async fetch(endpoint, params, data, appendData = []) {
```
ppershing: do we use `appendData` somewhere outside of recursive paging? If not, I would rather have loop than recursive invocations with magic parameter

lib/plugins/wallet/cardano/api.js
```diff
@@ -0,0 +1,161 @@
...
+          params,
+          data
+        })
+      })
+      .then(response => {
+        if (response.error) throw r.error
+
+        const responseData = Array.isArray(response.data.data)
+          ? [...response.data.data, ...appendData]
+          : response.data.data
```
ppershing: this seems suspicious that we don't append anything even if `appendData` might not be empty list

lib/plugins/wallet/cardano/api.js
ppershing: Also, I would call it then `fetchAllPages`

lib/plugins/wallet/cardano/api.js
```diff
@@ -0,0 +1,161 @@
...
+      walletId,
+      accountIndex: index
+    })
+      .then(transactions =>
+        transactions.reduce((acc, item) => {
+          if (item.confirmations < confirmations) {
+            return acc
+          }
+
+          return acc + (item.direction === 'outgoing' ? -1 : 1) * item.amount
```
ppershing: I would rather use mapping
```
sign = { incoming: +1, outgoing: -1}[item.direction]
```
so that it will crash if it doesn't parse direction properly.

lib/plugins/wallet/cardano/api.js
```diff
@@ -0,0 +1,161 @@
...
+      })
+  }
+
+  async walletBalance(walletId, index, confirmations = 1) {
+    return this.fetch('transactions', {
+      walletId,
+      accountIndex: index
+    })
+      .then(transactions =>
+        transactions.reduce((acc, item) => {
```
ppershing: this is a rather complicated reduce. Maybe let's just do `.filter`, then `.map` and finally `.reduce` (or equivalently `_.sum`)

lib/plugins/wallet/cardano/api.js
```diff
@@ -0,0 +1,161 @@
...
+            .filter(
+              record =>
+                record.address === address &&
+                item.confirmations >= confirmations
+            )
+            .reduce((acc, item) => acc + item.amount, 0)
+
+          const outputs = item.outputs
+            .filter(
+              record =>
```
ppershing: we repeat the same logic here. Maybe we can extract it to a function with some good name?

lib/plugins/wallet/cardano/api.js
```diff
@@ -0,0 +1,161 @@
...
+            return acc
+          }
+
+          return acc + (item.direction === 'outgoing' ? -1 : 1) * item.amount
+        }, 0)
+      )
+      .then(r => BN(r).round())
+  }
+
+  async addressBalance(address, confirmations) {
```
ppershing: I think `confirmations` isn't descriptive enough name. At least `min_confirmations` or something like that

lib/plugins/wallet/cardano/api.js
```diff
@@ -0,0 +1,161 @@
...
+  async newWallet(mnemonic, name = 'Server Wallet', operation = 'create') {
+    return this.fetch('wallets', null, {
+      assuranceLevel: 'normal',
+      backupPhrase: mnemonic,
+      name,
+      operation
+    })
+      .then(data => data.id)
+      .catch(e => {
+        if (e.response.status === 403) {
```
ppershing: could this throw? I mean, `e.response` might not be defined

lib/plugins/wallet/cardano/cardano.js
```diff
@@ -0,0 +1,85 @@
+const { walletId } = require('./mnemonic')
+const cardanoApi = require('./client')
+
+// TODO: work with BN here instead of api.js?
+
+async function checkCryptoCode(cryptoCode) {
+  if (cryptoCode !== 'ADA') {
+    return Promise.reject(new Error(`Unsupported crypto: ${cryptoCode}`))
```
ppershing: Hmm, shouldn't we be able to simply `throw` here? It should handle creating promise automatically ... Or is this a general lamassu code-style?

lib/plugins/wallet/cardano/cardano.js
```diff
@@ -0,0 +1,85 @@
+const { walletId } = require('./mnemonic')
+const cardanoApi = require('./client')
+
+// TODO: work with BN here instead of api.js?
+
+async function checkCryptoCode(cryptoCode) {
```
ppershing: does this need to be `async` at all?

lib/plugins/wallet/cardano/cardano.js
```diff
@@ -0,0 +1,85 @@
...
+ *       inject walletId/index into account object
+ *       in wallet.js - it's better?
+ */
+async function balance(account, cryptoCode) {
+  return checkCryptoCode(cryptoCode)
+    .then(() => walletId())
+    .then(({ walletId }) => cardanoApi.walletBalance(walletId, 1))
+}
+
+async function sendCoins(account, address, cryptoAtoms, cryptoCode) {
```
ppershing: `account` is unused. Is that intended? Same above ...

lib/plugins/wallet/cardano/cardano.js
```diff
@@ -0,0 +1,85 @@
...
+  return checkCryptoCode(info.cryptoCode)
+    .then(() => walletId())
+    .then(({ walletId, index }) => cardanoApi.newAddress(walletId, index))
+}
+
+async function getStatus(account, toAddress, requested, cryptoCode) {
+  return checkCryptoCode(cryptoCode)
+    .then(() =>
+      Promise.all([
+        cardanoApi.addressBalance(toAddress, 1),
```
ppershing: is `1` a good constant for confirmed balance? I think it
1) might need to be more
2) should a named constant

lib/plugins/wallet/cardano/cardano.js
```diff
@@ -0,0 +1,85 @@
...
+        cardanoApi.addressBalance(toAddress, 0)
+      ])
+    )
+    .then(([confirmedBalance, totalBalance]) => {
+      if (confirmedBalance.gte(requested)) {
```
ppershing: Here I have question. Can we have *multiple* transactions for an address? If so, how this code should behave?

lib/plugins/wallet/cardano/cardano.js
```diff
@@ -0,0 +1,85 @@
...
+      return { status: 'notSeen' }
+    })
+}
+
+async function newFunding(account, cryptoCode) {
+  return checkCryptoCode(cryptoCode)
+    .then(() => walletId())
+    .then(({ walletId, index }) =>
+      Promise.all([
+        cardanoApi.walletBalance(walletId, 0),
```
ppershing: this is probably a bit redundant use of api as we could calculate both balances in one go

package-lock.json
```diff
@@ -16,18 +16,18 @@
...
-        "babel-plugin-transform-es2015-destructuring": "^6.19.0",
-        "babel-plugin-transform-es2015-function-name": "^6.9.0",
-        "babel-plugin-transform-es2015-modules-commonjs": "^6.18.0",
-        "babel-plugin-transform-es2015-parameters": "^6.21.0",
-        "babel-plugin-transform-es2015-spread": "^6.8.0",
-        "babel-plugin-transform-es2015-sticky-regex": "^6.8.0",
-        "babel-plugin-transform-es2015-unicode-regex": "^6.11.0",
-        "babel-plugin-transform-exponentiation-operator": "^6.8.0",
-        "package-hash": "^1.2.0"
+        "babel-plugin-check-es2015-constants": "6.22.0",
```
ppershing: It is interesting that package-lock changed here. Do you know which npm version they are using? Should we use the same?

lib/plugins/wallet/cardano/cardano.js
detrandix: yes, it's needed header for this "common" functions  (same for BTC, etc)

lib/blockchain/cardano.js
detrandix: it's used below in `setup` regexp for change default folder, but in this regexp we use `#`, so i build own escape function instead of using `_.escapeRegExp`

lib/blockchain/cardano.js
detrandix: ok

lib/blockchain/cardano.js
detrandix: my fault

lib/coin-utils.js
detrandix: as I wrote you, this I will change to mainet after everything else will be ok

lib/plugins/wallet/cardano/api.js
detrandix: ok, I will break this to two functions `fetch` and `fetchAllPages`

lib/plugins/wallet/cardano/cardano.js
detrandix: they used `1` for all BTC like cryptocurrencies, but we may change this - have you idea for good count of confirmations? Daedalus use 1-2 for low, 3-8 for medium and 9+ for high assurance. Constant is a good idea

lib/plugins/wallet/cardano/api.js
detrandix: :+1:

lib/plugins/wallet/cardano/api.js
detrandix: ok, that's make sense

lib/plugins/wallet/cardano/api.js
detrandix: maybe `sumBalanceByAddressAndCofirmations`, but I'm not really sure about that name... if this function will be local, maybe only `sumBatchBalance`?

lib/plugins/wallet/cardano/api.js
detrandix: ok, it will be also ok `minConfirmations`?

lib/plugins/wallet/cardano/cardano.js
detrandix: it's copied from other currencies code

lib/plugins/wallet/cardano/api.js
detrandix: i suppose it's exception from axios, but you are right, i will check this before

lib/plugins/wallet/cardano/cardano.js
detrandix: i copy this from other currencies, they always create Promise chain, so from this point of view i suppose that's make sense

lib/plugins/wallet/cardano/cardano.js
detrandix: maybe `walletBalance` can return total sum, and sum for min confirmed transactions - it's good solution for you?

lib/blockchain/cardano.js
izri16: If the only difference compared to lodash is that we need to escape extra `#`, what about calling `_.escapeRegExp` and we will do only escape '#' on the result of it? It might be safer than enumerating all escape characters ourselves.
Something like:
```
let res = _.escapeRegExp(str)
// Note: we also need to escape '#'
return res.replace(/#/g, '\\$&')
```

lib/blockchain/cardano.js
```diff
@@ -0,0 +1,47 @@
...
+
+const es = common.es
+
+// todo: maybe add '#' as second parameter?
+function escapeRegExp(str) {
+  return str.replace(/[.*+?^${}()|[\]\\#]/g, '\\$&')
+}
+
+function setup(dataDir) {
+  const coinRec = coinUtils.getCryptoCurrency('ADA')
```
izri16: Is `coinRec` shortcut of `coinRecord`? :thinking: If yes, could you use the longer it might remove possible confusion when someone sees it below in the code.

lib/blockchain/cardano.js
```diff
@@ -0,0 +1,47 @@
...
+  )
+
+  common.firewall([coinRec.defaultPort])
+
+  const config = buildConfig(null, null, null)
+  common.writeFile(path.resolve(dataDir, coinRec.configFile), config)
+
+  const seed = es('openssl rand -hex 16')
+  const mnemonic = mnemonicHelpers.fromSeed(seed.trim())
+  fs.writeFileSync(options.cardanoMnemonicPath, mnemonic)
```
izri16: Just for curiosity, is it common in the lamassu server that they use sync versions of functions?

lib/blockchain/cardano.js
```diff
@@ -0,0 +1,47 @@
...
+
+  const cmd = `/usr/local/bin/${coinRec.daemon}`
+  common.writeSupervisorConfig(coinRec, cmd)
+}
+
+function buildConfig(walletId, index, mnemonicHash) {
```
izri16: So those are always `null` as it seems to be called only as `(null, null, null`) ? :thinking:  Do we then need this abstraction?

lib/blockchain/common.js
```diff
@@ -99,7 +105,12 @@ function fetchAndInstall (coinRec) {
   es(`wget -q ${url}`)
   es(`tar -xzf ${downloadFile}`)
 
-  if (_.isEmpty(binaries.files)) {
+  if (binaries.build) {
```
izri16: Seems this was only used for debugging and will be removed right?

lib/plugins/wallet/cardano/api.js
```diff
@@ -0,0 +1,161 @@
...
+    this.port = port
+    this.hostname = hostname
+  }
+
+  async fetch(endpoint, params, data, appendData = []) {
+    const method = data ? 'POST' : 'GET'
+
+    const url = `https://${this.hostname}:${this.port}/api/v1/${endpoint}`
+
+    return Promise.resolve(true)
```
izri16: Is this needed? Is not `this.client` async by default?

lib/plugins/wallet/cardano/api.js
```diff
@@ -0,0 +1,161 @@
...
+      .then(() => {
+        return this.client({
+          method,
+          url,
+          params,
+          data
+        })
+      })
+      .then(response => {
+        if (response.error) throw r.error
```
izri16: `r.error` is probably typo as you did not define `r`

lib/plugins/wallet/cardano/api.js
izri16: I like the longer one. It is quite long however it would be clear what it does. Or maybe just `sumBalancForAddress(address, minConfirmations)`. So when you rename `confirmations` -> `minConfirmations` also this shorter variant might be clear enough.

lib/plugins/wallet/cardano/api.js
izri16: But if it fails it is fine right? As we would want to throw anyway then I guess.

lib/plugins/wallet/cardano/cardano.js
```diff
@@ -0,0 +1,85 @@
...
+}
+
+/**
+ * TODO: instead of using walletId function we can
+ *       inject walletId/index into account object
+ *       in wallet.js - it's better?
+ */
+async function balance(account, cryptoCode) {
+  return checkCryptoCode(cryptoCode)
+    .then(() => walletId())
```
izri16: you could do `.then(walletId)` if you find it nicer, also below

lib/plugins/wallet/cardano/cardano.js
izri16: So please just add some comment, explaining why it is not used, so in the future, there is no need to think about it.

package.json
```diff
@@ -31,6 +31,7 @@
     "got": "^7.1.0",
     "helmet": "^3.8.1",
     "inquirer": "^5.2.0",
+    "js-sha256": "^0.9.0",
```
izri16: Did you check if none of the packages that are already included does not support `sha256`? As it is quite common hash function.

ppershing:
I think I prefer full lambda. The reason is that you explicitly see
argument count and names

lib/plugins/wallet/cardano/mnemonic.js
```diff
@@ -0,0 +1,64 @@
...
+
+const coinUtils = require('../../../coin-utils')
+const options = require('../../../options')
+
+const cryptoRec = coinUtils.getCryptoCurrency('ADA')
+const configPath = coinUtils.configPath(cryptoRec)
+
+function getCardanoMnemonic() {
+  return fs
+    .readFileSync(options.cardanoMnemonicPath, 'utf8')
```
izri16: As `walletId` is async why this needs to be sync? Maybe some question applies also for other "sync" usage

lib/blockchain/cardano.js
detrandix: yes, only i change this to `common.writeFile`

lib/plugins/wallet/cardano/mnemonic.js
```diff
@@ -0,0 +1,64 @@
...
+function getWalletConfiguration() {
+  return JSON.parse(fs.readFileSync(configPath))
+}
+
+function saveWalletConfiguration(configuration) {
+  fs.writeFileSync(configPath, JSON.stringify(configuration))
+}
+
+async function walletId() {
+  let configuration = getWalletConfiguration()
```
izri16: There are a lot of things that can throw inside this function. Should not we catch for the errors or for some reason it is never needed and we just want to throw anyway?

lib/blockchain/cardano.js
detrandix: no, when i created this, i suppose we will use this more => removed

lib/blockchain/common.js
detrandix: no, `binaries.build` is new "command", and we need this for building cardano, but if you mean that `console.log` a few lines below, then yes, this must be removed :)

lib/plugins/wallet/cardano/api.js
detrandix: :+1:

lib/plugins/wallet/cardano/api.js
detrandix: :+1:

lib/plugins/wallet/cardano/mnemonic.js
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  return Promise.resolve({
+    walletId: configuration.walletId,
+    index: configuration.accountIndex
+  })
+}
+
+module.exports = {
+  walletId
```
izri16: I am not sure what should be the name of this module. Is `mnemonic` convention across the codebase? As it is true that this does operations on mnemonic however from the caller perspective it gives `walletId` which is the public key I suppose. And in general, this is lamassu `walletId` right? So maybe it also might make sense to call this module `lamassuWallet` although I am not really sure (and maybe I missed something). What do you think?

detrandix: `walletId` is something like a public key, but it isn't for Lamassu server, but for Cardano SL - we need this to work with our wallet over API - so maybe `cardanoWallet`? Because as you pointed, mnemonic "layer" is there only as support for wallet layer, so `wallet` in the name makes me sense.

lib/plugins/wallet/cardano/cardano.js
detrandix: i agree with ppershing about visible arguments

lib/blockchain/cardano-build.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
+
+if [ -z "$USER" ]; then
+    USER=`whoami`
+fi
+
+tmpdir=$(mktemp -d -t nix-XXXXXXXXXX)
```
izri16: Question: tmp folder is used to store files needed during installation but no longer after it passes?

detrandix: yes, because I don't wanna download nix installation files to any "our" folder

package.json
detrandix: yes, found alternative, thanks

lib/plugins/wallet/cardano/mnemonic.js
detrandix: yes, because most of them are fatal for us... but I will be happy to discuss about that

lib/blockchain/cardano.js
```diff
@@ -0,0 +1,41 @@
...
+  es(
+    `sudo sed -i 's#state-wallet-testnet/#${_.escapeRegExp(dataDir).replace(
+      /#/g,
+      '\\#'
+    )}/#g' /usr/local/bin/cardano-node`
+  )
+
+  common.firewall([coinSettings.defaultPort])
+
+  const defaultConfigData = {
```
detrandix, line 24: maybe add a comment why init data is empty (is empty because walletId & index will be fetched after first call API, now we are not sure that cardano node running)

detrandix: also maybe we can use functions from `lib/plugins/wallet/cardano/wallet.js` to have this (saving configuration) only on one place

lib/blockchain/cardano-build.sh
```diff
@@ -0,0 +1,18 @@
...
+cd $tmpdir
+wget -q https://nixos.org/nix/install
```
ppershing: They way the script is currently written, if any command fails, it will keep going with the next one. I think it shouldn't do that ...

lib/blockchain/cardano-build.sh
```diff
@@ -0,0 +1,18 @@
...
+chmod +x install
+OUTPUT=$(./install --no-daemon)
```
ppershing: why do we need `OUTPUT` ?

lib/blockchain/cardano.js
ppershing: An alternative -- can't we symlink `state-wallet-testnet` instead of rewriting the config?
Also, don't we have an "autoescaping" version of the shell launcher? Something like python's
`os.system(['sudo', 'sed', '-i', '.....', '/usr/...']) ?

lib/blockchain/cardano.js
```diff
@@ -0,0 +1,41 @@
...
+    index: null,
+    mnemonicHash: null
+  }
+  common.writeFile(path.resolve(dataDir, coinSettings.configFile), JSON.stringify(defaultConfigData))
+
+  const seed = es('openssl rand -hex 16')
+  const mnemonic = mnemonicHelpers.fromSeed(seed.trim())
+  common.writeFile(options.cardanoMnemonicPath, mnemonic)
+
+  const cmd = `/usr/local/bin/${coinSettings.daemon}`
```
ppershing: Now, this is interesting. Here you use `coinSettings.daemon` but up there on L20 you use hard-coded path

lib/blockchain/common.js
ppershing: I am not sure whether `build` is good enough name. I wanted to propose `buildCmd` but that would suggest that the argument is a string which it isn't ... :-(

lib/blockchain/common.js
```diff
@@ -99,7 +105,11 @@ function fetchAndInstall (coinRec) {
...
+    es(`${__dirname}/${binaries.build.script} \`realpath ${binaries.build.folder}\``)
```
ppershing: This is confusing as normally you would expect build script to be relative to `build.folder`, not `__dirname`. I understand why it is how it is but it is still unexpected.

lib/blockchain/common.js
```diff
@@ -99,7 +105,11 @@ function fetchAndInstall (coinRec) {
...
+  }
+
+  if (_.isEmpty(binaries.files) && binDir) {
```
ppershing: Why is there `&& binDir` now? Unless I am mistaken, this does not affect cardano at all but might affect other currencies?

lib/plugins/wallet/cardano/api.js
```diff
@@ -0,0 +1,223 @@
+const BN = require('../../../bn')
+
+const unitScale = 6
+
+class CardanoAPI {
+  static get INCLUDE_INCOME_TRANSACTIONS() {
+    return 1 >> 1
```
ppershing: These are certainly wrong. `1>>1` and `1>>2` is 0. Also, why is there need to have this as a function? a constant would be enough?

lib/plugins/wallet/cardano/api.js
```diff
@@ -0,0 +1,223 @@
...
+      }
+
+      return response.data.data
+    })
+  }
+
+  async fetchAllPages(endpoint, params, data) {
+    const fetchAllPagesAccumular = (page, dataAccumular = []) =>
+      this.fetch(endpoint, { ...params, page }, data, true).then(response => {
+        const responseData = Array.isArray(response.data.data)
```
ppershing: Are we sure there exists endpoints which
1) use paging
2) at the same time, won't return an array?

lib/plugins/wallet/cardano/wallet.js
```diff
@@ -0,0 +1,76 @@
...
+}
+
+async function getWalletConfiguration() {
+  return promisify(fs.readFile)(configPath)
+    .then(text => JSON.parse(text))
+    .catch(() => ({}))
+}
+
+async function saveWalletConfiguration(configuration) {
+  return fs.writeFile(configPath, JSON.stringify(configuration))
```
ppershing: uhuh, you are doing `promisify(fs.readFile)` but not `fs.writeFile`? Seems suspicious ...

lib/plugins/wallet/cardano/wallet.js
```diff
@@ -0,0 +1,76 @@
...
+}
+
+async function walletId() {
+  let configuration = await getWalletConfiguration()
+  const mnemonic = await getCardanoMnemonic()
+
+  if (
```
ppershing, line 58: I am not sure I like
1) that `walletId` might do some magic behind. Maybe `getOrCreateWalletId` would be better?
2) creating walletId from config should probably be separated into its own function

lib/plugins/wallet/cardano/client.js
```diff
@@ -0,0 +1,19 @@
...
+const https = require('https')
+const fs = require('fs')
+
+const coinUtils = require('../../../coin-utils')
+
+const cryptoRec = coinUtils.getCryptoCurrency('ADA')
+const cryptoDir = coinUtils.cryptoDir(cryptoRec)
+
+const httpsAgent = new https.Agent({
+  ca: fs.readFileSync(`${cryptoDir}/tls/client/ca.crt`),
```
ppershing: Can we assume that `cryptoDir` is a safe path? (I presume so but I want to check that an attacker cannot inject something like `../../` into it)

lib/plugins/wallet/cardano/cardano.js
```diff
@@ -0,0 +1,91 @@
...
+      toAddress,
+      MINIMUM_CONFIRMATIONS_FOR_APPROVE,
+      cardanoClient.INCLUDE_INCOME_TRANSACTIONS
+    )
+    .then(balance => {
+      if (balance.confirmed.gte(requested)) {
+        return { status: 'confirmed' }
+      } else if (balance.total.gte(request)) {
+        return { status: 'authorized' }
+      } else if (balance.confirmed.eq(balance.total)) {
```
ppershing: what is the difference between `insufficientFunds` and `notSeen` ? Importantly, what should we show with  `balance.confirmed == balance.total == 0`?

lib/plugins/wallet/cardano/cardano.js
```diff
@@ -0,0 +1,91 @@
...
+// TODO: work with BN here instead of api.js?
+
+function assertCryptoCode(cryptoCode) {
+  if (cryptoCode !== 'ADA') {
+    throw new Error(`Unsupported crypto: ${cryptoCode}`)
+  }
+}
+
+/**
+ * We don't use `account` parameter, because there is any interesting information for us,
```
ppershing: "the is any" .. typo?

lib/plugins/wallet/cardano/api.js
```diff
@@ -0,0 +1,223 @@
...
+        walletId
+      },
+      groupingPolicy: 'OptimizeForHighThroughput'
+    }).then(data => {
+      const inputs = data.inputs.reduce((acc, item) => acc + item.amount, 0)
+      const outputs = data.outputs.reduce((acc, item) => acc + item.amount, 0)
+
+      return {
+        fee: BN(inputs - outputs)
+          .abs()
```
ppershing: huh? why is there `abs` ? In fact, we should probably assert that `inputs >= outputs`

lib/plugins/wallet/cardano/api.js
```diff
@@ -0,0 +1,223 @@
...
+          address
+        }
+      ],
+      source: {
+        accountIndex,
+        walletId
+      },
+      groupingPolicy: 'OptimizeForHighThroughput'
+    }).then(data => {
+      const inputs = data.inputs.reduce((acc, item) => acc + item.amount, 0)
```
ppershing: this seems fishy. Shouldn't we use bignumbers already here?

lib/plugins/wallet/cardano/api.js
```diff
@@ -0,0 +1,223 @@
...
+            {
+              confirmed: 0,
+              total: 0
+            }
+          )
+      )
+      .then(({ confirmed, total }) => ({
+        confirmed: BN(confirmed)
+          .shift(unitScale)
+          .round(), // TODO: is that shift really need?
```
ppershing: I am more worried about `round` -- why would we need to do that? Shouldn't we have exact result here?

lib/plugins/wallet/cardano/api.js
```diff
@@ -0,0 +1,223 @@
...
+            }
+
+            return {
+              isConfirmed: transaction.confirmations >= minConfirmations,
+              amount
+            }
+          })
+          .reduce(
+            (acc, item) => {
+              acc.confirmed += item.isConfirmed ? item.amount : 0
```
ppershing: please, be more functional and do not modify object. Instead, just return new one

lib/plugins/wallet/cardano/api.js
```diff
@@ -0,0 +1,223 @@
...
+  static get INCLUDE_ALL_TRANSACTIONS() {
+    CardanoAPI.INCLUDE_INCOME_TRANSACTIONS |
+      CardanoAPI.INCLUDE_OUTCOME_TRANSACTIONS
+  }
+
+  static get PER_PAGE_MAXIMUM() {
+    return 50
+  }
+
+  constructor(client, port, hostname = 'localhost') {
```
ppershing: I think we can easily separate
"fetcher" object from api itself.

lib/plugins/wallet/cardano/api.js
```diff
@@ -0,0 +1,223 @@
...
+  async fetchAllPages(endpoint, params, data) {
+    const fetchAllPagesAccumular = (page, dataAccumular = []) =>
+      this.fetch(endpoint, { ...params, page }, data, true).then(response => {
+        const responseData = Array.isArray(response.data.data)
+          ? response.data.data
+          : [response.data.data]
+        const result = [...responseData, ...dataAccumular]
+
+        if (
+          response.data.meta.pagination.totalPages >
```
ppershing: Is there a `totalItems` count in the response? If so, we should assert that it won't change during "iteration"

lib/plugins/wallet/cardano/api.js
```diff
@@ -0,0 +1,223 @@
...
+      if (returnFullResponse) {
+        return response
+      }
+
+      return response.data.data
+    })
+  }
+
+  async fetchAllPages(endpoint, params, data) {
+    const fetchAllPagesAccumular = (page, dataAccumular = []) =>
```
ppershing: why don't we make use of `await` keyword?
I think we could code this as
```
let page = 1
let result = []
while (true) {
  res = await fetch(..., page)
  result = [...result, ...data]
  assert(totalPages >= page)
  if (page == totalPages) return result
  page++
}
```

lib/plugins/wallet/cardano/api.js
detrandix: Fixed to `1 << 0` & `1 << 1`. JS classes don't have constants - we can define consts outside this classes, but I want a connection between them. I can do this by `CardanoAPI.INCLUDE_INCOME_TRANSACTIONS = 1 << 0`, but this isn't constant at all.

lib/plugins/wallet/cardano/api.js
detrandix: Yes, pagination is in all responses - f.e. https://cardanodocs.com/technical/wallet/api/v1/?v=1.6.0#tag/Info - i don't suppose anybody will run `fetchAllPages` on endpoints with non-array data, but i take this as edge case

lib/plugins/wallet/cardano/wallet.js
detrandix: yes, because i suppose `async` make this for me - but i will test this and in case of failury i will change this

lib/plugins/wallet/cardano/wallet.js
detrandix: :+1:

lib/plugins/wallet/cardano/client.js
detrandix: Crypto dir is composed by Lamassu server itself, so if it will not be safe, then attacker is on server.

```
function cryptoDir (cryptoRec) {
  const code = cryptoRec.code
  return path.resolve(blockchainDir(), code)
}
```

&

```
function blockchainDir () {
  return options.blockchainDir
}
```

so there are two possible ways - `cryptoRec.code` which is get from `coin-utils.js` file, and `options.blockhainDir` which is get form `~/.lamassu/lamassu.json` - for me it seems that this path is safe (if server is safe).

lib/plugins/wallet/cardano/cardano.js
detrandix: yes

lib/plugins/wallet/cardano/api.js
detrandix: yes, this assertion is right, removed

lib/plugins/wallet/cardano/api.js
detrandix: and if it is changed, then what - start's again from the beginning? because this seems as only possible way for me

lib/blockchain/cardano-build.sh
detrandix: old thing when i parsed result of installation, can be removed

lib/plugins/wallet/cardano/api.js
ppershing: Then just assert. I would rather have defensive code than one multi-purpose blob

lib/plugins/wallet/cardano/wallet.js
ppershing: `async` will not do automatic promisify imho

lib/plugins/wallet/cardano/api.js
ppershing: I think simply `static CAPS_NAMED_VAR=value` should be enough as a "constant", no?

lib/plugins/wallet/cardano/api.js
ppershing: Just fail the whole call. The caller should be able to deal with failures anyway

lib/plugins/wallet/cardano/api.js
detrandix: ou, static properties works in node :+1: sorry, i was confused, becase in browser this now only works with transpilers

lib/plugins/wallet/cardano/api.js
detrandix: ok

lib/blockchain/common.js
detrandix:
```
  ADA: {
    url: 'https://github.com/input-output-hk/cardano-sl/archive/3.1.0.tar.gz',
    build: {folder: 'cardano-sl-3.1.0', script: 'cardano-build.sh'},
    dir: 'cardano-sl-3.1.0',
    files: [['cardano-testnet-wallet', 'cardano-node']]
  }
```

ok - when build will be using `dir` param, then we don't need `folder` and it can be `buildCmd` then.. do you agree?

lib/plugins/wallet/cardano/cardano.js
detrandix: insufficientFunds is for total > 0 (include pending), notSeen for 0, will update that

lib/blockchain/common.js
detrandix: I know, but have you any solution? That installation process is runned in `/tmp` folder where is downloaded Cardano SL source code, but i wanna to run our build script which is in Lamassu server folder...

lib/blockchain/cardano.js
detrandix: solved

lib/plugins/wallet/cardano/api/api.js
```diff
@@ -0,0 +1,174 @@
...
+    return this.fetcher
+      .fetch('addresses', null, {
+        accountIndex: index,
+        walletId
+      })
+      .then(data => data.id)
+  }
+}
+
+CardanoApi.INCLUDE_INCOME_TRANSACTIONS = 1;
```
ppershing: inconsistent use of semi-colons

lib/plugins/wallet/cardano/cardano.js
```diff
@@ -0,0 +1,101 @@
...
+
+  return getOrCreateWalletCredentials().then(({ walletId, index }) =>
+    cardanoClient.newAddress(walletId, index)
+  )
+}
+
+async function getStatus(account, toAddress, requested, cryptoCode) {
+  assertCryptoCode(cryptoCode)
+
+  // we include only income transactions, because we need only about
```
ppershing: Some weird sentence here: "we need only about"

lib/blockchain/cardano.js
detrandix: `state-wallet-testnet is symlink` to something like /nix/5435345/353.... - so i really copy this originall file - don't forget that we compile that build in /tmp folder...

but what i don't understand - where is advantage of that symlink? you still need to rewrite that datadir paths... 

yes, they have spawn method, i try to use this.

lib/plugins/wallet/cardano/wallet.js
```diff
@@ -0,0 +1,67 @@
...
+      hash: sha256(mnemonicText)
+    }))
+}
+
+async function getWalletConfiguration() {
+  return promisify(fs.readFile)(configPath).then(text => JSON.parse(text))
+}
+
+async function saveWalletConfiguration(configuration) {
+  return fs.writeFile(configPath, JSON.stringify(configuration))
```
izri16: I have seen usage of `common.writeFile` in some other modules can not (should not) it be used everywhere?

lib/blockchain/cardano.js
```diff
@@ -10,6 +10,12 @@ module.exports = { setup }
 
 const es = common.es
 
+function from24to12wordMnemonic(mnemonic) {
+  const entrophy = mnemonicHelpers.toEntropyBuffer(mnemonic).toString('hex')
+  const halfEntrophy = entrophy.substring(0, 32)
```
izri16: Not sure that `halfEntrophy` is the best name.  Maybe you can just directly use it as `mnemonicHelpers.fromSeed(entrophy.subscring(0, 32))`

lib/blockchain/cardano.js
```diff
@@ -0,0 +1,52 @@
...
+const coinUtils = require('../coin-utils')
+const options = require('../options')
+const _ = require('lodash')
+
+module.exports = { setup }
+
+const es = common.es
+
+function from24to12wordMnemonic(mnemonic) {
+  const entrophy = mnemonicHelpers.toEntropyBuffer(mnemonic).toString('hex')
```
ppershing: typo, should be `entropy`

lib/plugins/wallet/cardano/api/api.js
```diff
@@ -0,0 +1,167 @@
...
+            }
+          )
+      )
+  }
+
+  sumAmountByAddress(movements, address) {
+    return movements
+      .filter(movement => movement.address === address)
+      .map(movement => BN(movement.amount))
+      .reduce((acc, amount) => acc.add(amount), BC(0))
```
ppershing: isn't `BC` a typo ?

CrypticaScriptura:
Thanks for the precompiled binary! :+1:

However, upon running `lamassu-coins` with the latest changes, I run into the following error:

```
root@cardano-test:~# lamassu-coins
? Which cryptocurrencies would you like to install? Cardano
Thanks! Installing: cardano. Will take a while...
Volume is already mounted.
Installing Cardano...
(node:21170) UnhandledPromiseRejectionWarning: TypeError: path must be a string or Buffer
    at Object.fs.openSync (fs.js:646:18)
    at Object.fs.writeFileSync (fs.js:1299:33)
    at Object.writeFile (/usr/lib/node_modules/lamassu-server/lib/blockchain/common.js:130:8)
    at Object.setup (/usr/lib/node_modules/lamassu-server/lib/blockchain/cardano.js:39:10)
    at setupCrypto (/usr/lib/node_modules/lamassu-server/lib/blockchain/install.js:88:16)
    at r (/usr/lib/node_modules/lamassu-server/node_modules/lodash/lodash.min.js:5:356)
    at Function.Xe (/usr/lib/node_modules/lamassu-server/node_modules/lodash/lodash.min.js:67:74)
    at Function.l (/usr/lib/node_modules/lamassu-server/node_modules/lodash/lodash.min.js:50:137)
    at processCryptos (/usr/lib/node_modules/lamassu-server/lib/blockchain/install.js:59:5)
    at inquirer.prompt.then.answers (/usr/lib/node_modules/lamassu-server/lib/blockchain/install.js:123:22)
(node:21170) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:21170) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
root@cardano-test:~#
```

Steps to replicate on Ubuntu Server 16.04.6 LTS x64:

```
curl -sS https://raw.githubusercontent.com/lamassu/lamassu-install/defiant-dingirma/install | bash
supervisorctl stop all
cd /usr/lib/node_modules/
mv lamassu-server lamassu-server-orig
git clone https://github.com/lamassu/lamassu-server.git
cd lamassu-server
git fetch origin pull/321/head:pr-321
git checkout pr-321
npm install
supervisorctl start all
lamassu-coins
```

CrypticaScriptura:
Thanks, @detrandix. I see the install script changes now.

We currently use [lamassu-install/install](https://github.com/lamassu/lamassu-install/blob/defiant-dingirma/install) instead of lamassu-server/lamassu-remote-install/install.

I deployed as above, but with your changes:

https://github.com/lamassu/lamassu-install/blob/dd-cardano/install

`lamassu-coins` is now successful and the Cardano node is running, however, the admin's 'Maintenance > Funding > Cardano' tab shows `NetworkError` and `/var/log/supervisor/lamassu-admin-server.err.log` reports:

```
Error: Cannot find module '../../common'
    at Function.Module._resolveFilename (module.js:548:15)
    at Function.Module._load (module.js:475:25)
    at Module.require (module.js:597:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/usr/lib/node_modules/lamassu-server/lib/plugins/wallet/cardano/wallet.js:7:16)
    at Module._compile (module.js:653:30)
    at Object.Module._extensions..js (module.js:664:10)
    at Module.load (module.js:566:32)
    at tryModuleLoad (module.js:506:12)
    at Function.Module._load (module.js:498:3)
```

MichalPetro:
Hi @naconner, we had a bad path to commoj.js library which is fixed now, sorry for that. Please update lamassu server source code. Tomas also made some fix for the transactions list in ELM.

CrypticaScriptura:
Thanks @MichalPetro. Cash-in transactions are working. :+1:  Have you gotten success with cash-out transactions?

I'm getting this in `lamassu-server.err.log` when attempting (unsuccessfully) to generate an address on the machine's cash-out deposit screen:

```
2020-02-29T16:04:42.339Z - error:  message=cryptoCode is not defined, stack=ReferenceError: cryptoCode is not defined
    at Object.newAddress (/usr/lib/node_modules/lamassu-server/lib/plugins/wallet/cardano/cardano.js:42:20)
    at fetchWallet.then.r (/usr/lib/node_modules/lamassu-server/lib/wallet.js:81:25)
    at <anonymous>
```

MichalPetro:
Hi @naconner - the truth is that we originally planned the integration only for the cash-in machines and we didnt test the cash-out function thoroughly. I suppose this will need just some small adjustment as the code was changed appropriately. We took a look on it but we are unable to setup some virtual fiat on the lamassu machine in order to test the cash-out functionality properly, can you advise how to configure it?

EDIT: Disregard the above please, we got it, the above issue was fixed, we will do some last testing and let you know

MichalPetro:
@naconner this and all other stuff related to cash-out was fixed now

josepfo:
[Cardano - Add support](https://trello.com/c/uPeO8vAh/1069-cardano-add-support)

PR #329, opened 2019-11-12
Update crypto node versions, add Dash to Coinbase ticker

CrypticaScriptura:


CrypticaScriptura:
[Update crypto node versions, add Dash to Coinbase ticker](https://trello.com/c/fhMjpMcq/97-update-crypto-node-versions-add-dash-to-coinbase-ticker)

PR #331, opened 2019-11-15
Fix ETH sending status

CrypticaScriptura:
Fixes condition where transaction goes through, but history stays as 'pending'.


CrypticaScriptura:
[ETH cash-in: 'BN(...)multipliedBy is not a function'](https://trello.com/c/AT4V8qTb/100-eth-cash-in-bnmultipliedby-is-not-a-function)

RafaelTaranto (approved):

PR #333, opened 2019-11-20
fix: operator id fetching on scripts

RafaelTaranto:


CrypticaScriptura:
[Coin ATM Radar module reports global commission values, instead of that of individual machines](https://trello.com/c/x8myB5em/9-coin-atm-radar-module-reports-global-commission-values-instead-of-that-of-individual-machines)

PR #336, opened 2019-11-25
feat: transactions page

lcsfelix:


new-lamassu-admin/src/components/table/TableRow.js
```diff
@@ -24,7 +24,7 @@ const useStyles = makeStyles({
     boxShadow: '0 0 4px 0 rgba(0, 0, 0, 0.08)'
   },
   lg: {
-    extend: info2,
+    // extend: info2,
```
RafaelTaranto: Any reason why this was commented out?

new-lamassu-admin/src/components/LogsDownloaderPopper.js
```diff
@@ -132,36 +131,26 @@ const LogsDownloaderPopover = ({ id, open, anchorEl, onClose, logsResponse, ...p
...
 
-  const formatDateFile = date => {
-    return moment(date).format('YYYY-MM-DD_HH-mm')
-  }
-
   const handleRadioButtons = (event) => {
-    setRadioButtons(toInteger(event.target.value))
+    const radio = toInteger(event.target.value)
+    setRadioButtons(radio)
+    if (radio === 0) setRange({ from: null, to: null })
```
RafaelTaranto: why not just use boolean?

new-lamassu-admin/src/components/expandable-table/ExpTable.js
```diff
@@ -28,7 +28,7 @@ const ExpRow = ({ id, columns, details, sizes, expanded, className, expandRow, .
 
   const detailsRowClasses = {
     [classes.detailsRow]: true,
-    [classes.hideDetailsRow]: expanded
+    [classes.hideDetailsRow]: !expanded
```
RafaelTaranto: Prevent the component from being mounted instead of using a class 

`{expanded && (<Tr....)`

new-lamassu-admin/src/components/expandable-table/ExpTable.js
```diff
@@ -28,7 +28,7 @@ const ExpRow = ({ id, columns, details, sizes, expanded, className, expandRow, .
...
   }
 
   return (
```
RafaelTaranto, line 28: Why are you using uuidv1 on all the components here? You assign key for the components only for the react internals, rule of thumb is using `idx`. If you generate the key every time it will be as if you're handing a new component every time.

RafaelTaranto: Also I don't even see uuid lib on package.json, so it wouldn't work on production.

new-lamassu-admin/src/components/expandable-table/ExpTable.js
```diff
@@ -41,7 +41,8 @@ const ExpRow = ({ id, columns, details, sizes, expanded, className, expandRow, .
           <button onClick={() => expandRow(id)} className={classes.expandButton}>
             {expanded && <ExpandOpenIcon />}
             {!expanded && <ExpandClosedIcon />}
-          </button></Td>
+          </button>
+        </Td>
       </Tr>
       <Tr className={classnames(detailsRowClasses)}>
         <Td size={mainWidth}>
```
RafaelTaranto: rowStates logic is awfully complex here, if we need to keep only one row open why not just have a id state with the open component? Then we don't have to keep creating a new list and mapping everything on that list to false.

new-lamassu-admin/src/components/expandable-table/ExpTable.js
```diff
@@ -28,7 +28,7 @@ const ExpRow = ({ id, columns, details, sizes, expanded, className, expandRow, .
```
RafaelTaranto, line 27: This component has some serious performance issues, I'll try to lay it all out here. Some comments are not on the right place because I can only review the current PR

new-lamassu-admin/src/pages/ServerLogs.js
```diff
@@ -108,11 +109,34 @@ const Logs = () => {
...
+      return
+    }
+
+    if (range.from && range.to) {
+      const text = logs.filter((log) => moment(log.timestamp).isBetween(range.from, range.to, 'day', '[]')).map(it => JSON.stringify(it)).join('\n')
+      const blob = new window.Blob([text], {
+        type: 'text/plain;charset=utf-8'
+      })
+      FileSaver.saveAs(blob, `${formatDateFile(range.from)}_${formatDateFile(range.to)}_server`)
+    }
```
RafaelTaranto: I've seen this logic on another place, no way to move it to the base component?

new-lamassu-admin/src/pages/Transactions/Address.js
```diff
@@ -0,0 +1,73 @@
...
+import Popover from '../../components/Popper'
+import { ReactComponent as CopyIcon } from '../../styling/icons/action/copy/copy.svg'
+import { comet } from '../../styling/variables'
+
+const Address = ({ className, type, children, ...props }) => {
+  const [anchorEl, setAnchorEl] = useState(null)
+
+  useEffect(() => {
+    setTimeout(() => setAnchorEl(null), 3000)
+  }, [anchorEl])
```
RafaelTaranto: Why is this necessary?

new-lamassu-admin/src/pages/Transactions/tx.js
```diff
@@ -0,0 +1,12 @@
+function toUnit (cryptoAtoms, cryptoCode) {
+  const cryptoRec = {
+    cryptoCode: 'BTC',
+    display: 'Bitcoin',
+    code: 'bitcoin',
+    unitScale: 8
+  }
```
RafaelTaranto: Why only btc here?

RafaelTaranto (changes requested):

new-lamassu-admin/src/components/LogsDownloaderPopper.js
lcsfelix: The RadioGroup component is prepared for more than 2 options. Using the `value` identifier even when only two options are used is more consistent.

new-lamassu-admin/src/pages/Transactions/Address.js
lcsfelix: This auto closes the 'Copied to clipboard!' popper after 3s. The Material-Ui Popper uses a toggle anchor, so the user would have to click the Copy button again to close the popper. I think it's also a nice touch.

new-lamassu-admin/src/components/table/TableRow.js
lcsfelix: Hum... I'm sure there was, but now uncommenting doesn't seem to make a difference, so I'll revert it.

new-lamassu-admin/src/components/LogsDownloaderPopper.js
RafaelTaranto: better to have name values, then.

new-lamassu-admin/src/components/table/TableRow.js
joshmh: Why not remove it, in that case?

new-lamassu-admin/src/components/LogsDownloaderPopper.js
lcsfelix: Fixed in https://github.com/lamassu/lamassu-server/pull/336/commits/58c08e39d6c46210864dc97c5ae25ec7fae489b7

new-lamassu-admin/src/components/LogsDownloaderPopper.js
```diff
@@ -121,53 +121,71 @@ const styles = {
...
-    return moment(date).format('YYYY-MM-DD_HH-mm')
+    [classes.dateRangePickerShowing]: selectedRadio === radioButtonOptions[1].value,
+    [classes.dateRangePickerHidden]: selectedRadio === radioButtonOptions[0].value
   }
 
   const handleRadioButtons = (event) => {
-    setRadioButtons(toInteger(event.target.value))
+    const radio = event.target.value
+    setSelectedRadio(radio)
   }
```
RafaelTaranto: this can be reduced to `const handleRadioButtons = compose(setSelectedRadio, get('target.value'))`

new-lamassu-admin/src/components/LogsDownloaderPopper.js
```diff
@@ -121,53 +121,71 @@ const styles = {
 
 const useStyles = makeStyles(styles)
 
-const LogsDownloaderPopover = ({ id, open, anchorEl, onClose, logsResponse, ...props }) => {
-  const [radioButtons, setRadioButtons] = useState(0)
+const LogsDownloaderPopover = ({ id, name, open, anchorEl, getTimestamp, logs, title, ...props }) => {
+  const radioButtonOptions = [{ label: 'All logs', value: 'all' }, { label: 'Date range', value: 'range' }]
```
RafaelTaranto: Maybe create consts for `all` and `range`, that way you can reference that in code instead of `radioButtonOptions[0].value`.

new-lamassu-admin/src/components/expandable-table/ExpTable.js
```diff
@@ -28,26 +27,29 @@ const ExpRow = ({ id, columns, details, sizes, expanded, className, expandRow, .
 
   const detailsRowClasses = {
     [classes.detailsRow]: true,
-    [classes.hideDetailsRow]: expanded
+    [classes.hideDetailsRow]: !expanded
```
RafaelTaranto: We're not using this object anymore

new-lamassu-admin/src/pages/Transactions/DetailsCard.js
```diff
@@ -0,0 +1,209 @@
...
+              </div>
+              <div className={classes.commissionWrapper}>
+                <Label>Commission</Label>
+                <div>{`${commission} ${tx.fiatCode} (${commissionPercentage * 100} %)`}</div>
+              </div>
+            </div>
+          </div>
+          <div className={classnames(classes.col, classes.col3)}> {/* Column 3 */}
+            <div className={classnames(classes.innerRow)}>
+              <div style={{ height: 43.4 }}>
```
RafaelTaranto, line 152: we should prevent using sytles, why is this div empty?

new-lamassu-admin/src/pages/Transactions/DetailsCard.js
```diff
@@ -0,0 +1,209 @@
...
+                      ${addr.slice(5, 10)}
+                      ${addr.slice(10, 15)}
+                      ${addr.slice(15, 20)}
+                      ${addr.slice(20, 25)}
+                      ${addr.slice(25, 30)}
+                      ${addr.slice(30, 35)}
+                      ${addr.slice(35, 40)}
+                      ${addr.slice(40, 45)}
+                      ${addr.slice(45, 50)}`}
+                  </Address>
```
RafaelTaranto: Shouldn't the address component handle the output? Also this won't work for every address, check out the `formatAddress` function on `Funding.js`

new-lamassu-admin/src/pages/Transactions/DetailsCard.js
```diff
@@ -0,0 +1,209 @@
...
+          <div className={classnames(classes.col, classes.col2)}> {/* Column 2 */}
+            <div className={classes.innerRow}>
+              <div>
+                <Label>Transaction ID</Label>
+                <div>
+                  <Address type='txId'>
+                    {txHash &&
+                      `${txHash.slice(0, 41)}
+                      ${txHash.slice(41)}`}
+                  </Address>
```
RafaelTaranto: Why is this using the address component? If it's for both address and txhash shouldn't it be called something else? Something like copytoclipboard or the like.

As far as formatting goes I think it's better to make the div split the hash with break lines instead of doing it by hand.

new-lamassu-admin/src/pages/Transactions/DetailsCard.js
```diff
@@ -0,0 +1,209 @@
...
+          </div>
+          <div className={classnames(classes.col, classes.col3)}> {/* Column 3 */}
+            <div className={classes.innerRow}>
+              <div>
+                <Label>Session ID</Label>
+                <Address type='sessionId'>
+                  {tx.id &&
+                    `${tx.id.slice(0, 21)}
+                  ${tx.id.slice(21)}`}
+                </Address>
```
RafaelTaranto: Same as above

new-lamassu-admin/src/pages/Transactions/Transactions.js
```diff
@@ -0,0 +1,206 @@
...
+      textAlign: 'center'
+    },
+    {
+      value: ''
+    }
+  ]
+
+  const rows = txResponse && txResponse.data.map(tx => {
+    const customerName = tx.customerName ? tx.customerName
+      : (tx.customerIdCardData ? `${startCase(lowerCase(tx.customerIdCardData.firstName.slice(0, 1)))}. ${startCase(lowerCase(tx.customerIdCardData.lastName))}` : tx.customerPhone)
```
RafaelTaranto: This is complicated to read, split into a couple of functions.

new-lamassu-admin/src/pages/Transactions/Transactions.js
```diff
@@ -0,0 +1,206 @@
...
+    180, // Machine
+    162, // Customer
+    110, // Cash
+    141, // Crypto
+    136, // Address
+    124, // Date
+    124, // Time
+    90, // Trade
+    71 // Expand
+  ]
```
RafaelTaranto: I think it makes more sense to pass down the sizes with the main col object

```
 {
          value: moment(tx.created).format('HH:mm:ss'),
          textAlign: 'right',
          size: 12398612
        },
```

new-lamassu-admin/src/pages/Transactions/Transactions.js
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    if (range.from && range.to) {
+      const text = txs.filter((tx) => moment(tx.created).isBetween(range.from, range.to, 'day', '[]')).map(it => JSON.stringify(it)).join('\n')
+      const blob = new window.Blob([text], {
+        type: 'text/plain;charset=utf-8'
+      })
+      FileSaver.saveAs(blob, `${formatDateFile(range.from)}_${formatDateFile(range.to)}_transactions`)
+    }
+  }
```
RafaelTaranto: Didn't we abstracted this away to a component?

RafaelTaranto (changes requested):

new-lamassu-admin/src/pages/Transactions/CopyToClipboard.js
```diff
@@ -0,0 +1,89 @@
...
+    return address.replace(/(.{5})/g, '$1 ')
+  }
+
+  const formatTxId = (id = '') => {
+    return id.replace(/(.{41})/g, '$1 ')
+  }
+
+  const formatSessionId = (id = '') => {
+    return id.replace(/(.{21})/g, '$1 ')
+  }
```
RafaelTaranto: Still not what I meant. This component should do what it says it does, Copy to Clipboard. Formatting should be handled externally. Format address is something that we need, but the other two are just line breaking at the end of the div size, so you should make it break automatically instead of using a function for this.

new-lamassu-admin/src/pages/Transactions/Transactions.js
```diff
@@ -0,0 +1,159 @@
...
+  const { response: txResponse } = useAxios({
+    url: 'http://localhost:8070/api/txs/',
+    method: 'GET',
+    trigger: []
+  })
+
+  const formatCustomerName = (customer) => {
+    const { firstName, lastName } = customer
+
+    return `${startCase(lowerCase(firstName.slice(0, 1)))}. ${startCase(lowerCase(lastName))}`
```
RafaelTaranto: Why not do this:

```startCase(lowerCase(`${firstName.slice(0, 1)}. ${lastName}`))```

File: new-lamassu-admin/src/pages/Transactions/Transactions.js [hunk 22]
```diff
@@ -0,0 +1,159 @@
...
+        },
+        {
+          name: 'Crypto',
+          value: `${toUnit(new BigNumber(tx.cryptoAtoms), tx.cryptoCode).toFormat(5)} ${tx.cryptoCode}`,
+          textAlign: 'right',
+          size: 141
+        },
+        {
+          name: 'Address',
+          value: `${tx.toAddress.slice(0, 9)}...`,
```
RafaelTaranto: this can be done with css as well, it's usually better since we don't have to change here if we ever need to change this size.

new-lamassu-admin/src/pages/Transactions/Transactions.js
```diff
@@ -0,0 +1,159 @@
...
+          size: 136
+        },
+        {
+          name: 'Date (UTC)',
+          value: moment(tx.created).format('YYYY-MM-D'),
```
RafaelTaranto: does moment set UTC correctly?

RafaelTaranto (changes requested):

new-lamassu-admin/src/pages/Transactions/Transactions.js
lcsfelix: `startCase` removes the dot after the first name.

new-lamassu-admin/src/pages/Transactions/Transactions.js
lcsfelix: I assumed the date was already stored in the DB in the proper format, but I can use `moment.utc` to make sure.

lcsfelix:
[Create transactions page](https://trello.com/c/iEaZLvTE/119-create-transactions-page)

lcsfelix:
[Create transactions page](https://trello.com/c/iEaZLvTE/119-create-transactions-page)

new-lamassu-admin/src/pages/Transactions/Transactions.js
lcsfelix: https://github.com/lamassu/lamassu-server/pull/336/commits/7ecd04765289b4680c8ff903ecee0d0da44af8e6 removes formatting the last name altogether, to prevent misformatting names with special characters, such as [Double-barreled names](https://en.wikipedia.org/wiki/Double-barrelled_name).

RafaelTaranto (approved):
lgtm

RafaelTaranto:
Oh wait, this should be opened to dev.

PR #339, opened 2019-12-02
feat: machine status page WIP feedback needed

mautematico:
This PR uses ExpandableTable, and reuses Transactions styles to avoid repeating, so it depends on #336 (already rebased).

- adds some machine details to /api/machines endpoint
- exposes machines/:machineId/actions/:actions on the API
- adds a couple of missing icons
- adds expandableTable to show machine status
- registers machine-status on the router

TODO:
- add onClick to action buttons; I need feedback on this:
- expose 'shutdown' action
- expose machine 'error statuses'

> React Hook "useAxios" is called in 'reboot' function which is neither a React function component or a custom React Hook function


new-lamassu-admin/src/pages/maintenance/MachineDetailsCard.js
```diff
@@ -47,7 +58,7 @@ const MachineDetailsRow = ({ machine, ...props }) => {
                     {/* <div className={classes.commissionWrapper}> */}
                     <Label>Lamassu Support article</Label>
                     <div>
-                      {machine.statuses.map((status, index) => <span key={index} />)}
+                      {machine.statuses.map((_status, index) => <span key={index} />)}
```
RafaelTaranto, line 61: I think you can use `(...[, index]) => ` to avoid naming status

new-lamassu-admin/src/pages/maintenance/MachineDetailsCard.js
```diff
@@ -128,6 +143,7 @@ const MachineDetailsRow = ({ machine, ...props }) => {
                         color='primary'
                         Icon={ShutdownIcon}
                         InverseIcon={ShutdownReversedIcon}
+                        onClick={reboot}
```
RafaelTaranto, line 146: shouldn't this be shutdown?

new-lamassu-admin/src/components/Status.js
```diff
@@ -0,0 +1,89 @@
...
+const orange = theme => ({
+  root: {
+    backgroundColor: mistyRose,
+    borderRadius: 4,
+    margin: theme.spacing(0.5, 0.25),
+    height: 18
+  },
+  label: {
+    fontSize: smallestFontSize,
+    color: '#ff7311',
```
RafaelTaranto, line 34: Don't we have a named color for this? We should.

We might not need the three themed objects, I'm thinking we can have one and a class to change label color.

new-lamassu-admin/src/pages/maintenance/MachineDetailsCard.js
```diff
@@ -0,0 +1,148 @@
...
+                    <Label>Lamassu Support article</Label>
+                    <div>
+                      {machine.statuses.map((status, index) => <span key={index} />)}
+                    </div>
+                  </div>
+                </div>
+              </div>
+            </div>
+          </div>
+          <div className={classnames(classes.col, classes.col2)} style={{ background: '#ebefff', width: '2px' }} />
```
RafaelTaranto, line 68: same thing with the color here.

RafaelTaranto: also we should prevent using the "style" prop, can't we use a class here?

new-lamassu-admin/src/pages/maintenance/MachineDetailsCard.js
```diff
@@ -0,0 +1,148 @@
...
+                <div className={classes.innerRow}>
+                  <div>
+                    <Label>Paired at</Label>
+                    <div>{moment(machine.pairedAt).format('YYYY-MM-DD HH:mm:ss')}</div>
+                  </div>
+                  <div className={classes.commissionWrapper}>
+                    <Label>Software update</Label>
+                    <div className={classes.innerRow}>
+                      {machine.softwareVersion}
+                      &nbsp;
```
RafaelTaranto, line 96: We should prevent using this espaced sequence, we can do the same thing with css.

new-lamassu-admin/src/pages/maintenance/MachineStatus.js
```diff
@@ -0,0 +1,71 @@
...
+  }))
+
+  const sizes = [
+    232, // Machine Name
+    349, // Status
+    192, // Last ping
+    155, // Ping Time
+    201, // Software Version
+    71 // Expand
+  ]
```
RafaelTaranto, line 53: Luis will upload a new way of using this component shortly

RafaelTaranto (changes requested):

mautematico:
closed in favor of https://github.com/lamassu/lamassu-server/pull/344

PR #348, opened 2019-12-23
Implement alphapoint integration

akostiv:
TODO:
- add `alphapoint` as dependency after it will be public.
- change api url for tickers to prod env (`ticker/alphapoint.js`)


joshmh:
Please see the README, we don't accept outside pull requests.

PR #354, opened 2020-01-15
feat: add operator info page

lcsfelix:


lcsfelix:
[Create operator info screen](https://trello.com/c/0ukJnAUj/152-create-operator-info-screen)

lcsfelix:
I need to add error feedback before merging.

new-lamassu-admin/src/components/inputs/base/TextInput.js
```diff
@@ -60,7 +110,7 @@ const TextInput = memo(
           ) : null
         }}
         InputLabelProps={{ className: classes.labelRoot }}
-        {...props}
+        {...R.omit(['className'], props)}
```
RafaelTaranto: Why is className getting ommited and not used on the main parent? As it stands this prop does nothing.

new-lamassu-admin/src/pages/OperatorInfo/ContactInfo.js
```diff
@@ -0,0 +1,306 @@
...
+const Field = ({ editing, field, displayValue, ...props }) => {
+  const classes = fieldUseStyles()
+
+  const classNames = {
+    [classes.field]: true,
+    [classes.notEditing]: !editing
+  }
+
+  return (
+    <>
```
RafaelTaranto: There's only one parent div, no need for the `<>`

new-lamassu-admin/src/pages/OperatorInfo/ContactInfo.js
```diff
@@ -0,0 +1,306 @@
...
+            phoneNumber: Yup.string()
+              .matches(mask, { excludeEmptyString: true })
+              .max(100, 'Too long'),
+            email: Yup.string()
+              .email('Please enter a valid email address')
+              .max(100, 'Too long'),
+            website: Yup.string()
+              .url('Please enter a valid url')
+              .max(100, 'Too long'),
+            companyNumber: Yup.string().max(30, 'Too long')
```
RafaelTaranto: Can you move this to objects outside of the html tags, easier to read that way.

new-lamassu-admin/src/components/inputs/base/TextInput.js
```diff
@@ -40,14 +81,23 @@ const TextInput = memo(
...
+      [classes.filled]: !error && value,
+      [classes.empty]: !value || value === ''
+    }
+
     return (
       <TextField
         id={name}
         onChange={onChange}
         onBlur={onBlur}
-        error={!!(touched[name] && errors[name])}
```
RafaelTaranto, line 48: We need to remove this formik stuff from here, but it shouldn't break things. Errors should only show if it was already touched, can you add this logic to the formik specific text input?

RafaelTaranto (changes requested):

new-lamassu-admin/src/components/inputs/base/TextInput.js
lcsfelix: `props.className` is omitted because it is already passed into the classNames object on line 86, which is used on line 100. If it was not omitted, the child component would receive **two** `className` props and there would be a conflict. This is visible on the Phone Number input where, if `props.className` is not omitted, the label has the wrong color when the input is empty.

RafaelTaranto: Ahh, just saw that. Instead of using props.className you can desconstruct that.

`  ({ name, onChange, onBlur, value, error, suffix, large, ...props }) => {`
`  ({ name, onChange, onBlur, value, error, suffix, large, className, ...props }) => {`

and then you don't need to omit it.

lcsfelix: Ah, good point. Changed in https://github.com/lamassu/lamassu-server/pull/354/commits/cb69441239bf2b56da02775aad37a5a16c28fe85

RafaelTaranto:
lgtm!

RafaelTaranto (approved):

PR #355, opened 2020-01-15
feat: add coin atm radar settings page

liordino:


lib/coinatmradar/coinatmradar.js
```diff
@@ -25,12 +25,12 @@ function mapCoin (info, deviceId, settings, cryptoCode) {
   const rates = plugins(settings, deviceId).buildRates(info.rates)[cryptoCode] || { cashIn: null, cashOut: null }
   const cryptoConfig = configManager.scoped(cryptoCode, deviceId, config)
   const unscoped = configManager.unscoped(config)
-  const showRates = unscoped.coinAtmRadarShowRates
```
RafaelTaranto: Since we still didn't flip the switch on the breaking changes can you make this another file called new-coinatmradar?

new-lamassu-admin/src/components/booleanPropertiesTable/booleanPropertiesTable.styles.js
```diff
@@ -0,0 +1,61 @@
...
+  },
+  rightAligned: {
+    display: 'flex',
+    position: 'absolute',
+    right: 0
+  },
+  radioButtons: {
+    display: 'flex',
+    flexDirection: 'row',
+    marginRight: -15 // TODO: any other way to do this?
```
RafaelTaranto: You can send me on chat exactly what you want to do here with pictures and I can try to help.

new-lamassu-admin/src/pages/CoinATMRadar/CoinATMRadar.js
```diff
@@ -0,0 +1,180 @@
...
+          <BooleanPropertiesTable
+            title={'Contact info'}
+            disabled={!coinAtmRadarConfig.active}
+            data={coinAtmRadarConfig}
+            elements={mapPropertiesToTableElements(operatorInfo)}
+            save={save}
+          />
+        </>
+      ) : (
+        <></>
```
RafaelTaranto: This is basically returning null, what you can do here is `coinAtmRadarConfig.active && (` instead of a ternary if.

new-lamassu-admin/src/pages/CoinATMRadar/CoinATMRadar.js
```diff
@@ -0,0 +1,180 @@
...
+            <HelpIcon />
+            <Popper
+              open={helpPopperOpen}
+              anchorEl={helpPopperAnchorEl}
+              placement="bottom"
+              onClose={handleCloseHelpPopper}>
+              <div className={classes.popoverContent}>
+                <P>
+                  For details on configuring this panel, please read the
+                  relevant knowledgebase article{' '}
```
RafaelTaranto: what is the `{' '}` for?

new-lamassu-admin/src/pages/CoinATMRadar/CoinATMRadar.js
```diff
@@ -0,0 +1,180 @@
...
+  active: false,
+  location: null,
+  commissions: false,
+  supportedCryptocurrencies: false,
+  supportedFiat: false,
+  supportedBuySellDirection: false,
+  limitsAndVerification: false,
+  operatorName: null,
+  operatorPhoneNumber: null,
+  operatorEmail: null
```
RafaelTaranto: Why some using null and other false?

new-lamassu-admin/src/pages/CoinATMRadar/CoinATMRadar.js
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  const mapPropertiesToTableElements = properties =>
+    Object.entries(properties)
+      .map(it => ({
+        name: it['0'],
+        value: it['1']
+      }))
+      .filter(it => it.value != null)
+
```
RafaelTaranto: I feel like if this is needed things could be simplified, feel free to hit me up on chat and talk about this.

new-lamassu-admin/src/pages/CoinATMRadar/CoinATMRadar.js
```diff
@@ -0,0 +1,180 @@
...
+      <div className={classes.rowWrapper}>
+        <P>Share information?</P>
+        <div className={classes.switchWrapper}>
+          <Switch
+            checked={coinAtmRadarConfig.active}
+            onChange={event =>
+              save({
+                ...coinAtmRadarConfig,
+                active: event.target.checked
+              })
```
RafaelTaranto: Shouldn't the switch on flip that one variable on the server? We don't want to save everything based on a switch flip.

new-lamassu-admin/src/components/booleanPropertiesTable/booleanPropertiesTable.js
```diff
@@ -0,0 +1,113 @@
...
+    const radioButtonOptions = [
+      { label: 'Yes', value: true },
+      { label: 'No', value: false }
+    ]
+
+    if (!elements || radioGroupValues?.length === 0) return null
+
+    const useStyles = makeStyles(booleanPropertiesTableStyles)
+
+    const classes = useStyles()
```
RafaelTaranto: Make styles should be outside of the component and hooks calls should be the first thing on the component.

new-lamassu-admin/src/components/booleanPropertiesTable/booleanPropertiesTable.js
```diff
@@ -0,0 +1,113 @@
...
+
+    return (
+      <div className={classes.booleanPropertiesTableWrapper}>
+        <div className={classes.rowWrapper}>
+          <H4>{title}</H4>
+          {editing ? (
+            <div className={classes.rightAligned}>
+              <Link
+                style={{ marginRight: '20px' }}
```
RafaelTaranto: We should prevent from using styles directly, move this to jss

new-lamassu-admin/src/pages/CoinATMRadar/CoinATMRadar.js
```diff
@@ -0,0 +1,180 @@
...
+
+  const mapPropertiesToTableElements = properties =>
+    Object.entries(properties)
+      .map(it => ({
+        name: it['0'],
+        value: it['1']
+      }))
+      .filter(it => it.value != null)
+
+  const useStyles = makeStyles(mainStyles)
```
RafaelTaranto: Make styles should be outside of the component.

new-lamassu-admin/src/pages/CoinATMRadar/CoinATMRadar.js
```diff
@@ -0,0 +1,180 @@
...
+
+  const classes = useStyles()
```
RafaelTaranto: Hook calls should be the first calls on the component.

new-lamassu-admin/src/pages/CoinATMRadar/CoinATMRadar.js
```diff
@@ -0,0 +1,180 @@
...
+  const helpPopperOpen = Boolean(helpPopperAnchorEl)
+
+  if (!coinAtmRadarConfig) return null
```
RafaelTaranto: I don't get this one, are we not showing the screen if the config is null? What does it mean to have this configuration null?

RafaelTaranto (changes requested):

lib/coinatmradar/coinatmradar.js
liordino: Sure! Just renaming it should do the trick, right? Or should I also readd the old file?

new-lamassu-admin/src/pages/CoinATMRadar/CoinATMRadar.js
liordino: I though about showing only what had values, and not showing the ones with null value. But I think that this can be better. Maybe just not adding the properties that will not be used right now and add them later.

new-lamassu-admin/src/pages/CoinATMRadar/CoinATMRadar.js
liordino: To add an space between the words 'article' and 'here', as suggested by the prettier extension. The only other way was to open the a tag on the same line as the 'article' word, after a space, but that lead to a fail on the compilation.

new-lamassu-admin/src/pages/CoinATMRadar/CoinATMRadar.js
liordino: I agree. I think that improving the structure of this page properties will help it.

new-lamassu-admin/src/pages/CoinATMRadar/CoinATMRadar.js
liordino: Agreed.

new-lamassu-admin/src/components/booleanPropertiesTable/booleanPropertiesTable.js
liordino: Roger! But why exactly hooks should be the first thing?

EDIT: neverming, I got it. Should be guaranteed to be executed everytime, right?

liordino: I also couldn't think of a better name for this component... any suggestions?

new-lamassu-admin/src/pages/CoinATMRadar/CoinATMRadar.js
liordino: In this case we agreed to keep showing the table and just make it disabled, so this isn't needed anymore.

new-lamassu-admin/src/pages/CoinATMRadar/CoinATMRadar.js
liordino: The config will be null only until useQuery is completed. After that, the page is rendered. As the code is now removing this would cause some errors. Any suggestions to improve this flow?

RafaelTaranto: Ohh, got it. It can stay as is.

RafaelTaranto (approved):

RafaelTaranto:
[Create the coinATMradar screen](https://trello.com/c/rRG3mzzc/151-create-the-coinatmradar-screen)

PR #357, opened 2020-01-15
fix: downgrade BTC, BCH, LTC daemons

CrypticaScriptura:
BTC & LTC 0.17 and above change the account RPC such that unconfirmed deposits are not seen until confirmed. This happens on BCH 0.20.6 and above as well. Remain on older releases until new RPC calls are accounted for.

### DIFF
--- a/lib/blockchain/common.js
+++ b/lib/blockchain/common.js
@@ -21,8 +21,8 @@ module.exports = {
 
 const BINARIES = {
   BTC: {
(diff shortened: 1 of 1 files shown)


CrypticaScriptura:
[fix: downgrade BTC, BCH, LTC daemons](https://trello.com/c/hdYcbFeP/171-fix-downgrade-btc-bch-ltc-daemons)

PR #358, opened 2020-01-17
fix: make nav and sidebar styleguide compliant

lcsfelix:


new-lamassu-admin/src/components/Header.js
```diff
@@ -14,7 +16,7 @@ const renderSubheader = (item, classes) => {
   return (
     <div className={classes.subheader}>
       <div className={classes.content}>
-        <H4>{item.label}</H4>
+        {/* <H4>{item.label}</H4> */}
```
RafaelTaranto: Hmm, I don't remember this being removed from the subheader.

RafaelTaranto (changes requested):

RafaelTaranto (approved):

PR #360, opened 2020-01-21
fix: use getbalance as intended, include 0conf spends

CrypticaScriptura:


CrypticaScriptura:
[Ability to spend wallet's own unconfirmed change](https://trello.com/c/4WdvrXJR/11-ability-to-spend-wallets-own-unconfirmed-change)

joshmh:
Note: wait on inclusion until we have this: https://trello.com/c/TZ9EHWyQ/2-new-wallet-releases-break-pending-balance-rpc

joshmh:
Also, getbalance is being phased out: https://github.com/bitcoin/bitcoin/blob/53f26cd11dc04a0ab7a3ee72a3469050c4c2bc17/doc/release-notes/release-notes-0.17.0.md

We should probably remove the deprecated fields.

Since 0.17 (end of 2018) we can use ``getbalances`` and use ``mine.trusted`` from the response. Do we need to support earlier versions of bitcoind?

CrypticaScriptura:
@joshmh We don't need to support earlier versions of bitcoind if operators ensure to upgrade the wallet daemon at the same time as server upgrade. Currently, we don't do that in the upgrade script, though we could incorporate it. Doing so would better keep operators ahead of ZEC & BCH hard fork upgrades.

CrypticaScriptura:
Solved with https://github.com/lamassu/lamassu-server/pull/386

PR #362, opened 2020-01-22
fix: reduce network usage

RafaelTaranto:


RafaelTaranto:
[High machine data usage](https://trello.com/c/HME4AzX5/49-high-machine-data-usage)

PR #364, opened 2020-01-28
fix: change rpc call after get balance deprecation

RafaelTaranto:


RafaelTaranto:
[New wallet releases break pending balance RPC](https://trello.com/c/TZ9EHWyQ/170-new-wallet-releases-break-pending-balance-rpc)

PR #366, opened 2020-02-02
chore: default to twilio and mailgun

CrypticaScriptura:


CrypticaScriptura:
[Default to Twilio and SMS gateways](https://trello.com/c/wXuewqjm/24-default-to-twilio-and-sms-gateways)

PR #372, opened 2020-02-11
hotfix: add fudge factor

RafaelTaranto:


RafaelTaranto:
[Add fudgeFactor to cryptoAtom amount for incoming cash-out deposits](https://trello.com/c/fYleoCiW/12-add-fudgefactor-to-cryptoatom-amount-for-incoming-cash-out-deposits)

lib/cash-out/cash-out-helper.js
```diff
@@ -11,14 +11,17 @@ module.exports = { redeemableTxs, toObj, toDb, REDEEMABLE_AGE }
...
-    'rawTickerPrice'
-  ])
-    ? value.toString()
-    : value
+  const convert = (value, key) => {
+    if (_.includes(key, [ 'cryptoAtoms', 'receivedCryptoAtoms', 'fiat' ])) {
+      return value.toString()
+    }
+
+    if (_.includes(key, [ 'commissionPercentage', 'rawTickerPrice' ]) && value) {
```
joshmh: Why not always check for existence of value when returning ``toString()``? Also, check for ``isNil``, not raw value since it could be 0.

lib/wallet.js
```diff
@@ -122,8 +122,11 @@ function mergeStatusMode (a, b) {
 }
 
 function getWalletStatus (settings, tx) {
+  const fudgeFactorEnabled = configManager.unscoped(settings.config).fudgeFactor
+  const fudgeFactor = fudgeFactorEnabled ? 10 : 0
+
   const walletStatusPromise = fetchWallet(settings, tx.cryptoCode)
-    .then(r => r.wallet.getStatus(r.account, tx.toAddress, tx.cryptoAtoms, tx.cryptoCode))
+    .then(r => r.wallet.getStatus(r.account, tx.toAddress, tx.cryptoAtoms.add(fudgeFactor), tx.cryptoCode))
```
joshmh: Don't we want to subtract the fudgeFactor here?

joshmh (changes requested):

lib/cash-out/cash-out-helper.js
RafaelTaranto: Since the other fields should never be null I would rather them to break, added comment on it.

lib/wallet.js
RafaelTaranto: Good catch, this was previously inside each wallet, hence the error.

CrypticaScriptura:
[Recurring BigNumber errors](https://trello.com/c/yD992sTc/49-recurring-bignumber-errors)

PR #377, opened 2020-02-24
fix: prevent trading on tx rollback

RafaelTaranto:


RafaelTaranto:
[Trade executed for twice the expected value](https://trello.com/c/I3bsZL9Z/44-trade-executed-for-twice-the-expected-value)

joshmh:
Explain the logic behind the changes.

PR #378, opened 2020-02-26
feat: add customer profile page

liordino:


liordino:
[Create the customer profile screen](https://trello.com/c/uBVJcwJF/31-create-the-customer-profile-screen)

lib/new-admin/admin-server.js
```diff
@@ -53,6 +57,9 @@ apolloServer.applyMiddleware({
 // cors on app for /api/register endpoint.
 app.use(cors({ credentials: true, origin: devMode && 'https://localhost:3000' }))
 
+app.use('/id-card-photo', serveStatic(idPhotoCardBasedir, {index: false}))
+app.use('/front-camera-photo', serveStatic(frontCameraBasedir, {index: false}))
```
RafaelTaranto, line 63: This is a security issue since auth is being handled by graphql, needs to be further discussed.

lib/new-admin/graphql/schema.js
```diff
@@ -159,12 +187,14 @@ const typeDefs = gql`
...
     machines: [Machine]
     customers: [Customer]
+    customer(customerId: ID!): Customer
     machineLogs(deviceId: ID!): [MachineLog]
     funding: [CoinFunds]
     serverVersion: String!
     uptime: [ProcessStatus]
     serverLogs: [ServerLog]
     transactions: [Transaction]
+    customerTransactions(customerId: ID!): [Transaction]
```
RafaelTaranto: I think this should be a field under customer with a custom resolver.

lib/customers.js
```diff
@@ -85,6 +85,28 @@ function update (id, data, userToken, txId) {
...
+ * @name updateCustomer
+ * @function
+ *
+ * @param {string} id Customer's id
+ * @param {object} data Fields to update
+ *
+ * @returns {Promise} Newly updated Customer
+ */
+async function updateCustomer (id, data) {
+  const formattedData = _.omit(['id'], _.mapKeys(_.snakeCase, data))
```
RafaelTaranto: we should be really careful with this and pick only the fields we want to update from data

new-lamassu-admin/src/components/photo/index.js
```diff
@@ -0,0 +1,3 @@
+import Photo from './Photo'
```
RafaelTaranto: I don't see this used anywhere, nor anything under this photo component. Also remove .DS_Store from the commit

new-lamassu-admin/src/pages/Customers/CustomerProfile.js
```diff
@@ -0,0 +1,378 @@
...
+      header: 'Last active',
+      size: 142,
+      value: moment.utc(customer.lastActive).format('YYYY-MM-D')
+    },
+    {
+      header: 'Last transaction',
+      size: 198,
+      value: (
+        <>
+          {`${Number.parseFloat(customer.lastTxFiat)}
```
RafaelTaranto: I think this is the stuff that can be causing all of the NaNs. Lets make a method to cleanly handle invalid numbers and use it instead of the parses under Number

new-lamassu-admin/src/pages/Customers/CustomerProfile.js
```diff
@@ -0,0 +1,378 @@
+import { makeStyles } from '@material-ui/core/styles'
```
RafaelTaranto, line 1: This became a huge component. Do try and create some smaller ones under /pages/Customer/components/ and add whatever you can there. I see a bunch of repetition here. Also feel free to add page specific components there, like the PropertyCard one.

RafaelTaranto (changes requested):

lib/new-admin/admin-server.js
liordino: Got it. Is it a good idea to get the images from gql itself?

new-lamassu-admin/src/components/photo/index.js
liordino: This was an idea that I abandoned but forgot to delete.

lib/new-admin/graphql/schema.js
liordino: Didn't know how to do it or even thought about it, but agreed, waaaaay better

lib/customers.js
liordino: I don't know if I understood it right, but my idea was that the fields that we want to update are all of the ones that came in 'data', considering that it isn't always with all of the customer fields. Is it the requested change?

new-lamassu-admin/src/pages/Customers/CustomerProfile.js
liordino: I think that all of the NaNs are caused by null values in the parse methods... should I just test for a null value prior to parsing it then?

lib/new-admin/admin-server.js
RafaelTaranto: I'm thinking we can either grab them from graphql, pull auth to the outside of graphql or anonymize it with random strings

lib/customers.js
RafaelTaranto: Right, what I'm saying is that you should use `_.pick` instead of `_.omit`.

The idea is to safeguard against future changes on data and only update the field when explicitly told to. So even if someone adds new stuff in data, we only update what we're passing to `_.pick`

new-lamassu-admin/src/pages/Customers/CustomerProfile.js
RafaelTaranto: You should, what I'm saying is to create a method for this and use it everywhere instead of doing the test everywhere.

new-lamassu-admin/src/components/fake-table/Table.js
```diff
@@ -108,13 +111,10 @@ const Th = ({ children, ...props }) => {
 const Tr = ({ error, errorMessage, children, className }) => {
   const classes = useStyles()
   const cardClasses = { root: classes.cardContentRoot }
-  const classNames = {
-    [classes.trError]: error
-  }
```
RafaelTaranto: Why did this got removed? This is probably being used by other screens. 

You can do the following:

```
const classNames = {
  [classes.trError]: error,
  [classes.card]: true,
  className
}
```

RafaelTaranto: Also this has conflicts now, do rebase after doing the changes.

lib/new-admin/admin-server.js
RafaelTaranto: @liordino just look at this server side code and its already being handled like you're doing, so no harm done here. Image strings are random enough as it is.

RafaelTaranto (approved):

PR #384, opened 2020-03-25
feat: reintroduce lowBalanceMargin

CrypticaScriptura:
We used to employ a low balance margin on our legacy platform, such that we'd consider the wallet balance prematurely low by a certain percentage. This avoided scenarios where a customer attempted a purchase up to the full wallet balance, but which didn't leave enough for bitcoin transaction fees, thus failing to send. Since the transaction fee isn't known until sending, including a buffer largely accounted for this. This adds it back at 3%.

### DIFF
--- a/lib/plugins.js
+++ b/lib/plugins.js
@@ -308,7 +308,7 @@ function plugins (settings, deviceId) {
 
         const rate = rawRate.div(cashInCommission)
 
-        const lowBalanceMargin = BN(1)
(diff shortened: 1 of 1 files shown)


CrypticaScriptura:
[Reintroduce low balance margin to avoid insufficient fund errors](https://trello.com/c/QrbBLF1D/90-reintroduce-low-balance-margin-to-avoid-insufficient-fund-errors)

PR #387, opened 2020-03-26
chore: update BTC, BCH, LTC, ETH wallets

CrypticaScriptura:
Depends on https://github.com/lamassu/lamassu-server/pull/386.


CrypticaScriptura:
[Update BTC, BCH, LTC, ETH wallets](https://trello.com/c/PMQpMsfn/91-update-btc-bch-ltc-eth-wallets)

PR #390, opened 2020-03-30
feat: allow override of edit buttons on EditableTable

mautematico:


mautematico:
[Modify EditableTable to allow override of edit button and header](https://trello.com/c/ktUgBOPm/95-modify-editabletable-to-allow-override-of-edit-button-and-header)

PR #391, opened 2020-03-30
fix: customer list bugfixes

liordino:
style: added stripes as bg to table rows

feat: created the stripped lines pattern for null values in tables

style: center objects vertically in the tables

fix: removed the vertical alignment on the Td component (broke some
tables)

fix: coalesce the last_active customer property to it's own creation
when there's no tx associated

fix: check for null values in the customer properties

fix: fix the ordering of the customers table

style: centered td contents vertically

### DIFF
--- a/lib/customers.js
+++ b/lib/customers.js
@@ -386,14 +386,16 @@ function batch () {
  */
 function getCustomersList () {
   const sql = `select name, phone, total_txs, total_spent, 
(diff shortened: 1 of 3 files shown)


liordino:
[Customer list bugfixes](https://trello.com/c/Wf7iySG8/84-customer-list-bugfixes)

RafaelTaranto (approved):

PR #392, opened 2020-04-06
feat: notifications screen

RafaelTaranto:


RafaelTaranto:
[Rework the Notifications screen](https://trello.com/c/oHBKIoCp/100-rework-the-notifications-screen)

new-lamassu-admin/src/components/editableTable/Row.js
```diff
@@ -1,250 +1,150 @@
...
-              {...inputProps}
-            />
-          )}
-          {!editing && type === 'text' && (
-            <TextInputDisplay display={display} {...inputProps} />
-          )}
+  return (
+    <>
+      {editing && (
+        <Td textAlign="center" width={ACTION_COL_SIZE}>
```
liordino, line 36: My understanding here is that the ACTION_COL_SIZE is used to compute the actual actionColSize that is used and that it either renders just the cancel button with the full ACTION_COL_SIZE or both the edit and delete buttons with half of ACTION_COL_SIZE each. Is that right? If it is, why don't use the computed variable here too? This way, if I got it right, it seems that the ternary if isn't necessary (could just have a halfActionColSize variable, for instance).

new-lamassu-admin/src/pages/Notifications/sections/CryptoBalanceOverrides.js
```diff
@@ -0,0 +1,118 @@
...
+    [LOW_BALANCE_KEY]: '',
+    [HIGH_BALANCE_KEY]: ''
+  }
+
+  const validationSchema = Yup.object().shape({
+    [CRYPTOCURRENCY_KEY]: Yup.string().required(),
+    [LOW_BALANCE_KEY]: Yup.number()
+      .integer()
+      .min(0)
+      .max(99999999)
```
liordino, line 54: Could the max value be on a const? And why exactly this value?

new-lamassu-admin/src/components/editableTable/Row.js
RafaelTaranto: Hey, this was a bit confusing at this version. The reason why it was left with the constant is because when editing all actions icons are replaced with 'cancel' and 'save' buttons. So this is always the same size regardless of how many actions exist.

new-lamassu-admin/src/pages/Notifications/sections/CryptoBalanceOverrides.js
RafaelTaranto: Eh, good question. Just left as it was. I'll just remove this random max.

RafaelTaranto: Removed on the latest PR. Once this gets a review by @mautematico I'll probably close every PR before the big one.

new-lamassu-admin/src/components/buttons/Link.js
```diff
@@ -6,23 +6,25 @@ import styles from './Link.styles'
...
-  }
+const Link = memo(
+  ({ submit, className, children, color = 'primary', ...props }) => {
+    const classes = useStyles()
+    const classNames = {
+      [classes.link]: true,
+      [classes.primary]: color === 'primary',
+      [classes.secondary]: color === 'secondary',
+      [classes.noColor]: color === 'noColor'
+    }
```
mautematico, line 17: Is this the way we are doing "dynamyc classes" on React? I get that this is not what the commit is changing, but this is the first time that I really understand how this piece of code works and I like it, so just wanted to know if this is _the way we are doing it_.

If answer is yes, then I shall modify a couple of my components where I reinvented the wheel.

mautematico (approved):

new-lamassu-admin/src/components/buttons/Link.js
mautematico: ooops, pressed wrong button on the review screen and had not noticed my comment/question was on pending state.

but yes, it looks goof for me

RafaelTaranto: @mautematico this can be done either like this or passing the variable through useStyles, like shown here: https://github.com/lamassu/lamassu-server/blob/bda29ad150354b1915e298ec3972f8ac06023860/new-lamassu-admin/src/components/buttons/Button.js#L10

PR #397, opened 2020-05-06
fix: rename fudge factor and change value

RafaelTaranto:


RafaelTaranto:
[Add fudgeFactor to cryptoAtom amount for incoming cash-out deposits](https://trello.com/c/fYleoCiW/12-add-fudgefactor-to-cryptoatom-amount-for-incoming-cash-out-deposits)

PR #404, opened 2020-07-03
fix: don't assume '+' when using a friendly name with Twilio

liordino:
fix: added missing lodash dependency to accounts lib

fix: don't prepend '+' when using a friendly name with Twilio

### DIFF
--- a/lib/plugins/sms/twilio/twilio.js
+++ b/lib/plugins/sms/twilio/twilio.js
@@ -14,7 +14,8 @@ function sendMessage (account, rec) {
       const client = twilio(account.accountSid, account.authToken)
       const body = rec.sms.body
       const _toNumber = rec.sms.toNumber || account.toNumber
-      const from = _.startsWith('+')(account.fromNumber)
+      const from = (_.startsWith('+')(account.fromNumber) 
+        || !_.isNumber(String(account.fromNumber).replace(/\s/g,'')))
         ? account.fromNumber : `+${account.fromNumber}`
 
       const opts = {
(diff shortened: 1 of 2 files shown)


liordino:
[Don't assume '+' when using a friendly name with Twilio](https://trello.com/c/IAG0lT07/109-dont-assume-when-using-a-friendly-name-with-twilio)

RafaelTaranto (approved):

lib/plugins/sms/twilio/twilio.js
```diff
@@ -14,7 +14,8 @@ function sendMessage (account, rec) {
       const client = twilio(account.accountSid, account.authToken)
       const body = rec.sms.body
       const _toNumber = rec.sms.toNumber || account.toNumber
-      const from = _.startsWith('+')(account.fromNumber)
+      const from = (_.startsWith('+')(account.fromNumber) 
+        || isNaN(String(account.fromNumber).replace(/\s/g,'')))
```
joshmh: Prefer ``_.isNumber``, since ``isNaN`` isn't really meant for this and has some weird special cases.

liordino: Fixed!

lib/plugins/sms/twilio/twilio.js
```diff
@@ -14,7 +14,8 @@ function sendMessage (account, rec) {
       const client = twilio(account.accountSid, account.authToken)
       const body = rec.sms.body
       const _toNumber = rec.sms.toNumber || account.toNumber
-      const from = _.startsWith('+')(account.fromNumber)
+      const from = (_.startsWith('+')(account.fromNumber) 
+        || _.isNumber(String(account.fromNumber).replace(/\s/g,'')))
```
joshmh: Shouldn't this be ``!_.isNumber()``?

liordino: Yes, it should. My bad, already fixed.

PR #405, opened 2020-07-04
fix: Locale fixes

liordino:
fix: added missing lodash dependency to accounts lib

fix: show full languages name on the dropdown

fix: search crypto per name as well

fix: limit selected languages to 4 (the same as the old admin)

### DIFF
--- a/lib/new-admin/config/accounts.js
+++ b/lib/new-admin/config/accounts.js
@@ -1,6 +1,5 @@
 const _ = require('lodash/fp')
 const { COINS, ALL_CRYPTOS } = require('./coins')
-const _ = require('lodash/fp')
 
 const { BTC, BCH, DASH, ETH, LTC, ZEC } = COINS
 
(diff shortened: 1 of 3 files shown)


liordino:
[Locale fixes](https://trello.com/c/q6LuiPLv/141-locale-fixes)

RafaelTaranto (approved):

lib/new-admin/config/accounts.js
```diff
@@ -1,5 +1,6 @@
 const _ = require('lodash/fp')
 const { COINS, ALL_CRYPTOS } = require('./coins')
+const _ = require('lodash/fp')
```
joshmh: This is already required. Why didn't this cause a lint error?

joshmh (changes requested):

lib/new-admin/config/accounts.js
liordino: Fixed!

PR #410, opened 2020-07-14
fix: machine status layout bugs

mautematico:
- [x]     legend colors are different from the spec
- [x]     align popup title with content (unpair action)
- [x]     reboot icon is being cut off, fix it.
- [x]     make all actions open up a pop-up like the unpair one
- [x]     popups should only close in case of success. If it errors show the error like the 3rd party service ones.
- [x]     add support articles for every status
- [ ]     load machine model from l-m
- [ ]     properly display statuses based on machine-pings and machine-state

fix: reboot icon looks cropped
fix: confirm dialog layout
fix: Status chip background colors
fix: detailed machine status layout
fix: machine detailed status layout
fix: machine status article links, status chip size
fix: confirmDialog for all machine actions
fix: confirm dialog on every action. reload when success

### DIFF
--- a/new-lamassu-admin/src/components/ConfirmDialog.js
+++ b/new-lamassu-admin/src/components/ConfirmDialog.js
@@ -2,50 +2,50 @@ import {
   Dialog,
   DialogActions,
   DialogContent,
-  DialogContentText,
   makeStyles
 } from '@material-ui/core'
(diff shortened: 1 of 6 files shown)


mautematico:
[Machine status fixes](https://trello.com/c/nERdgEIJ/144-machine-status-fixes)

PR #412, opened 2020-07-20
fix: logs download popup bug

liordino:
fix: added timestamp parameters for a date range on the gql queries for
machineLogs, serverLogs and transactions

feat: added optional limit and offset variables for the logs queries,
for filtering and pagination

feat: adapted the LogsDownloaderPopper to download the logs by whats set
on the filters

### DIFF
--- a/new-lamassu-admin/src/components/LogsDownloaderPopper.js
+++ b/new-lamassu-admin/src/components/LogsDownloaderPopper.js
@@ -1,3 +1,4 @@
+import { useLazyQuery } from '@apollo/react-hooks'
 import { makeStyles } from '@material-ui/core'
 import classnames from 'classnames'
 import FileSaver from 'file-saver'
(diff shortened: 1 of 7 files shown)


liordino:
[Logs download popup bug](https://trello.com/c/wWsre81D/85-logs-download-popup-bug)

new-lamassu-admin/src/components/LogsDownloaderPopper.js
```diff
@@ -153,39 +157,36 @@ const LogsDownloaderPopover = ({ name, getTimestamp, logs, title }) => {
...
     if (!range) return
 
-    if (range.from && !range.to) range.to = moment()
+    fetchLogs({
+      variables: {
+        ...args,
+        ...(selectedRadio === RANGE && {
+          ...(range.from && { from: range.from }),
+          ...(range.to && { until: range.to })
+        })
```
RafaelTaranto: this is pretty complicated to understand.

Better if you just create a empty object and add fields based on ifs or create something with _.fromPairs

RafaelTaranto (changes requested):

new-lamassu-admin/src/components/LogsDownloaderPopper.js
liordino: Done!

RafaelTaranto: @liordino I'm not sure if this new version would work if we pass down null in from and in to. Did you manage to test that?

liordino: You mean by selecting the ALL option on the radio button group and them passing null to the gql query? Yes, it's working.

RafaelTaranto (approved):

PR #416, opened 2020-07-22
fix: server_logs creates inf loop on l-s install

mautematico:


mautematico:
[Saving server logs to db creates a infinite loop on initial l-s install.](https://trello.com/c/HpmFDkm3/160-saving-server-logs-to-db-creates-a-infinite-loop-on-initial-l-s-install)

RafaelTaranto:
@mautematico we don't want to change migration order.

RafaelTaranto (changes requested):
We don't want to change migration order

bin/lamassu-migrate
```diff
@@ -23,10 +23,21 @@ const getMigrateFile = () => {
...
+  })
   .then(([qResult, migrateFile]) => {
     if (!qResult && migrateFile) {
       return db.none('insert into migrations (id, data) values (1, $1)', [migrateFile])
     }
+  })
+  .then(() => {
+
+
+
```
RafaelTaranto: Remove this empty promise.

RafaelTaranto (changes requested):

bin/lamassu-migrate
mautematico: lol, shame on me!

done, empty promise removed, and actually no additional promise was needed so I merged the other two into a single one.

RafaelTaranto (approved):

PR #417, opened 2020-07-22
feat: Update settings loader to use db instead of json file

liordino:


lib/new-settings-loader.js
```diff
@@ -1,71 +1,115 @@
...
-  return new Promise((resolve) => {
-    if (!db) {
-      setTimeout(() => {
-        return resolve(db.getState())
-      }, 1000)
-    } else {
-      return resolve(db.getState())
-    }
-  })
+  if (!versionId) throw new Error('versionId is required')
```
RafaelTaranto: Better to change this to Promise.reject

lib/new-settings-loader.js
```diff
@@ -1,71 +1,115 @@
...
-const adapter = new FileAsync('db.json')
-let db = null
-
-// TODO new-admin save to actual db, like the old settings
-low(adapter).then(it => {
-  db = it
-})
+const NEW_SETTINGS_LOADER_SCHEMA_VERSION = 2
 
+// TODO: validate
```
RafaelTaranto: We don't want to leave TODOs on our code base, is there a reason for these here? What needs to be validated?

If this is referring to the "valid" row we won't need it anymore, new admin should work in a way that the user never inputs a invalid config.

RafaelTaranto (changes requested):

lib/new-settings-loader.js
liordino: Yeah, it refers to a valid row. I'll remove that, shouldn't be commited.

lib/new-settings-loader.js
```diff
@@ -1,71 +1,111 @@
...
+      return db.none(sql, ['accounts', { accounts: newAccounts }, true, NEW_SETTINGS_LOADER_SCHEMA_VERSION])
+    })
 }
 
-function getAccounts () {
-  const state = db.getState()
-  return state ? state.accounts : null
+function loadAccounts () {
+  const toFields = fieldArr => _.fromPairs(_.map(r => [r.code, r.value], fieldArr))
+  const toPairs = r => [r.code, toFields(r.fields)]
```
RafaelTaranto: Why is this needed? I don't see why it was working before now it's required.

lib/new-settings-loader.js
```diff
@@ -1,71 +1,111 @@
...
+  const sql = `select data
+  from user_config
+  where type=$1
+  and schema_version=$2
+  and valid
+  order by id desc
+  limit 1`
+
+  return db.oneOrNone(sql, ['config', NEW_SETTINGS_LOADER_SCHEMA_VERSION])
+    .then(row => !row ? {} : row.data.config)
```
RafaelTaranto: Is it possible for the query to return a null response? Isn't this QueryResultError already?

Also we don't want to return a empty object for config, just throw the error if we dont find a config.

RafaelTaranto (changes requested):

liordino:
[Update settings loader to use db instead of json file](https://trello.com/c/33AkxRRN/153-update-settings-loader-to-use-db-instead-of-json-file)

RafaelTaranto (approved):

migrations/1595295132361-schema-version-on-user-config.js
```diff
@@ -0,0 +1,13 @@
+const db = require('./db')
+
+module.exports.up = function (next) {
+  var sql = [
+    'alter table user_config add column schema_version smallint DEFAULT 1'
```
RafaelTaranto: Ahh, this should not be nullable.

RafaelTaranto (changes requested):

PR #423, opened 2020-08-17
Feat: Compliance Triggers and Customers

RafaelTaranto:


RafaelTaranto:
[Review Customer screen + Compliance triggers](https://trello.com/c/IgROorRd/130-review-customer-screen-compliance-triggers)

lib/customers.js
```diff
@@ -454,7 +454,7 @@ function getCustomerById (id) {
     c.id_card_data, c.id_card_data_override, c.id_card_data_expiration, c.id_card_photo_path,
     c.id_card_photo_override, t.tx_class, t.fiat, t.fiat_code, t.created, 
     row_number() over (partition by c.id order by t.created desc) as rn,
-    count(0) over (partition by c.id) as total_txs,
+    sum(case when t.id is not null then 1 else 0 end) over (partition by c.id) as total_txs,
```
joshmh, line 457: Can we filter by NOT NULL t.id?

joshmh (changes requested):

lib/customers.js
RafaelTaranto: at the end this wasn't really needed

RafaelTaranto: The issue is when the subqueries return no results, the count returns as 1 instead of 0

PR #424, opened 2020-08-18
feat: Create migration from old config to new

liordino:


liordino:
[Create migration from old config to new](https://trello.com/c/HKZowKPG/164-create-migration-from-old-config-to-new)

lib/config-migration.js
```diff
@@ -0,0 +1,442 @@
...
+function migrateAccounts(accounts) {
+  function renameAccountFields(account, replacements) {
+    return _.fromPairs(
+      _.toPairs(account)
+        .map(p => [
+          replacements[p[0]],
+          p[1]
+        ])
+    )
+  }
```
RafaelTaranto: This can be replaced with `_.mapKeys(it => replacements[it])` (not tested code)

this won't be needed because of the bitgo change.

lib/config-migration.js
```diff
@@ -0,0 +1,442 @@
...
+
+  return {
+    accounts: _.fromPairs(accountsProps)
+  }
+}
+
+function migrate(config, accounts) {
+  return {
+    ...migrateConfig(config),
+    ...migrateAccounts(accounts)
```
RafaelTaranto: better to just return a plain object on these methods and do:

```
config: migrateConfig(config),
accounts: migrateAccounts(accounts)
```

lib/config-migration.js
```diff
@@ -0,0 +1,442 @@
...
+    BTCWalletPassphrase: 'btcWalletPassphrase',
+    DASHWalletId: 'dashWalletId',
+    DASHWalletPassphrase: 'dashWalletPassphrase',
+    environment: 'environment',
+    LTCWalletId: 'ltcWalletId',
+    LTCWalletPassphrase: 'ltcWalletPassphrase',
+    token: 'token',
+    ZECWalletId: 'zecWalletId',
+    ZECWalletPassphrase: 'zecWalletPassphrase'
+  }
```
RafaelTaranto: If it's only the casing that's changing I think it's better to rename it on the new admin, this is actually a bug since the bitgo module expects upper case coins to work.

lib/config-migration.js
```diff
@@ -0,0 +1,442 @@
...
+
+  let accountsProps = []
+  if (accounts.bitgo) accountsProps.push(['bitgo', renameAccountFields(accounts.bitgo, bitgoFields)])
+  if (accounts.bitstamp) accountsProps.push(['bitstamp', accounts.bitstamp])
+  if (accounts.blockcypher) accountsProps.push(['blockcypher', accounts.blockcypher])
+  if (accounts.infura) accountsProps.push(['infura', accounts.infura])
+  if (accounts.itbit) accountsProps.push(['itbit', accounts.itbit])
+  if (accounts.kraken) accountsProps.push(['kraken', accounts.kraken])
+  if (accounts.mailgun) accountsProps.push(['mailgun', accounts.mailgun])
+  if (accounts.twilio) accountsProps.push(['twilio', accounts.twilio])
```
RafaelTaranto: Instead of doing all of this the whole method will become:

``` js
const accountArray = ['bitgo', 'bitstamp'] // plus all other accounts
return _.pick(accountArray)(accounts)
```

lib/config-migration.js
```diff
@@ -0,0 +1,442 @@
+const _ = require('lodash/fp')
+const uuid = require('uuid')
+const { COINS } = require('../lib/new-admin/config/coins')
+
+const GLOBAL = 'global'
+const ALL_CRYPTOS = _.values(COINS).sort()
+const ALL_MACHINES = 'all machines'
```
RafaelTaranto: better to make this `const ALL_MACHINES =  'ALL_MACHINES'`

lib/config-migration.js
```diff
@@ -0,0 +1,442 @@
...
+}
+
+function migrateCommissions(config) {
+  function areArraysEquals(arr1, arr2) {
+    return _.xor(arr1, arr2).length === 0
+  }
+
+  function getAllMachines(config) {
+    return _.uniq(config.map(f => f.fieldLocator.fieldScope.machine).filter(m => m !== GLOBAL))
+  }
```
RafaelTaranto: No longer being used.

lib/config-migration.js
```diff
@@ -0,0 +1,442 @@
...
+
+  const codes = {
+    minimumTx: 'minimumTx',
+    cashInFee: 'fixedFee',
+    cashInCommission: 'cashIn',
+    cashOutCommission: 'cashOut'
+  }
+
+  const { global, scoped } = getConfigFields(_.keys(codes), config)
+  const allMachines = getAllMachines(config).sort()
```
RafaelTaranto: No longer being used.

lib/config-migration.js
```diff
@@ -0,0 +1,442 @@
...
+        scope: JSON.parse(f[0]), 
+        values: f[1]
+      })) || []
+  }
+}
+
+function migrateCommissions(config) {
+  function areArraysEquals(arr1, arr2) {
+    return _.xor(arr1, arr2).length === 0
+  }
```
RafaelTaranto: Maybe change to:
`const areArraysEquals = (arr1, arr2) => _.isEmpty(_.xor(arr1, arr2))`

lib/config-migration.js
```diff
@@ -0,0 +1,442 @@
...
+            .filter(innerF => innerF.scope.machine.includes(f.scope.machine))
+            .map(innerF => innerF.scope.crypto)
+        )
+        
+        f.scope.crypto = _.difference(
+          ALL_CRYPTOS,
+          cryptosAlreadyScopedWithThisFieldMachine
+        )
+
+        return f
```
RafaelTaranto: This is pretty hard to understand. Some things you can do to make this clearer:

1. Try to make it step by step. For instance do only the machineScoped.map and create a variable for that. Afterwards you can concat in another variable.

2. Names are pretty explanatory, but it's too long for comfort. When we're reading this we only need some context if the code shows intent. 

``` js
// machineAndCryptoScopedPlusCryptoScopedCommissionsOverrides changed to withCryptoScoped
// cryptosAlreadyScopedWithThisFieldMachine changed to coinsToRemove or toRemove
```

3. Doing it more functionally (with the help of lodash) can help breaking things into smaller chunks and make intent clearer.

We can break logic into pretty bite sized chunks that are not that hard to follow:

``` js


const getMachine = _.get('scope.machine')
const getCrypto = _.get('scope.crypto')

const filteredMachineScoped = _.map(it => {
     const filterByMachine = _.filter(_.includes(getMachine(it)))
     const flattenCoins =  _.compose(_.flatten, _.map(getCrypto))
     const diffAllCryptos = _.compose(_.diference(ALL_CRYPTOS))

     const unrepeatedCryptos = _.compose(diffAllCryptos, flattenCoins, filterByMachine)(withCryptoScoped) 

     return _.set('scope.crypto', unrepeatedCryptos)(it)
)(machineScoped)

```

Note how even flattenCoins and diffAllCryptos can be moved to the outside of the function since they don't require any context. It would make the function a 3 liner.

RafaelTaranto (changes requested):

lib/config-migration.js
liordino: Way better! Thanks!

liordino:
[Create a migration to update the machine names on the db](https://trello.com/c/1DpEXWcL/191-create-a-migration-to-update-the-machine-names-on-the-db)

lib/new-settings-loader.js
```diff
@@ -1,71 +1,114 @@
...
+const db = require('./db')
 
-const adapter = new FileAsync('db.json')
-let db = null
-
-// TODO new-admin save to actual db, like the old settings
-low(adapter).then(it => {
-  db = it
-})
+const NEW_SETTINGS_LOADER_SCHEMA_VERSION = 2
```
RafaelTaranto: Wasn't this already merged in another PR? There's something strange here.

lib/config-migration.js
```diff
@@ -0,0 +1,462 @@
...
+  if (global.frontCameraVerificationActive) {
+    triggers.push(
+      createTrigger(requirements.facePhoto, global.smsVerificationThreshold)
+    )
+  }
+  if (global.sanctionsVerificationActive) {
+    triggers.push(
+      createTrigger(requirements.sanctions, global.smsVerificationThreshold)
+    )
+  }
```
RafaelTaranto: Check if active and threshold are set before creating the trigger

lib/config-migration.js
```diff
@@ -0,0 +1,462 @@
...
+    sanctions: 'sanctions'
+  }
+
+  function createTrigger (
+    requirement,
+    threshold,
+    thresholdDays = 1,
+    triggerType = triggerTypes.volume,
+    id = uuid.v4(),
+    cashDirection = 'both'
```
RafaelTaranto: On this file you don't need all of this. Just set it straight on the return statement

RafaelTaranto (changes requested):

lib/new-settings-loader.js
liordino: Yeah, it's already on dev. Dunno why it's considered a change here

lib/config-migration.js
```diff
@@ -386,6 +382,11 @@ function migrateComplianceTriggers (config) {
     getConfigFields(codes, config).global.map(f => [f.code, f.value])
   )
 
+  if (!_.isNumber(global.smsVerificationThreshold))
+    return {
+      triggers: []
+    }
+
```
RafaelTaranto: I don't think we want this. There can be other requirements active without a sms threshold. What I meant on the my comment was to test both *VerificationActive and *Threshold for every IF of the triggers creation logic.

RafaelTaranto (changes requested):

PR #425, opened 2020-08-18
fix: Locale overrides should only allow one row per machine

liordino:
fix: filter already overriden machines on locales

### DIFF
--- a/new-lamassu-admin/src/pages/Locales/Locales.js
+++ b/new-lamassu-admin/src/pages/Locales/Locales.js
@@ -69,6 +69,18 @@ const Locales = ({ name: SCREEN_KEY }) => {
     return saveConfig({ variables: { config } })
   }
 
+  const removeOverridenMachines = data => {
+    if (data) {
+      const overridenMachines = locale.overrides?.map(o => o.machine)
+      const machinesIndex = data.findIndex(o => o.name === 'machine')
+      data[machinesIndex].inputProps.options = data[
+        machinesIndex
+      ].inputProps.options?.filter(m => !overridenMachines.includes(m.deviceId))
+    }
+
+    return data
+  }
+
   return (
     <>
       <TitleSection title="Locales" />
@@ -97,7 +109,7 @@ const Locales = ({ name: SCREEN_KEY }) => {
           save={saveOverrides}
           validationSchema={OverridesSchema}
           data={locale.overrides ?? []}
-          elements={overrides(data)}
+          elements={removeOverridenMachines(overrides(data))}
         />
       </Section>
     </>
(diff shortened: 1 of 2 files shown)


liordino:
[Locale overrides should only allow one row per machine](https://trello.com/c/5IvfE8nD/190-locale-overrides-should-only-allow-one-row-per-machine)

new-lamassu-admin/src/pages/Locales/Locales.js
```diff
@@ -69,6 +69,18 @@ const Locales = ({ name: SCREEN_KEY }) => {
...
+      const overridenMachines = locale.overrides?.map(o => o.machine)
+      const machinesIndex = data.findIndex(o => o.name === 'machine')
+      data[machinesIndex].inputProps.options = data[
+        machinesIndex
+      ].inputProps.options?.filter(m => !overridenMachines.includes(m.deviceId))
+    }
+
+    return data
+  }
+
```
RafaelTaranto: This wont work. 

- empty overrides table break the page.
- already existing overrides won't show the machine when being edited.

Take a look at how the FiatBalanceOverrides does it. Better to just follow that logic.

new-lamassu-admin/src/pages/Locales/helper.js
```diff
@@ -39,7 +54,11 @@ const allFields = getData => {
       view: getView(machineData, 'name', 'deviceId'),
       input: Autocomplete,
       inputProps: {
-        options: machineData,
+        options: it =>
+          R.concat(
+            suggestionFilter(machineData),
+            findSuggestion(it, machineData)
```
RafaelTaranto: Probably better to do

```
R.concat(it?.machine ? [it.machine] : [])(suggestionFilter...)
```

After the changes on Autocomplete the code checks for id to match a option, so I don't think we need to fetch it within the machineData.

new-lamassu-admin/src/pages/Locales/helper.js
```diff
@@ -20,12 +23,24 @@ const allFields = getData => {
...
   const displayCodeArray = data => it => {
     if (!it) return it
 
     return R.compose(R.join(', '), R.map(getView(data, 'code')))(it)
   }
 
+  const overridenMachines = R.map(override => override.machine, auxElements)
+  const suggestionFilter = R.filter(
+    it => !R.contains(it.deviceId, overridenMachines)
+  )
```
RafaelTaranto: Better to use https://ramdajs.com/docs/#differenceWith

RafaelTaranto (changes requested):

RafaelTaranto (approved):

PR #428, opened 2020-08-18
fix: Machine status actions popups should not have its input with the error color until unfocused

liordino:
fix: machine status popup validation now occurs only on lose of focus


liordino:
[Machine status actions popups should not have its input with the error color until unfocused.](https://trello.com/c/kjyOmYBG/181-machine-status-actions-popups-should-not-have-its-input-with-the-error-color-until-unfocused)

RafaelTaranto (approved):

PR #429, opened 2020-08-20
fix: Table fixes

liordino:
feat: first field now autofocus when editing a row

fix: move the focus to the first editable element

fix: make the Autocomplete options popup on autoFocus

feat: allow saving on table only when changes occurred

fix: compare only the editing row instead of the entire list to decide
if the save method will be called


liordino:
[Table fixes](https://trello.com/c/uxaylacS/184-table-fixes)

RafaelTaranto (approved):

PR #432, opened 2020-08-30
fix: Editing/saving fixes

liordino:
feat: table add button is now hidden when adding/editing

feat: disable every other action on editable table when editing/adding

feat: hide add button instead of disabling it when can't add

feat: disable every other action when adding/editing on the commissions
page

feat: disable every other action when adding/editing on the locales
page

feat: disable every other action when adding/editing on the
notifications page

feat: disable save buttons while waiting for server response on tables
and editable numbers


liordino:
[Editing/saving fixes](https://trello.com/c/qgKzDCE3/188-editing-saving-fixes)

new-lamassu-admin/src/components/editableTable/Table.js
```diff
@@ -80,7 +91,11 @@ const ETable = ({
     setEditing && setEditing(it, true)
   }
 
-  const addField = () => setAdding(true)
+  const addField = () => {
+    setAdding(true)
+    setEditing && setEditing(true, true)
+    // TODO: maybe create a setAdding argument would be better?
```
RafaelTaranto: Eh, this is fine as is. Remove the TODO and fix the conflics

RafaelTaranto (changes requested):

RafaelTaranto (approved):

PR #436, opened 2020-09-11
feat: Add warning on locale when changing fiat currency

liordino:
feat: added an alert to confirm fiat currency changes on locales


liordino:
[Add warning on locale when changing fiat currency](https://trello.com/c/KIxYQQWC/185-add-warning-on-locale-when-changing-fiat-currency)

RafaelTaranto (approved):

PR #437, opened 2020-09-12
fix: Admin "Server" page fixes

liordino:
style: fix uptime chip style

style: reduced the dropdown border-radius from 16px to 8px only when it's open

feat: scroll the table back up with a log level filter change

fix: dropdown border-radius bottom was still 16px instead of 8px when
opened


liordino:
[Admin "Server" page fixes.](https://trello.com/c/83EatHuq/182-admin-server-page-fixes)

RafaelTaranto (approved):

PR #438, opened 2020-09-13
[WIP] feat: Create a migration to update the machine names on the db

liordino:
feat: create migration to move machine names from file to db


liordino:
[Create a migration to update the machine names on the db](https://trello.com/c/1DpEXWcL/191-create-a-migration-to-update-the-machine-names-on-the-db)

PR #439, opened 2020-09-24
fix: No reason for 'Enable' switch on 'Wallet Settings' panel

liordino:
fix: remove the "enable" toggle

fix: make edit available at all times and make it toggle the wizard if
the coin config is empty

fix: disable focus on the first edit

feat: open the coin config wizard when adding a not configured coin on
locales

### DIFF
--- a/new-lamassu-admin/src/components/editableTable/Row.js
+++ b/new-lamassu-admin/src/components/editableTable/Row.js
@@ -175,7 +175,8 @@ const ERow = ({ editing, disabled, lastOfGroup }) => {
     enableDelete,
     enableToggle,
     rowSize,
-    stripeWhen
+    stripeWhen,
+    focusOnEditWhen
   } = useContext(TableCtx)
 
   const classes = useStyles()
@@ -214,7 +215,12 @@ const ERow = ({ editing, disabled, lastOfGroup }) => {
             key={idx}
             config={it}
             editing={editing}
-            focus={idx === elementToFocusIndex && editing}
+            focus={
+              idx === elementToFocusIndex &&
+              editing &&
+              focusOnEditWhen &&
+              focusOnEditWhen(values)
+            }
             extraPaddingRight={extraPaddingRightIndex === idx}
             extraPaddingLeft={extraPaddingLeftIndex === idx}
           />
--- a/new-lamassu-admin/src/components/editableTable/Row.js
+++ b/new-lamassu-admin/src/components/editableTable/Row.js
@@ -176,7 +176,7 @@ const ERow = ({ editing, disabled, lastOfGroup }) => {
     enableToggle,
     rowSize,
     stripeWhen,
-    focusOnEditWhen
+    focusOnEditWhen = () => true
   } = useContext(TableCtx)
 
   const classes = useStyles()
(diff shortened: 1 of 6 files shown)


liordino:
[No reason for 'Enable' switch on 'Wallet Settings' panel](https://trello.com/c/y0lhZQpC/123-no-reason-for-enable-switch-on-wallet-settings-panel)

new-lamassu-admin/src/pages/Locales/Locales.js
```diff
@@ -112,7 +129,9 @@ const Locales = ({ name: SCREEN_KEY }) => {
   }
 
   const save = config => {
+    console.log(config)
```
RafaelTaranto: remove this config.log

new-lamassu-admin/src/pages/Locales/Locales.js
```diff
@@ -121,6 +140,13 @@ const Locales = ({ name: SCREEN_KEY }) => {
     return saveConfig({ variables: { config } })
   }
 
+  const enableCoin = it => {
+    if (!it) return
+
+    const namespaced = fromNamespace(it)(wallets)
+    if (!namespaced?.active) return setWizard(it)
```
RafaelTaranto: I don't get this one. If we're removing the active prop when would this ever be true? Don't we want to check if coin is configured here?

new-lamassu-admin/src/pages/Locales/Locales.js
```diff
@@ -10,7 +10,9 @@ import { Table as EditableTable } from 'src/components/editableTable'
 import Section from 'src/components/layout/Section'
 import TitleSection from 'src/components/layout/TitleSection'
 import { P } from 'src/components/typography'
-import { fromNamespace, toNamespace } from 'src/utils/config'
+import { fromNamespace, toNamespace, namespaces } from 'src/utils/config'
+
+import Wizard from '../Wallet/Wizard'
```
RafaelTaranto: I would rather we never have to use `../`, since this is a import from something outside of the module we should make them full named imports like the ones above.

new-lamassu-admin/src/components/editableTable/Row.js
```diff
@@ -174,7 +174,8 @@ const ERow = ({ editing, disabled, lastOfGroup }) => {
     enableDelete,
     enableToggle,
     rowSize,
-    stripeWhen
+    stripeWhen,
+    focusOnEditWhen
```
RafaelTaranto: I don't think we really need focusOnEditWhen.

new-lamassu-admin/src/components/editableTable/Row.js
```diff
@@ -213,7 +214,12 @@ const ERow = ({ editing, disabled, lastOfGroup }) => {
             key={idx}
             config={it}
             editing={editing}
-            focus={idx === elementToFocusIndex && editing}
+            focus={
+              idx === elementToFocusIndex &&
+              editing &&
+              focusOnEditWhen &&
+              focusOnEditWhen(values)
+            }
```
RafaelTaranto: Logic here would disable focus for every other table.

RafaelTaranto (changes requested):

new-lamassu-admin/src/pages/Locales/Locales.js
RafaelTaranto: Still here.

RafaelTaranto (changes requested):

RafaelTaranto (approved):

PR #442, opened 2020-09-25
fix: v7.5 - Don't include 'bitcoincash:' URI in admin text

liordino:
fix: removed the 'bitcoincash:' substring from bch addresses on the
admin


liordino:
[v7.5 - Don't include 'bitcoincash:' URI in admin text](https://trello.com/c/CgzvICUs/216-v75-dont-include-bitcoincash-uri-in-admin-text)

RafaelTaranto (approved):

PR #443, opened 2020-09-28
fix: remove lamassu-update

CrypticaScriptura:
We control the update process outside of versioned lamassu-server scripts, as otherwise update scripts would rarely be up-to-date. Removes a foot-shooter.


RafaelTaranto (approved):

RafaelTaranto (changes requested):
@naconner we probably want to merge this into dev, I don't think we'll be releasing another 7.4 version just for this change!

RafaelTaranto (approved):

PR #445, opened 2020-09-30
fix: Fiat currency on locales showing empty option

liordino:
fix: filter out the empty code currencies from the list on locales

### DIFF
--- a/new-lamassu-admin/src/pages/Locales/helper.js
+++ b/new-lamassu-admin/src/pages/Locales/helper.js
@@ -73,7 +73,7 @@ const allFields = (getData, auxElements = []) => {
       view: getView(currencyData, 'code'),
       input: Autocomplete,
       inputProps: {
-        options: currencyData?.filter(c => c.code !== ''),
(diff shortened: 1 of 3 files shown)


liordino:
[Fiat currency on locales showing empty option](https://trello.com/c/nervZZbf/213-fiat-currency-on-locales-showing-empty-option)

new-lamassu-admin/src/pages/Locales/helper.js
```diff
@@ -73,7 +73,7 @@ const allFields = (getData, auxElements = []) => {
       view: getView(currencyData, 'code'),
       input: Autocomplete,
       inputProps: {
-        options: currencyData,
+        options: currencyData?.filter(c => c.code !== ''),
```
RafaelTaranto: Move the filter to the massageCurrencies server-side code.

RafaelTaranto (changes requested):

RafaelTaranto (approved):

PR #446, opened 2020-10-01
Fix: Make "save" always show up first on forms

liordino:
fix: make save buttons always shows up prior to cancel buttons

fix: make all save buttons of submit type


liordino:
[Make "save" always show up first on forms](https://trello.com/c/pfVQFKJB/234-make-save-always-show-up-first-on-forms)

RafaelTaranto (approved):

PR #450, opened 2020-10-05
feat: Add a rename machine action.

liordino:
feat: created the rename action on the machine status page

fix: added missing 'shutdown' action on the MachineActions enum of the
gql schema

style: set the Edit icon for the rename action

style: fixed first and last action buttons spacing

### DIFF
--- a/lib/machine-loader.js
+++ b/lib/machine-loader.js
@@ -45,31 +45,25 @@ function getMachineNames (config) {
       [machines, notifier.checkPings(machines), dbm.machineEvents(), config]
     ))
     .then(([machines, pings, events, config]) => {
-      const getPingStatus = (ping) => {
-        if (!ping) return fullyFunctionalStatus
+      const getStatus = (ping, stuck) => {
+        if (stuck && stuck.age) return stuckStatus
 
-        if (ping.age) return unresponsiveStatus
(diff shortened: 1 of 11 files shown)


RafaelTaranto:
[Add a "rename" machine action.](https://trello.com/c/6Us4AQ6D/235-add-a-rename-machine-action)

lib/machine-loader.js
```diff
@@ -34,15 +36,40 @@ function getConfig (defaultConfig) {
 }
 
 function getMachineNames (config) {
+  const fullyFunctionalStatus = {label: 'Fully functional', type: 'success'}
```
RafaelTaranto, line 39: Hmm, thinking more about this we just want fully functional if we're not stuck and the machine is pinging. There's no reason to show the operator "Fully functional" and "Stuck" at the same time.

RafaelTaranto (changes requested):
There also conflicts to resolve

lib/machine-loader.js
RafaelTaranto: Just noticed that this is from another PR. Fix that there and rebase. This also has conflicts

RafaelTaranto (approved):

PR #451, opened 2020-10-06
Fix: Make every field of the notifications page optional

liordino:
fix: make table fields show a '-' instead of an empty string when the
col content is null

fix: make all notifications fields optional

fix: make the notifications overrides fields valid only if the machine/
cryptocurrency and at least one of the other options are filled

### DIFF
--- a/new-lamassu-admin/src/components/editableTable/Row.js
+++ b/new-lamassu-admin/src/components/editableTable/Row.js
@@ -102,7 +102,7 @@ const ECol = ({
     textAlign,
     suffix,
     SuffixComponent = TL2,
-    view = it => it?.toString(),
+    view = it => (!it ? '-' : it.toString()),
     inputProps = {}
   } = config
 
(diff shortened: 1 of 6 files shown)


liordino:
[Make every field of the notifications page optional](https://trello.com/c/LW3sGru0/238-make-every-field-of-the-notifications-page-optional)

RafaelTaranto (approved):

new-lamassu-admin/src/components/editableTable/Row.js
```diff
@@ -102,7 +102,7 @@ const ECol = ({
     textAlign,
     suffix,
     SuffixComponent = TL2,
-    view = it => it?.toString(),
+    view = it => (!it ? '-' : it.toString()),
```
RafaelTaranto: Lets leave with empty string for now.

RafaelTaranto (changes requested):

RafaelTaranto (approved):

PR #452, opened 2020-10-07
fix: avoid warning on first fiat currency selection

liordino:
fix: avoid warning on first fiat currency selection


liordino:
[Add warning on locale when changing fiat currency](https://trello.com/c/KIxYQQWC/185-add-warning-on-locale-when-changing-fiat-currency)

RafaelTaranto (approved):

PR #456, opened 2020-10-11
fix: pending value on funding page

RafaelTaranto:


RafaelTaranto:
[v7.5 - Funding panel gives erroneous 'pending' amount](https://trello.com/c/IDzMAQca/228-v75-funding-panel-gives-erroneous-pending-amount)

PR #461, opened 2020-10-16
fix: suspend when no ff camera

RafaelTaranto:


RafaelTaranto:
[Implement fallback for front-facing camera on 7.5](https://trello.com/c/IAg5zmtV/257-implement-fallback-for-front-facing-camera-on-75)

PR #462, opened 2020-10-16
fix: commissions overrides and coinatmradar

RafaelTaranto:


RafaelTaranto:
[CoinATMRadar](https://trello.com/c/8MsWh92R/261-coinatmradar)

PR #466, opened 2020-10-22
fix: add empty table text

RafaelTaranto:


RafaelTaranto:
[Empty response indication on screens](https://trello.com/c/3BWQaRao/265-empty-response-indication-on-screens)

PR #468, opened 2020-10-22
Fix/chromium issues

RafaelTaranto:


RafaelTaranto:
[Table with multi level headers alignment issue](https://trello.com/c/mNKn2NVG/186-table-with-multi-level-headers-alignment-issue)

PR #469, opened 2020-10-23
fix: overrides stuck on infinite loop if blocked

RafaelTaranto:


RafaelTaranto:
[Overrides stuck in a infinite loop if blocked](https://trello.com/c/qfN1LLSJ/279-overrides-stuck-in-a-infinite-loop-if-blocked)

PR #471, opened 2020-10-23
Fix/wizard regression

RafaelTaranto:


RafaelTaranto:
[Neal bugs quick feedback card](https://trello.com/c/ob1JTdxo/283-neal-bugs-quick-feedback-card)

PR #474, opened 2020-10-25
fix: Fix prompt when dirty for editable tables

liordino:
fix: prompt when leaving page without saving changes on editable tables


liordino:
[Prevent leaving the page without saving](https://trello.com/c/fqVxcOWq/189-prevent-leaving-the-page-without-saving)

PR #475, opened 2020-10-26
feat: Dropdowns should change to previous value if user closes configuration popup without saving

liordino:
fix: keep previous value when closing the configuration popup without
saving on locales screen

refactor: renamed enableCoin function to configureCoin

fix: keep previous value when closing the configuration popup without
saving on wallets screen

refactor: renamed enableThirdPartyService function to
configureThirdPartyService


liordino:
[Dropdowns should change to previous value if user closes configuration popup without saving](https://trello.com/c/Dgd5vKmZ/290-dropdowns-should-change-to-previous-value-if-user-closes-configuration-popup-without-saving)

RafaelTaranto (approved):

PR #480, opened 2020-10-28
fix: Cash-out cassette denominations are transposed in admin

liordino:
fix: renamed column headers to 'Cassete' instead of 'Cashout'

fix: cassete denominations were inverted


liordino:
[Cash-out cassette denominations are transposed in admin](https://trello.com/c/ZigEyeQK/304-cash-out-cassette-denominations-are-transposed-in-admin)

RafaelTaranto (approved):

PR #481, opened 2020-10-28
fix: Replace extracted Country info in admin's Customer page with State/Province, add birthdate

liordino:
fix: replaced country info with state/province and added birth date on
the id data card

### DIFF
--- a/new-lamassu-admin/src/pages/Customers/components/IdDataCard.js
+++ b/new-lamassu-admin/src/pages/Customers/components/IdDataCard.js
@@ -15,6 +15,7 @@ import Field from './Field'
 const IdDataCard = memo(({ customerData, updateCustomer }) => {
   const idData = R.path(['idCardData'])(customerData)
   const rawExpirationDate = R.path(['expirationDate'])(idData)
+  const country = R.path(['country'])(idData)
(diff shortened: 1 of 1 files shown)


liordino:
[Replace extracted Country info in admin's Customer page with State/Province, add birthdate](https://trello.com/c/EKXJBFff/297-replace-extracted-country-info-in-admins-customer-page-with-state-province-add-birthdate)

new-lamassu-admin/src/pages/Customers/components/IdDataCard.js
```diff
@@ -23,30 +24,35 @@ const IdDataCard = memo(({ customerData, updateCustomer }) => {
...
     },
     {
       header: 'ID number',
       display: R.path(['documentNumber'])(idData),
-      size: 190
+      size: 160
+    },
+    {
+      header: 'Birth Date',
+      display: ifNotNull(rawDob, moment.utc(rawDob).format('YYYY-MM-D')),
```
RafaelTaranto: We probably want DD here instead of just D

RafaelTaranto (changes requested):

RafaelTaranto (approved):

PR #482, opened 2020-10-28
fix: Reject / Authorise buttons in compliance overrides have no effect

liordino:
chore: removed unused variable from ComplianceDetails

fix: fixed the customer mutation, which was causing errors on all
customer compliance buttons


liordino:
[Reject / Authorise buttons in compliance overrides have no effect](https://trello.com/c/xfnpI76i/299-reject-authorise-buttons-in-compliance-overrides-have-no-effect)

RafaelTaranto (approved):

PR #483, opened 2020-10-28
fix: Remove support logs related code from server

liordino:
fix: removed unnecessary support log code from server


liordino:
[Remove support logs related code from server](https://trello.com/c/hGYVXHCS/305-remove-support-logs-related-code-from-server)

RafaelTaranto (approved):

PR #486, opened 2020-10-29
fix: Maintenance - cancelling cassette bill count edit resets both cassettes to 0

liordino:
fix: refetch cassetes data to avoid resetting to the initial value on UI
when canceling edit


liordino:
[Maintenance - cancelling cassette bill count edit resets both cassettes to 0](https://trello.com/c/SqWGZ4fZ/329-maintenance-cancelling-cassette-bill-count-edit-resets-both-cassettes-to-0)

RafaelTaranto (changes requested):
Why don't we just return it to the previous value? I don't see why we should refetch something that didn't change.

Edit: I misunderstood the fix!

RafaelTaranto (approved):

PR #488, opened 2020-10-30
fix: receipt printing new admin

RafaelTaranto:


RafaelTaranto:
[Receipt printing happening on v7.4, but not on v7.5](https://trello.com/c/hVnUw5NQ/326-receipt-printing-happening-on-v74-but-not-on-v75)

PR #489, opened 2020-10-30
Feat/csv logs

RafaelTaranto:


RafaelTaranto:
[v7.5 admin download of machine, server, transaction logs should be formatted in CSV, rather than JSON](https://trello.com/c/KAChW7rQ/317-v75-admin-download-of-machine-server-transaction-logs-should-be-formatted-in-csv-rather-than-json)

PR #490, opened 2020-10-30
fix: overrides should not merge child arrays

RafaelTaranto:


RafaelTaranto:
[Locales override not disabling cryptocurrency](https://trello.com/c/6ZOiDtP1/321-locales-override-not-disabling-cryptocurrency)

PR #491, opened 2020-11-01
fix: High-value transaction notifications not triggering

liordino:
fix: high value notifications, if configured, are now triggered even if
transactions notifications are globally disabled (for sms or email)

fix: added extra information on the sms notification informing if the
transaction that just happened is a high value one


liordino:
[High-value transaction notifications not triggering](https://trello.com/c/zDsXeVWt/322-high-value-transaction-notifications-not-triggering)

RafaelTaranto (approved):

PR #494, opened 2020-11-03
fix: 'Shutdown' admin button not shutting down machine

liordino:
feat: added shutdown function to the machine loader

feat: created shutdown route and call it from gql server (called from
admin UI button)


liordino:
['Shutdown' admin button not shutting down machine](https://trello.com/c/4GdLCkw0/320-shutdown-admin-button-not-shutting-down-machine)

RafaelTaranto (approved):

PR #495, opened 2020-11-03
fix: Percentage number in cash cassette is the same color as the background when the bill counts go below the red threshold

liordino:
fix: move number alongside bar on empty part until 51%, and them move it
to the filled part


liordino:
[Percentage number in cash cassette is the same color as the background when the bill counts go below the red threshold](https://trello.com/c/4WPYp8qB/330-percentage-number-in-cash-cassette-is-the-same-color-as-the-background-when-the-bill-counts-go-below-the-red-threshold)

RafaelTaranto (approved):

PR #496, opened 2020-11-05
Feat/all coins on commissions

RafaelTaranto:


RafaelTaranto:
[Exceptions to commissions require one to specify coins, which should be optional](https://trello.com/c/9FHUhvhb/318-exceptions-to-commissions-require-one-to-specify-coins-which-should-be-optional)

PR #497, opened 2020-11-05
fix: hide navbar before user authenticates

RafaelTaranto:


RafaelTaranto:
[hide the navbar while user isn't properly authenticated](https://trello.com/c/snk0VFy0/333-hide-the-navbar-while-user-isnt-properly-authenticated)

PR #504, opened 2020-11-06
fix: wizard locales regression

RafaelTaranto:


RafaelTaranto:
[Wizard locales not allowing selection of cryptocurrency](https://trello.com/c/ATDNrReu/349-wizard-locales-not-allowing-selection-of-cryptocurrency)

PR #509, opened 2020-11-11
feat: Search component PT3: add support for multiple filters

liordino:
feat: allow multiple filter selection on the search component


liordino:
[Search component PT3: add support for multiple filters](https://trello.com/c/7LyMgtmc/365-search-component-pt3-add-support-for-multiple-filters)

PR #513, opened 2020-11-16
feat: create blacklist page

hellypi:


lib/new-admin/graphql/schema.js
```diff
@@ -246,6 +260,8 @@ const typeDefs = gql`
     createPairingTotem(name: String!): String
     saveAccounts(accounts: JSONObject): JSONObject
     revokeToken(token: String!): UserToken
+    deleteBlacklistRow(cryptoCode: String, address: String): Blacklist
```
RafaelTaranto: cryptoCode and address should be required

new-lamassu-admin/src/pages/Blacklist/Blacklist.js
```diff
@@ -0,0 +1,183 @@
...
+  const toggleModal = () => setShowModal(!showModal)
+
+  const addToBlacklist = (cryptoCode, address) => {
+    addEntry({ variables: { cryptoCode, address } })
+  }
+
+  return (
+    <>
+      <TitleSection title="Blacklisted addresses">
+        <Link onClick={toggleModal}>Blacklist new addresses</Link>
```
RafaelTaranto: Just like we don't want toggle on closing the modal, we don't want it here. This button should only set the modal to true. 

`() => setShowModal(true)`

RafaelTaranto (changes requested):

RafaelTaranto:
[Create the blacklist screen](https://trello.com/c/OjxjUUH7/165-create-the-blacklist-screen)

RafaelTaranto (approved):

PR #516, opened 2020-11-17
[WIP] feat: commissions list

jsousa127:
Missing save to server


jsousa127:
[Commission list PT3: Create the filters](https://trello.com/c/SYRGQsmx/362-commission-list-pt3-create-the-filters)

chaotixkilla:
[Page crash when opening the /system/token-management page, if a user had an old token set, before this page was made. The reason of the crash was because the user_agent field was undefined in the old token.](https://trello.com/c/nNbfKusf/387-page-crash-when-opening-the-system-token-management-page-if-a-user-had-an-old-token-set-before-this-page-was-made-the-reason-of)

PR #518, opened 2020-11-17
feat: support shielded zec sends

CrypticaScriptura:
Depends on https://github.com/lamassu/lamassu-machine/pull/555

### DIFF
--- a/lib/plugins/wallet/zcashd/zcashd.js
+++ b/lib/plugins/wallet/zcashd/zcashd.js
@@ -46,11 +46,42 @@ function balance (account, cryptoCode) {
 
 function sendCoins (account, address, cryptoAtoms, cryptoCode) {
   const coins = cryptoAtoms.shift(-unitScale).toFixed(8)
-
+  const checkSendStatus = function (opid) {
+    return new Promise((resolve, reject) => {
+      function checker(timeout) {
+        setTimeout(() => {
+          fetch('z_getoperationstatus', [[opid]])
+            .then(res => {
+              const status = _.get('status', res[0])
+              switch(status) {
+                case 'success': 
+                  resolve(res[0])
+                  break
+                case 'failed': 
+                  reject(res[0].error)
+                  break
+                case 'executing':
+                  if (timeout > 0) {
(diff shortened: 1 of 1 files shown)


RafaelTaranto (approved):

lib/plugins/wallet/zcashd/zcashd.js
```diff
@@ -46,11 +46,42 @@ function balance (account, cryptoCode) {
...
+                  if (timeout > 0) {
+                    checker(timeout - 500)
+                  } else {
+                    reject(res)
+                  }
+              }
+            })
+        }, 500)
+      }
+      checker(10000)
```
joshmh, line 73: Shouldn't this be ``return checker(10000)``?

lib/plugins/wallet/zcashd/zcashd.js
```diff
@@ -46,11 +46,42 @@ function balance (account, cryptoCode) {
...
+              switch(status) {
+                case 'success': 
+                  resolve(res[0])
+                  break
+                case 'failed': 
+                  reject(res[0].error)
+                  break
+                case 'executing':
+                  if (timeout > 0) {
+                    checker(timeout - 500)
```
joshmh, line 65: Do we want a recursive call here? Could get messy.

joshmh (changes requested):

lib/plugins/wallet/zcashd/zcashd.js
CrypticaScriptura: The `Promise` constructor expects some executor function that returns void, so we don't need to return anything from this function. The `checker` function returns whatever `setTimeout` returns, which I believe is an ID that identifies the timeout, so that it could be canceled somewhere else if needed, but this is not what we need.

lib/plugins/wallet/zcashd/zcashd.js
CrypticaScriptura: True, could get messy. Suggestions for writing a retry-until-time-limit thing in a promise chain without using recursion?

PR #519, opened 2020-11-17
fix: fudge factor text(crypto atoms)

jsousa127:


jsousa127:
[Fudge factor text says 10 crypto atoms but we use 100](https://trello.com/c/Mth7KRJr/368-fudge-factor-text-says-10-crypto-atoms-but-we-use-100)

RafaelTaranto (approved):

PR #521, opened 2020-11-19
fix: user_agent nullable

chaotixkilla:


chaotixkilla:
[Page crash when opening the /system/token-management page, if a user had an old token set, before this page was made. The reason of the crash was because the user_agent field was undefined in the old token.](https://trello.com/c/nNbfKusf/387-page-crash-when-opening-the-system-token-management-page-if-a-user-had-an-old-token-set-before-this-page-was-made-the-reason-of)

RafaelTaranto (approved):

PR #522, opened 2020-11-21
chore: add georgian as support language

RafaelTaranto:


RafaelTaranto:
[Add Georgian to the support language list](https://trello.com/c/76JPoQsR/395-add-georgian-to-the-support-language-list)

PR #527, opened 2020-11-26
Feat/warn admin before restart services

hellypi:
Admin gets a warning in the confirm dialog if they try to restart the machine services while a user is in the middle of a transaction

### DIFF
--- a/lib/postgresql_interface.js
+++ b/lib/postgresql_interface.js
@@ -1,4 +1,6 @@
+const _ = require('lodash/fp')
 const db = require('./db')
+const pgp = require('pg-promise')()
 
 function getInsertQuery (tableName, fields) {
   // outputs string like: '$1, $2, $3...' with proper No of items
@@ -46,6 +48,16 @@ exports.machineEvent = function machineEvent (rec) {
 
   return db.none(sql, values)
     .then(() => db.none(deleteSql))
+}
+
+exports.machineEventsByIdBatch = function machineEventsByIdBatch (machineIds) {
+  const formattedIds = _.map(pgp.as.text, machineIds).join(',')
+  const sql = `SELECT *, (EXTRACT(EPOCH FROM (now() - created))) * 1000 AS age FROM machine_events WHERE device_id IN ($1^) ORDER BY age ASC LIMIT 1`
+  return db.any(sql, [formattedIds]).then(res => {
+    const events = _.map(_.mapKeys(_.camelCase))(res)
+    const eventMap = _.groupBy('deviceId', events)
+    return machineIds.map(id => _.prop([0], eventMap[id]))
+  })
 }
 
 exports.machineEvents = function machineEvents () {
--- a/new-lamassu-admin/src/pages/Maintenance/MachineDetailsCard.js
+++ b/new-lamassu-admin/src/pages/Maintenance/MachineDetailsCard.js
@@ -3,7 +3,7 @@ import { Grid, Divider } from '@material-ui/core'
 import { makeStyles } from '@material-ui/core/styles'
 import gql from 'graphql-tag'
 import moment from 'moment'
-import React, { useState, useEffect } from 'react'
+import React, { useState } from 'react'
 
 import { ConfirmDialog } from 'src/components/ConfirmDialog'
 import { Status } from 'src/components/Status'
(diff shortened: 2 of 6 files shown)


hellypi:
[Restart services admin action should warn user if the machine is amidst a transaction.](https://trello.com/c/WUihMC3g/340-restart-services-admin-action-should-warn-user-if-the-machine-is-amidst-a-transaction)

hellypi:
[Restart services warning PT2: optimize graphql usage](https://trello.com/c/8avGtnIR/409-restart-services-warning-pt2-optimize-graphql-usage)

chaotixkilla (commented):
Everything seems fine, but how beneficial is dataloader? I have no experience with it, so I don't really have an informed opinion about it. Just seems odd to use it for this specific case

lib/postgresql_interface.js
```diff
@@ -48,6 +49,15 @@ exports.machineEvent = function machineEvent (rec) {
     .then(() => db.none(deleteSql))
 }
 
+exports.machineEventsByIdBatch = function machineEventsByIdBatch (machineIds) {
+  const sql = `SELECT *, (EXTRACT(EPOCH FROM (now() - created))) * 1000 AS age FROM machine_events WHERE device_id LIKE ANY($1) ORDER BY age ASC LIMIT 1`
```
RafaelTaranto: change the `like any` to `in`

new-lamassu-admin/package.json
```diff
@@ -23,6 +23,7 @@
     "graphql-tag": "^2.10.3",
     "jss-plugin-extend": "^10.0.0",
     "libphonenumber-js": "^1.7.50",
+    "loader": "^2.1.1",
```
RafaelTaranto: Undo this change and the package-lock for the new-lamassu-admin package.

new-lamassu-admin/src/pages/Maintenance/MachineDetailsCard.js
```diff
@@ -69,22 +79,82 @@ const Item = ({ children, ...props }) => (
...
+      variables: {
+        deviceId: machine.deviceId
+      },
+      onCompleted: machineEventsLazy => {
+        if (machineEventsLazy.machine.latestEvent) {
+          setMachineState(
+            JSON.parse(machineEventsLazy.machine.latestEvent.note).state
+          )
+        }
+      }
```
RafaelTaranto: I don't think we need this to be on a `onCompleted` nor we need the `machineState` state. Doesn't the query hook return a response?

new-lamassu-admin/src/pages/Maintenance/MachineDetailsCard.js
```diff
@@ -69,22 +79,82 @@ const Item = ({ children, ...props }) => (
...
+        'pendingIdle',
+        'dualIdle',
+        'networkDown',
+        'unpaired',
+        'maintenance',
+        'virgin',
+        'wifiList'
+      ]
+      return staticStates.includes(machineState)
+    }
```
RafaelTaranto: You can pass `machineState` as a variable and make this a pure function. Move it to the outside of this `MachineDetailRow` component.

new-lamassu-admin/src/pages/Maintenance/MachineDetailsCard.js
```diff
@@ -69,22 +79,82 @@ const Item = ({ children, ...props }) => (
...
+    if (action.command === 'restartServices') {
+      setAction(action => ({
+        ...action,
+        ...(!isStaticState() && {
+          message: (
+            <span className={classes.warning}>
```
RafaelTaranto: You can create a message object outside of the effect hook.This would also remove `classes.warning` from the dependency array.

new-lamassu-admin/src/pages/Maintenance/MachineDetailsCard.js
```diff
@@ -127,25 +197,26 @@ const MachineDetailsRow = ({ it: machine, onActionSuccess }) => {
           className={classes.separator}
         />
         <ConfirmDialog
+          disabled={disabled}
           open={confirmDialogOpen}
-          title={`${action?.command} this machine?`}
+          title={`${action?.humanReadable} this machine?`}
```
RafaelTaranto: We can change `humanReadable` to `display`, didn't we do this before?

new-lamassu-admin/src/pages/Maintenance/MachineDetailsCard.js
```diff
@@ -69,22 +79,82 @@ const Item = ({ children, ...props }) => (
...
+
+  useEffect(() => {
+    if (action.command === 'restartServices') {
+      fetchMachineEvents()
+    }
+  }, [action.command, fetchMachineEvents])
+
+  // on slow connections, we could have the confirm dialog open before the query is fetched from the server
+  // this would make the warning not appear in some situations
+  // so, have to check if we should update the message on completion of the fetch query
```
RafaelTaranto: Is this comment still relevant? Didn't we add the behavior to make the button disabled?

RafaelTaranto (changes requested):

new-lamassu-admin/src/pages/Maintenance/MachineDetailsCard.js
hellypi: yup, but that was in the branch to add the button, this one uses the button but is a different branch. I'll change it here as well

hellypi:
> 
> 
> Everything seems fine, but how beneficial is dataloader? I have no experience with it, so I don't really have an informed opinion about it. Just seems odd to use it for this specific case

DataLoader is important to fix n+1 problem:
When we fetch every machine, and we also want to fetch the latest event for each machine, because of our usage of resolvers we will actually make an additional query for each machine.
The DataLoader is a batching mechanism where we store every fetch in a promess and do it all at once, reducing n fetches to just 1.
https://medium.com/the-marcy-lab-school/what-is-the-n-1-problem-in-graphql-dd4921cb3c1a

@chaotixkilla

new-lamassu-admin/src/pages/Maintenance/MachineDetailsCard.js
```diff
@@ -69,22 +97,63 @@ const Item = ({ children, ...props }) => (
...
+    variables: {
+      deviceId: machine.deviceId
+    }
+  })
+
+  useEffect(() => {
+    if (action.command === 'restartServices') {
+      fetchMachineEvents()
+    }
+  }, [action.command, fetchMachineEvents])
```
RafaelTaranto: since fetchMachineEvents is a common local variable we don't really need a use effect here.

new-lamassu-admin/src/pages/Maintenance/MachineDetailsCard.js
```diff
@@ -69,22 +97,63 @@ const Item = ({ children, ...props }) => (
...
+        }))
+      } else {
+        // clear message from object when state goes from not static to static
+        setAction(action => ({
+          ...action,
+          message: null
+        }))
+      }
+    }
+  }, [action.command, classes.warning, machineEventsLazy])
```
RafaelTaranto: Ok, now it's easier to rework this part of the code. 

- Since machineEventsLazy is a common variable we can skip using useEffect altogether.
- Move the span to a variable
- Since the only difference between the if/else is message, might as well do a ternary if for a message const

RafaelTaranto (changes requested):

RafaelTaranto (approved):

PR #533, opened 2020-12-03
Show error on hover in the transactions screen

chaotixkilla:


chaotixkilla:
[Show error on hover in the transactions screen](https://trello.com/c/naj42qoi/319-show-error-on-hover-in-the-transactions-screen)

new-lamassu-admin/src/pages/Transactions/helper.js
```diff
@@ -21,11 +18,7 @@ const getCashInStatus = it => {
 }
 
 const getCashInStatusDetails = it => {
```
liordino: The getCashInStatusDetails and getCashOutStatusDetails functions have the same body. Can turn these into a single function

new-lamassu-admin/src/components/Tooltip.js
```diff
@@ -51,4 +51,35 @@ const Tooltip = memo(({ children, width, Icon = HelpIcon }) => {
   )
 })
 
-export default Tooltip
+const HoverableTooltip = memo(({ parentElements, children, width }) => {
```
liordino, line 66: There's a good chunk of repeated code within Tooltip and HoverableTooltip. Creating another component used by both of them with the common parts or parameterizing the behaviour could clean it up.

liordino (changes requested):

new-lamassu-admin/src/components/Tooltip.js
chaotixkilla: I agree, My solution was creating a hook which can be used by other Tooltip-like components, regarding the popper handling

new-lamassu-admin/src/pages/Transactions/helper.js
chaotixkilla: Although the bodies are the same, I believe it is important to keep them separated for readability's sake and as they serve different directions of transactions, their logic should be kept separated, despite being equal, for now.

Tell me what you think about this and if my argument has changed your mind about this

liordino: I think that we can keep a single one while the logic is one and separate when they become different things in fact, if this ever occurs (may as well be never). By doing this we avoid, for example, updating one function while forgeting to change the other, which can cause some hard to track bugs. I agree that some readability may be lost in the process (even though while having only one behaviour keeping a single function is simpler and represents the current truth), but we avoid adding unnecessary complexity to the code, which can add up fast.

liordino (changes requested):
Didn't noticed that before, but needs to sign the commits.

chaotixkilla:
@liordino All 4 commits are showing up as signed to me, can you please double-check?

liordino:
> @liordino All 4 commits are showing up as signed to me, can you please double-check?

@chaotixkilla you're right, my mistake. It shows the signature as invalid in sublime merge for me, and I trusted it, but it's indeed signed.

liordino (approved):

RafaelTaranto (approved):

PR #536, opened 2020-12-14
Override fields update fix

chaotixkilla:


RafaelTaranto:
[Updating customers is not updating override_by and override_at fields](https://trello.com/c/jylhNiy2/453-updating-customers-is-not-updating-overrideby-and-overrideat-fields)

RafaelTaranto (approved):

shivchawla:
I don't if it's relevant here but please take a look.

After rejecting the customer ID photo once from Admin panel, customer is not able to update the ID card. Is this expected? Shouldn't the customer be allowed to update the ID Photo and override is set to 'automatic'? Or it this happening somewhere else?

Capture

PR #538, opened 2020-12-16
fix: Compliance triggers allowing to create a suspend trigger with no days set

liordino:
fix: make the number of suspension days required only when the
'Suspend' requirement is selected


liordino:
[Compliance triggers allowing to create a suspend trigger with no days set](https://trello.com/c/YMGXO9QA/411-compliance-triggers-allowing-to-create-a-suspend-trigger-with-no-days-set)

RafaelTaranto (approved):

PR #544, opened 2020-12-21
fix/screen height problem

jsousa127:


RafaelTaranto:
[fix screen-height issues](https://trello.com/c/inFGrqpH/484-fix-screen-height-issues)

PR #562, opened 2021-01-11
Fix: make OperatorInfo receipts table reset values on cancel

RafaelTaranto:


RafaelTaranto:
[Boolean tables not updating when clicking 'cancel'](https://trello.com/c/bkH397sE/501-boolean-tables-not-updating-when-clicking-cancel)

PR #564, opened 2021-01-14
Fix machine crashing when a cassette denomination is set to zero

chaotixkilla:


chaotixkilla:
[l-s continually errors if cassette denomination is set to '0', rendering machines down](https://trello.com/c/PpTupVxm/512-l-s-continually-errors-if-cassette-denomination-is-set-to-0-rendering-machines-down)

hellypi (approved):

RafaelTaranto (approved):

PR #565, opened 2021-01-14
Server stress testing

chaotixkilla:


chaotixkilla:
[Stress testing l-s PT1: Testing idle machines](https://trello.com/c/KAL482V2/509-stress-testing-l-s-pt1-testing-idle-machines)

chaotixkilla:
[Stress testing l-s https requests](https://trello.com/c/oPcPedVg/490-stress-testing-l-s-https-requests)

test/stress/child.js
```diff
@@ -0,0 +1,78 @@
...
+function getCert (machineIndex) {
+  try {
+    return {
+      key: fs.readFileSync(path.resolve(__dirname, 'machines', `${machineIndex}`, 'client.key')),
+      cert: fs.readFileSync(path.resolve(__dirname, 'machines', `${machineIndex}`, 'client.pem'))
+    }
+  } catch (e) {
+    return null
+  }
+}
```
hellypi: better avoid blocking synchronous code whenever we can, try something like this instead:

```suggestion
const pify = require('pify')
const fs = pify(require('fs'))
function getCert (machineIndex) {
  return fs.readFile(...).then(file => ...).catch("you can also catch this somewhere else")
}
```

test/stress/child.js
```diff
@@ -0,0 +1,78 @@
...
+
+function connectionInfo (machineIndex) {
+  try {
+    return JSON.parse(fs.readFileSync(path.resolve(__dirname, 'machines', `${machineIndex}`, 'connection_info.json')))
```
hellypi: same as above

hellypi (changes requested):

test/stress/child.js
chaotixkilla: My initial idea was really to keep the blocking going. This is all running as a single script, and I kind of need the machine creation to be predictable. I'll ask Rafael about it tho

chaotixkilla: Just confirmed with Rafael, he believes it is worth the machine creation to be asynchronous. Working on it!

test/stress/child.js
```diff
@@ -0,0 +1,87 @@
...
+var certificate = {}
+var connectionInfo = {}
+
+async function getCert (machineIndex) {
+  const key = fs.readFile(path.resolve(__dirname, 'machines', `${machineIndex}`, 'client.key'))
+  const cert = fs.readFile(path.resolve(__dirname, 'machines', `${machineIndex}`, 'client.pem'))
+
+  return Promise.all([key, cert]).then(values => {
+    return { key: values[0], cert: values[1] }
+  }).catch(err => {
```
hellypi: what do you think of this?
```suggestion
  return Promise.all([key, cert]).then([key, cert]=> {
    return { key, cert }
  }).catch(err => {
```

test/stress/child.js
```diff
@@ -0,0 +1,87 @@
...
+const _ = require('lodash/fp')
+const { PerformanceObserver, performance } = require('perf_hooks')
+
+const utils = require('./utils')
+const variables = require('./utils/variables')
+
+var certificate = {}
+var connectionInfo = {}
+
+async function getCert (machineIndex) {
```
hellypi: no need for async anymore

test/stress/child.js
```diff
@@ -0,0 +1,87 @@
...
+  const key = fs.readFile(path.resolve(__dirname, 'machines', `${machineIndex}`, 'client.key'))
+  const cert = fs.readFile(path.resolve(__dirname, 'machines', `${machineIndex}`, 'client.pem'))
+
+  return Promise.all([key, cert]).then(values => {
+    return { key: values[0], cert: values[1] }
+  }).catch(err => {
+    console.err('The following error when reading the certificate: ', err)
```
hellypi: did you mean console.error?

test/stress/child.js
```diff
@@ -0,0 +1,87 @@
...
+    return null
+  })
+}
+
+async function getConnectionInfo (machineIndex) {
```
hellypi: no need for async
also I think you can do this:
```suggestion
const getConnectionInfo = (machineIndex) => fs.readFile(path.resolve(__dirname, 'machines', `${machineIndex}`, 'connection_info.json'))
```

test/stress/child.js
```diff
@@ -0,0 +1,87 @@
...
+
+process.on('message', async (msg) => {
+  console.log('Message from parent:', msg)
+
+  const promises = [getCert(msg.machineIndex), getConnectionInfo(msg.machineIndex)]
+  Promise.all(promises).then(values => {
+    certificate = values[0]
+    connectionInfo = JSON.parse(values[1])
+  }).catch(err => {
+    console.err('The following error occurred during certificate parsing: ', err)
```
hellypi: console.error

hellypi (changes requested):
Other than the ones I marked, you have several functions marked as async. Since this is a test script I'm not sure this is a problem but I've been told that we should avoid blocking code (and thus, avoid using async functions).

hellypi (approved):

test/stress/scripts/create-machines.sh
```diff
@@ -0,0 +1,62 @@
...
+
+  # Get device_id
+  DEVICE_ID=`openssl x509 -outform der -in ./machines/$NUMBER/client.pem | sha256sum | cut -d " " -f 1`
+
+  # Update db config
+  sudo -u postgres psql -t -d lamassu_stress -c "select data from user_config where type='config' order by id desc limit 1" > config.json
+  NEW_CONFIG=$(node ./utils/save-config.js $NUMBER $DEVICE_ID)
+  sudo -u postgres psql "lamassu_stress" << EOF
+    insert into user_config(type, data, created, valid) 
+    values('config', '$NEW_CONFIG', now(), 't')
```
RafaelTaranto: Is this necessary? Won't lamassu_stress be empty?

RafaelTaranto (changes requested):

test/stress/scripts/create-machines.sh
chaotixkilla: hmm, you are correct if the server is in a virgin state (only migration performed, no configuration whatsoever done). If the server was already configured with the initial wizard, it will work properly. I think I will implement a default backup config, for the flow to not break

RafaelTaranto (approved):

PR #566, opened 2021-01-15
Pass raw ID data

chaotixkilla:


chaotixkilla:
[Passing along the raw data for IDs](https://trello.com/c/68Vbflqn/513-passing-along-the-raw-data-for-ids)

lib/customers.js
```diff
@@ -281,6 +281,18 @@ function enhanceAtFields (fields) {
...
+    'id_card_data'
+  ]
+
+  const fieldsToUpdate = _.pick(fieldsWithRawData, fields)
+  const rawFields = _.fromPairs(_.map(f => [`${f}_raw`, JSON.stringify(fieldsToUpdate[f].raw)], _.keys(fieldsToUpdate)))
+  const fieldsWithoutRawData = _.mapValues(f => _.omit(['raw'], f), fieldsToUpdate)
+
+  return _.merge(fieldsWithoutRawData, rawFields)
+}
+
```
RafaelTaranto: isn't this removing any other fields and just updating id_card_data?

RafaelTaranto (changes requested):

lib/customers.js
chaotixkilla: You are correct! This next commit should solve this, nice catch

lib/customers.js
```diff
@@ -281,6 +281,20 @@ function enhanceAtFields (fields) {
...
+function updateRawData (fields) {
+  const fieldsWithRawData = [
+    'id_card_data'
+  ]
+
+  const fieldsToUpdate = _.intersection(fieldsWithRawData, _.keys(fields))
+  const rawFields = _.fromPairs(_.map(f => [`${f}_raw`, JSON.stringify(fields[f].raw)], fieldsToUpdate))
+  const fieldsWithoutRawData = _.mapValues(f => {
+    return _.has(f, 'raw') ? _.omit(['raw'], f) : f
+  }, fields)
```
RafaelTaranto: this still has the potential of breaking fields not on fieldsWithRawData. What about something like:

```js
const fieldsWithRawData = [
  'id_card_data'
]

const mapPairToRaw = _.map(([key, value]) => ([
  [key, _.omit(['raw'], value)],
  [`${key}_raw`, JSON.stringify(value.raw)]
]))

function updateRawData (fields) {
  const pickTransformable = _.pick(fieldsWithRawData)
  const extractRaws = _.compose(_.fromPairs, _.flatten, mapPairToRaw, _.toPairs)
  const assignOriginal = _.assign(fields)

  return _compose(assignOriginal, extractRaws, pickTransformable)(fields)
}
```

RafaelTaranto (changes requested):

RafaelTaranto (approved):

PR #567, opened 2021-01-15
chore: remove old markets, add new coins for itBit

CrypticaScriptura:


RafaelTaranto:
Will be  handled by this one (already partially implemented there): https://github.com/lamassu/lamassu-server/pull/568

PR #568, opened 2021-01-18
Migrate current plugins to ccxt

josepfo:


josepfo:
[Integrate CCXT for multiple exchanges and tickers](https://trello.com/c/vw48Scyb/459-integrate-ccxt-for-multiple-exchanges-and-tickers)

lib/plugins/common/cctx.js
```diff
@@ -0,0 +1,28 @@
+const _ = require('lodash/fp')
```
hellypi: is the file supposed to be called cctx instead of ccxt?

lib/plugins/exchange/cctx/cctx.js
```diff
@@ -0,0 +1,57 @@
...
+  try {
+    const exchange = setUpExchange(account, exchangeName)
+    const symbol = common.verifyCurrencies(exchangeName, fiatCode, cryptoCode)
+    const amount = coinUtils.toUnit(cryptoAtoms, cryptoCode).toFixed(8)
+    if (exchangeName === 'itbit') {
+      return exchange.fetchOrderBook(symbol)
+        .then(orderBook => {
+          return exchange.createOrder(symbol, 'limit', side, amount, calculatePrice(side, amount, orderBook), { walletId: account.walletId })
+        })
+    }
```
hellypi: I think we can remove the { } wrapped around this return statement. From what I gather, our styling convention dictates that if we have only a return statement inside an if, we don't use curly braces

lib/plugins/exchange/cctx/cctx.js
```diff
@@ -0,0 +1,57 @@
...
+    return exchange.createOrder(symbol, 'market', side, amount)
+  } catch (e) {
+    return Promise.reject(e)
+  }
+}
+
+function setUpExchange (account, exchangeName) {
+  // map given credentials to cctx properties
+  if (!_.includes(exchangeName, ccxt.exchanges)) {
```
hellypi: I think you can use _.negate instead of ! but I'm not sure if it's worth the extra verbosity... confirm with Rafael
also, remove braces

lib/plugins/exchange/cctx/cctx.js
```diff
@@ -0,0 +1,57 @@
...
+    throw new Error(`Exchange ${exchangeName} not supported by ccxt.`)
+  }
+
+  switch (exchangeName) {
+    case 'itbit':
+      if (!account.clientKey || !account.clientSecret || !account.userId || !account.walletId)
+        throw new Error('Must provide user ID, wallet ID, client key, and client secret')
+      return new ccxt[exchangeName](_.mapKeys((key) => { return key === 'clientKey' ? 'apiKey' : key === 'clientSecret' ? 'secret' : key === 'userId' ? 'uid' : key }, _.omit(['walletId'], account)))
```
hellypi: If we move the ternary operators to a const it becomes more readable
`
const someName = key === 'clientKey' ? 'apiKey' : key === 'clientSecret' ? 'secret' : key === 'userId' ? 'uid' : key
`
`
_.mapKeys((key) => someName
`

same for the others

lib/plugins/ticker/cctx/cctx.js
```diff
@@ -0,0 +1,57 @@
...
+const axios = require('axios')
+const _ = require('lodash/fp')
+const common = require('../../common/cctx')
+
+function ticker (exchangeName, fiatCode, cryptoCode) {
+  const exchange = new ccxt[exchangeName]()
+
+  if (fiatCode === 'EUR' || fiatCode === 'USD' || exchange.id === 'coinbase') {
+    return getCurrencyRates(exchange, fiatCode, cryptoCode)
+  }
```
hellypi: remove braces

hellypi (changes requested):

lib/plugins/common/cctx.js
josepfo: No, it's not, good catch!

lib/plugins/exchange/cctx/cctx.js
RafaelTaranto: Hey, this is only for ifs that can be put in a single line. If it goes multiline {} is required (even by standard)

lib/plugins/exchange/cctx/cctx.js
RafaelTaranto: Yeah, `!` is fine here.

lib/plugins/exchange/cctx/cctx.js
RafaelTaranto: This one should have brackets IMO. Also there's too much code a single line, try to stick to 80 col length.

lib/plugins/ticker/cctx/cctx.js
RafaelTaranto: That would force it into multi-line.

lib/plugins/common/ccxt.js
```diff
@@ -0,0 +1,28 @@
...
+  kraken: ['USD', 'EUR']
+}
+
+module.exports = { verifyCurrencies }
+
+function verifyCurrencies (exchangeName, fiatCode, cryptoCode) {
+  if (!_.includes(cryptoCode, CRYPTO[exchangeName])) {
+    throw new Error('Unsupported crypto: ' + cryptoCode)
+  }
+  if (!(exchangeName === 'coinbase')) { // coinbase is only used for ticker and it's expected to support most of the fiat
```
hellypi: why not exchangeName !== 'coinbase'?
also move the comment to its own line above the if

lib/plugins/exchange/ccxt/ccxt.js
```diff
@@ -0,0 +1,73 @@
...
+        })
+    }
+    return exchange.createOrder(symbol, 'market', side, amount)
+  } catch (e) {
+    return Promise.reject(e)
+  }
+}
+
+function setUpExchange (account, exchangeName) {
+  // map given credentials to cctx properties
```
hellypi: typo

hellypi (changes requested):

hellypi (approved):

lib/exchange.js
```diff
@@ -10,23 +10,22 @@ function lookupExchange (settings, cryptoCode) {
...
   return fetchExchange(settings, cryptoCode)
-    .then(r => r.exchange.buy(r.account, cryptoAtoms, fiatCode, cryptoCode))
+    .then(r => ccxt.trade('buy', r.account, cryptoAtoms, fiatCode, cryptoCode, r.exchangeName))
 }
 
 function sell (settings, cryptoAtoms, fiatCode, cryptoCode) {
   return fetchExchange(settings, cryptoCode)
-    .then(r => r.exchange.sell(r.account, cryptoAtoms, fiatCode, cryptoCode))
+    .then(r => ccxt.trade('sell', r.account, cryptoAtoms, fiatCode, cryptoCode, r.exchangeName))
 }
```
RafaelTaranto: Hmm, with this we lost our mock-exchange. I think the best way to implement that right now is with a simple if.

lib/ticker.js
```diff
@@ -11,14 +10,10 @@ function _getRates (settings, fiatCode, cryptoCode) {
...
-      const plugin = configManager.getWalletSettings(cryptoCode, config).ticker
-
-      const account = settings.accounts[plugin]
-      const ticker = ph.load(ph.TICKER, plugin)
-
+      const exchangeName = configManager.getWalletSettings(cryptoCode, config).ticker
       const market = [cryptoCode, fiatCode].join('-')
 
-      return ticker.ticker(account, fiatCode, cryptoCode)
+      return ccxt.ticker(exchangeName, fiatCode, cryptoCode)
```
RafaelTaranto: Same thing here, we lost the mock-ticker.

lib/plugins/common/ccxt.js
```diff
@@ -0,0 +1,28 @@
...
+  itbit: ['BTC', 'ETH', 'LTC', 'BCH'],
+  kraken: ['BTC', 'ETH', 'LTC', 'DASH', 'ZEC', 'BCH'],
+  coinbase: ['BTC', 'ETH', 'LTC', 'BCH', 'ZEC', 'DASH']
+}
+
+const FIAT = {
+  bitstamp: ['USD', 'EUR'],
+  itbit: ['USD'],
+  kraken: ['USD', 'EUR']
+}
```
RafaelTaranto: I figure this should be the other way around, every exchange exports the support FIAT and CRYPTO and common/ccxt.js unifies them.

lib/plugins/common/ccxt.js
```diff
@@ -0,0 +1,28 @@
...
+
+function verifyCurrencies (exchangeName, fiatCode, cryptoCode) {
```
RafaelTaranto: Better to keep this called buildMarket. If we need something else for tickers create a specific function for that.

lib/plugins/common/ccxt.js
```diff
@@ -0,0 +1,28 @@
...
+  if (!_.includes(cryptoCode, CRYPTO[exchangeName])) {
+    throw new Error('Unsupported crypto: ' + cryptoCode)
+  }
+  if (exchangeName !== 'coinbase') {
```
RafaelTaranto: I don't get this if. Why does coinbase behaves differently?

lib/plugins/exchange/bitstamp.js
```diff
@@ -0,0 +1,21 @@
+const common = require('../common/ccxt')
+const _ = require('lodash/fp')
+const consts = require('./consts')
+const ORDER_TYPE = consts.ORDER_TYPES.MARKET
+const FIAT = common.FIAT['kraken']
+const CRYPTO = common.CRYPTO['kraken']
```
RafaelTaranto: minor nitpick in import orders, i like to group them in 3 groups: node modules, "external" lib, local lib. Make sure to separate them with a empty line, also add a empty line between imports and normal code. 

do this for all files on the PR.

```js
const nodeModulesLib = require('node_modules_lib')
const nodeModulesLib2 = require('node_modules_lib2')

const externalLib = require('../externalLib')
const externalLib2 = require('../externalLib2')

const localLib = require('./localLib')
const localLib2 = require('./localLib2')

const normalCode = 1 + 2
```

lib/ticker.js
```diff
@@ -11,14 +10,10 @@ function _getRates (settings, fiatCode, cryptoCode) {
   return Promise.resolve()
     .then(() => {
       const config = settings.config
-      const plugin = configManager.getWalletSettings(cryptoCode, config).ticker
-
-      const account = settings.accounts[plugin]
-      const ticker = ph.load(ph.TICKER, plugin)
-
+      const exchangeName = configManager.getWalletSettings(cryptoCode, config).ticker
```
RafaelTaranto: this should be tickerName

lib/plugins/exchange/bitstamp.js
```diff
@@ -0,0 +1,21 @@
+const common = require('../common/ccxt')
+const _ = require('lodash/fp')
+const consts = require('./consts')
+const ORDER_TYPE = consts.ORDER_TYPES.MARKET
```
RafaelTaranto: This can be changed into:

```js
const { ORDER_TYPES } = require('./consts')

const ORDER_TYPE = ORDER_TYPES.MARKET
```

This applies for all places where `consts` is being imported, we don't really get a lot of value from having a full `consts` object

lib/plugins/exchange/bitstamp.js
```diff
@@ -0,0 +1,21 @@
...
+    'key': 'apiKey',
+    'clientId': 'uid'
+  }
+  const mapped = _.mapKeys(key => mapper[key] ? mapper[key] : key)(account)
+  return { ...mapped, timeout: 3000 }
+}
+
+const isConfigValid = ({ key, clientId, secret }) => key && secret && clientId
+
+const amountPrecision = () => 8
```
RafaelTaranto: I don't see why this would need to be a function, might as well export a constant.

lib/plugins/exchange/ccxt.js
```diff
@@ -0,0 +1,51 @@
...
+const kraken = require('./kraken')
+const bitstamp = require('./bitstamp')
+const itbit = require('./itbit')
+const consts = require('./consts')
+
+const ALL_EXCHANGES = {
+  kraken,
+  bitstamp,
+  itbit
+}
```
RafaelTaranto: Might as well export this from `common/ccxt.js`

lib/plugins/exchange/ccxt.js
```diff
@@ -0,0 +1,51 @@
+var ccxt = require('ccxt')
+const coinUtils = require('../../coin-utils')
+const common = require('../common/ccxt')
```
RafaelTaranto: use desconstruct for common too, just like consts we don't really have reason to hold the entire object

```js
const { ALL_EXCHANGES, buildMarket } = require('../common/ccxt')
```

lib/plugins/exchange/ccxt.js
```diff
@@ -0,0 +1,51 @@
...
+  kraken,
+  bitstamp,
+  itbit
+}
+
+function trade (side, account, cryptoAtoms, fiatCode, cryptoCode, exchangeName) {
+  try {
+    const exchangeConfig = ALL_EXCHANGES[exchangeName]
+
+    if (!exchangeConfig) throw Error('no exchange')
```
RafaelTaranto:
```suggestion
    if (!exchangeConfig) throw Error('Exchange configuration not found')

```

lib/plugins/exchange/ccxt.js
```diff
@@ -0,0 +1,51 @@
...
+    if (exchangeConfig.isConfigValid && !exchangeConfig.isConfigValid(account)) throw Error('Invalid config')
```
RafaelTaranto: Maybe it would be good to do a desconstruct of exchangeConfig as well. Makes the code more readable.

```js
const { isConfigValid, loadConfig, loadOption, AMOUNT_PRECISION, ORDER_TYPE} = exchangeConfig
```

RafaelTaranto: This line can be

```js
     if (_.isFunction(isConfigValid) && !isConfigValid(account)) throw Error('Invalid config')
```

lib/plugins/exchange/ccxt.js
```diff
@@ -0,0 +1,51 @@
...
+
+    const exchange = new ccxt[exchangeName](exchangeConfig.loadConfig(account))
```
RafaelTaranto: we can probably make loadConfig default to noop, it's a identity function so it will just return `account`

```js
const { isConfigValid, loadConfig = _.noop, loadOption, AMOUNT_PRECISION, ORDER_TYPE} = exchangeConfig
```

RafaelTaranto: Also it looks like you're only using `exchange` on line 29 and 31, might as well move this line closer to that and add a empty line before it.

lib/plugins/exchange/ccxt.js
```diff
@@ -0,0 +1,51 @@
...
+
+    const symbol = common.verifyCurrencies(exchangeName, fiatCode, cryptoCode)
+    const precision = exchangeConfig.amountPrecision ? exchangeConfig.amountPrecision() : consts.DECIMAL_PRECISION.DEFAULT_AMOUNT
```
RafaelTaranto:
```suggestion
    const precision = _.defaultTo(DECIMAL_PRECISION.DEFAULT_AMOUNT, AMOUNT_PRECISION)

```

lib/plugins/exchange/ccxt.js
```diff
@@ -0,0 +1,51 @@
...
+    const amount = coinUtils.toUnit(cryptoAtoms, cryptoCode).toFixed(precision)
+    const options = exchangeConfig.loadOptions ? exchangeConfig.loadOptions(account) : {}
```
RafaelTaranto: use _.isFunction here as well

lib/plugins/exchange/consts.js
```diff
@@ -0,0 +1,11 @@
+
+const ORDER_TYPES = {
+  MARKET: 'market',
+  LIMIT: 'limit'
+}
+const DECIMAL_PRECISION = {
+  PRICE: 2,
+  DEFAULT_AMOUNT: 8
+}
```
RafaelTaranto: This looks to be used only in `exchange/ccxt.js. You just add these consts there

```
const DEFAULT_PRICE_PRECISION = 2
const DEFAULT_AMOUNT_PRECISION = 8
```

RafaelTaranto (changes requested):

lib/plugins/common/ccxt.js
```diff
@@ -0,0 +1,28 @@
...
+  bitstamp: ['BTC', 'ETH', 'LTC', 'BCH'],
+  itbit: ['BTC', 'ETH', 'LTC', 'BCH'],
+  kraken: ['BTC', 'ETH', 'LTC', 'DASH', 'ZEC', 'BCH'],
+  coinbase: ['BTC', 'ETH', 'LTC', 'BCH', 'ZEC', 'DASH']
+}
+
+const FIAT = {
+  bitstamp: ['USD', 'EUR'],
+  itbit: ['USD'],
+  kraken: ['USD', 'EUR']
```
CrypticaScriptura: Kraken has CAD and GBP markets that [some prefer](https://support.lamassu.is/hc/en-us/articles/360039433291-Setting-Kraken-to-trade-on-other-currency-markets).

But moreover, we should have an admin toggle, if not already planned, whereby one can choose which market to trade on, rather than automatically assume based on machine currency.

lib/plugins/exchange/itbit.js
```diff
@@ -0,0 +1,24 @@
+const _ = require('lodash/fp')
+
+const { ORDER_TYPES } = require('./consts')
+const { COINS } = require('../../new-admin/config/coins') // para cada exchange
```
RafaelTaranto: remove this comment here.

lib/plugins/ticker/ccxt.js
```diff
@@ -0,0 +1,57 @@
...
+const ccxt = require('ccxt')
+
+const BN = require('../../bn')
+const { buildMarket } = require('../common/ccxt')
+
+function ticker (fiatCode, cryptoCode, tickerName) {
+  const ticker = new ccxt[tickerName]({ timeout: 3000 })
+  if (fiatCode === 'EUR' || fiatCode === 'USD' || ticker.id === 'coinbase') {
+    return getCurrencyRates(ticker, fiatCode, cryptoCode)
+  }
```
RafaelTaranto: Shouldn't this be generic with the buildMarket thing?

RafaelTaranto (changes requested):
Still missing the update on accounts.js here.

lib/plugins/ticker/ccxt.js
josepfo: This is used only to skip Bitpay forex rates under those circumstances.

lib/plugins/ticker/ccxt.js
```diff
@@ -0,0 +1,56 @@
...
+    if (ticker.has['fetchTicker']) {
+      const symbol = buildMarket(fiatCode, cryptoCode, ticker.id)
+      return ticker.fetchTicker(symbol)
+        .then(res => ({
+          rates: {
+            ask: BN(res.ask),
+            bid: BN(res.bid)
+          }
+        }))
+    }
```
RafaelTaranto: What happens if ticker doesn't have 'fetchTicker'? Shouldn't we throw a error?

lib/plugins/ticker/ccxt.js
```diff
@@ -0,0 +1,56 @@
...
+
+const BN = require('../../bn')
+const { buildMarket, verifyFiatSupport } = require('../common/ccxt')
+
+function ticker (fiatCode, cryptoCode, tickerName) {
+  const ticker = new ccxt[tickerName]({ timeout: 3000 })
+  if (verifyFiatSupport(fiatCode, tickerName)) {
+    return getCurrencyRates(ticker, fiatCode, cryptoCode)
+  }
+  return axios.get('https://bitpay.com/rates')
```
RafaelTaranto, line 14: Add a empty line before this return, easier to see that we can return on line 11 as well.

lib/new-admin/config/accounts.js
```diff
@@ -13,11 +15,11 @@ const EMAIL = 'email'
...
-  { code: 'bitpay', display: 'Bitpay', class: TICKER, cryptos: [BTC, BCH] },
-  { code: 'kraken', display: 'Kraken', class: TICKER, cryptos: [BTC, ETH, LTC, DASH, ZEC, BCH] },
-  { code: 'bitstamp', display: 'Bitstamp', class: TICKER, cryptos: [BTC, ETH, LTC, BCH] },
-  { code: 'coinbase', display: 'Coinbase', class: TICKER, cryptos: [BTC, ETH, LTC, DASH, ZEC, BCH] },
-  { code: 'itbit', display: 'itBit', class: TICKER, cryptos: [BTC, ETH] },
+  { code: 'bitpay', display: 'Bitpay', class: TICKER, cryptos: ALL['bitpay'].CRYPTO },
+  { code: 'kraken', display: 'Kraken', class: TICKER, cryptos: ALL['kraken'].CRYPTO },
+  { code: 'bitstamp', display: 'Bitstamp', class: TICKER, cryptos: ALL['bitstamp'].CRYPTO },
+  { code: 'coinbase', display: 'Coinbase', class: TICKER, cryptos: ALL['coinbase'].CRYPTO },
+  { code: 'itbit', display: 'itBit', class: TICKER, cryptos: ALL['itbit'].CRYPTO },
```
RafaelTaranto: No reason to have this `ALL['bitpay']` notation, you already know the names. `ALL.bitpay.CRYPTO` would work just fine. Maybe you want to destruct `ALL` just like you do for `COINS`

lib/plugins/exchange/ccxt.js
```diff
@@ -0,0 +1,48 @@
...
+    const symbol = buildMarket(fiatCode, cryptoCode, exchangeName)
+    const precision = _.defaultTo(DEFAULT_AMOUNT_PRECISION, AMOUNT_PRECISION)
+    const amount = toUnit(cryptoAtoms, cryptoCode).toFixed(precision)
+    const options = _.isFunction(loadOptions) ? loadOptions(account) : {}
+    const exchange = new ccxt[exchangeName](loadConfig(account))
+
+    if (ORDER_TYPE === ORDER_TYPES.MARKET) {
+      return exchange.createOrder(symbol, ORDER_TYPES.MARKET, side, amount, null, options)
+    }
+    return exchange.fetchOrderBook(symbol)
```
RafaelTaranto, line 29: Add empty line above this return as well

lib/plugins/exchange/bitstamp.js
```diff
@@ -0,0 +1,23 @@
...
+const loadConfig = (account) => {
+  const mapper = {
+    'key': 'apiKey',
+    'clientId': 'uid'
+  }
+  const mapped = _.mapKeys(key => mapper[key] ? mapper[key] : key)(account)
+  return { ...mapped, timeout: 3000 }
+}
+
+const isConfigValid = ({ key, clientId, secret }) => key && secret && clientId
```
RafaelTaranto: Better to use `_.every` and `_.notEmpty` on isConfigValid for every exchange.

RafaelTaranto (changes requested):

lib/plugins/exchange/bitstamp.js
```diff
@@ -18,6 +18,9 @@ const loadConfig = (account) => {
   return { ...mapped, timeout: 3000 }
 }
 
-const isConfigValid = ({ key, clientId, secret }) => key && secret && clientId
+const isConfigValid = options => {
+  const requiredOptions = _.pick(['key', 'secret', 'clientId'], options)
+  return _.isEqual(options, requiredOptions)
```
RafaelTaranto: I'm not following this logic here, just testing if the original options is equal to the picked option but they all can still be null.

RafaelTaranto (changes requested):

lib/coin-utils.js
```diff
@@ -103,3 +103,8 @@ function toUnit (cryptoAtoms, cryptoCode) {
   const unitScale = cryptoRec.unitScale
   return cryptoAtoms.shift(-unitScale)
 }
+
+function isConfigValid (config, fields) {
+  const values = _.map(it => _.get(it)(config))(fields)
+  return _.every(it => it || it === 0)(values)
+}
```
RafaelTaranto: This isn't really a coin utils thing. Export it from the ccxt commons

RafaelTaranto (changes requested):

RafaelTaranto (approved):

PR #571, opened 2021-01-22
Admin server refactoring

chaotixkilla:


chaotixkilla:
[Refactor l-a-s entrypoint express config](https://trello.com/c/k6mGlh91/522-refactor-l-a-s-entrypoint-express-config)

lib/new-admin/admin-server.js
```diff
@@ -66,35 +65,7 @@ app.use(cors({ credentials: true, origin: devMode && 'https://localhost:3001' })
...
-        expires: NEVER
-      }
-
-      const token = r.token
-      req.token = token
-      res.cookie('token', token, cookieOpts)
-      res.sendStatus(200)
-    })
-})
+app.use('/', register)
```
hellypi: I think this will bring us problems in the future like it is right now because if we want to add a different route like /newapi/register we will have to add it inside the './routes/authentication' file and soon we will run into collision problems. 
In my opinion this should become
`app.use('/api', register)`
and inside the './routes/authentication' file it becomes
`router.get('/register', register)`

hellypi (changes requested):

lib/new-admin/admin-server.js
chaotixkilla: That's a good point! Changed.

hellypi (approved):

RafaelTaranto (approved):

PR #573, opened 2021-01-25
Confirm delete popup

josepfo:


josepfo:
[Add a confirmation popup before deleting any rows on the admin](https://trello.com/c/wf9LbaK4/517-add-a-confirmation-popup-before-deleting-any-rows-on-the-admin)

new-lamassu-admin/src/components/DeleteDialog.js
```diff
@@ -0,0 +1,76 @@
...
+    paddingRight: spacer * 1.5,
+    display: 'flex',
+    'justify-content': 'space-between',
+    '& > h4': {
+      margin: 0
+    },
+    '& > button': {
+      padding: 0,
+      marginTop: -(spacer / 2)
+    }
```
hellypi: Make own classes instead of using selectors, unless absolutely necessary

new-lamassu-admin/src/components/DeleteDialog.js
```diff
@@ -0,0 +1,76 @@
...
+export const DialogTitle = ({ children, close }) => {
+  const classes = useStyles()
+  return (
+    <div className={classes.dialogTitle}>
+      {children}
+      {
+        <IconButton size={16} aria-label="close" onClick={close}>
+          <CloseIcon />
+        </IconButton>
+      }
```
hellypi: do we need these brackets here?

new-lamassu-admin/src/components/editableTable/Row.js
```diff
@@ -74,9 +82,20 @@ const ActionCol = ({ disabled, editing }) => {
...
+            }}>
             {disabled ? <DisabledDeleteIcon /> : <DeleteIcon />}
           </IconButton>
+          {
+            <DeleteDialog
+              open={deleteDialog}
+              setDeleteDialog={setDeleteDialog}
+              onConfirmed={onConfirmed}
+            />
+          }
```
hellypi: do we need these brackets here?

hellypi (changes requested):

hellypi (approved):

new-lamassu-admin/src/components/editableTable/Row.js
```diff
@@ -44,6 +45,13 @@ const ActionCol = ({ disabled, editing }) => {
     resetForm()
   }
 
+  const [deleteDialog, setDeleteDialog] = useState(false)
+
+  const onConfirmed = () => {
+    onDelete(values.id)
+    setDeleteDialog(false)
```
RafaelTaranto: What happens if the delete fails? You can simulate this by shutting down the server prior to confirming. 

The correct thing to do would be to only close the popup if the delete is successful, else show a error message on it. 

This might get pretty tricky on the editable table, so try on the other examples first.

RafaelTaranto (changes requested):

new-lamassu-admin/src/components/DeleteDialog.js
```diff
@@ -0,0 +1,97 @@
+import {
+  Dialog,
+  DialogActions,
+  DialogContent,
+  makeStyles
+} from '@material-ui/core'
+// import * as R from 'ramda'
```
RafaelTaranto: remove comment

RafaelTaranto (changes requested):

new-lamassu-admin/src/components/editableTable/Row.js
```diff
@@ -1,9 +1,11 @@
+/* eslint-disable import/order */
```
RafaelTaranto: Any reason for disabling this?

RafaelTaranto (changes requested):

RafaelTaranto (approved):

PR #582, opened 2021-02-05
Implement ERC-20 support

chaotixkilla:


chaotixkilla:
[ERC-20: Add USDT](https://trello.com/c/snrPYh1G/549-erc-20-add-usdt)

lib/plugins/wallet/geth/base.js
```diff
@@ -15,6 +17,9 @@ const paymentPrefixPath = "m/44'/60'/0'/0'"
 const defaultPrefixPath = "m/44'/60'/1'/0'"
 let lastUsedNonces = {}
 
+// const contract = web3.eth.contract(erc20ABIs.usdt, coins.utils.getErc20Token('USDT').contractAddress)
+// console.log(contract)
```
josepfo: remove comments

lib/plugins/wallet/geth/base.js
```diff
@@ -62,25 +67,86 @@ function sendCoins (account, toAddress, cryptoAtoms, cryptoCode) {
 }
 
 function checkCryptoCode (cryptoCode) {
-  if (cryptoCode === 'ETH') return Promise.resolve()
+  if (cryptoCode === 'ETH' || coins.utils.isErc20Token(cryptoCode)) {
+    // console.log(cryptoCode)
```
josepfo: same as above

josepfo (changes requested):

lib/admin/funding.js
```diff
@@ -4,7 +4,7 @@ const settingsLoader = require('./settings-loader')
 const configManager = require('./config-manager')
 const wallet = require('../wallet')
 const ticker = require('../ticker')
-const coinUtils = require('../coin-utils')
+const { utils } = require('lamassu-coins')
```
RafaelTaranto:
```suggestion
const { utils as coinUtils } = require('lamassu-coins')
```

lib/plugins/common/kraken.js
```diff
@@ -0,0 +1,32 @@
+const PAIRS = {
```
RafaelTaranto: I'm pretty sure this file is not needed anymore

lib/new-admin/config/accounts.js
```diff
@@ -24,7 +24,7 @@ const ALL_ACCOUNTS = [
   { code: 'mock-ticker', display: 'Mock (Caution!)', class: TICKER, cryptos: ALL_CRYPTOS, dev: true },
   { code: 'bitcoind', display: 'bitcoind', class: WALLET, cryptos: [BTC] },
   { code: 'no-layer2', display: 'No Layer 2', class: LAYER_2, cryptos: ALL_CRYPTOS },
-  { code: 'infura', display: 'Infura', class: WALLET, cryptos: [ETH] },
+  { code: 'infura', display: 'Infura', class: WALLET, cryptos: [ETH, USDT] },
   { code: 'geth', display: 'geth (DEPRECATED)', class: WALLET, cryptos: [ETH], deprecated: true },
```
RafaelTaranto: Geth will be un-deprecated after a merge, you can add USDT here as well.

lib/plugins/erc20/abi/usdt.json
```diff
@@ -0,0 +1 @@
...":"address"}],"name":"Deprecate","type":"event"},{"anonymous":false,"inputs":[{"indexed":false,"name":"feeBasisPoints","type":"uint256"},{"indexed":false,"name":"maxFee","type":"uint256"}],"name":"Params","type":"event"},{"anonymous":false,"inputs":[{"indexed":false,"name":"_blackListedUser","type":"address"},{"indexed":false,"name":"_balance","type":"uint256"}],"name":"DestroyedBlackFunds","type":"event"},{"anonymous":false,"inputs":[{"indexed":false,"name":"_user","type":"address"}],"name":"AddedBlackList","type":"event"},{"anonymous":false,"inputs":[{"indexed":false,"name":"_user","type":"address"}],"name":"RemovedBlackList","type":"event"},{"anonymous":false,"inputs":[{"indexed":true,"name":"owner","type":"address"},{"indexed":true,"name":"spender","type":"address"},{"indexed":false,"name":"value","type":"uint256"}],"name":"Approval","type":"event"},{"anonymous":false,"inputs":[{"indexed":true,"name":"from","type":"address"},{"indexed":true,"name":"to","type":"address"},{"indexed":false,"name":"value","type":"uint256"}],"name":"Transfer","type":"event"},{"anonymous":false,"inputs":[],"name":"Pause","type":"event"},{"anonymous":false,"inputs":[],"name":"Unpause","type":"event"}]
```
RafaelTaranto: isn't the ABI a definition of contract? If so I don't think there's anything specific to this ABI that relates to USDT, it should be a generic erc20 thing.

lib/plugins/exchange/bitstamp/bitstamp.js
```diff
@@ -0,0 +1,45 @@
+const common = require('../../common/bitstamp')
+const { utils } = require('lamassu-coins')
+
+function buy (account, cryptoAtoms, fiatCode, cryptoCode) {
+  return trade('buy', account, cryptoAtoms, fiatCode, cryptoCode)
+}
+
```
RafaelTaranto: another file that can be removed

lib/plugins/exchange/itbit/itbit.js
```diff
@@ -0,0 +1,45 @@
+const common = require('../../common/itbit')
+const { utils } = require('lamassu-coins')
+
+exports.buy = function (account, cryptoAtoms, fiatCode, cryptoCode) {
```
RafaelTaranto: another file to remove

lib/plugins/exchange/kraken/kraken.js
```diff
@@ -0,0 +1,44 @@
+// Note: Using DeX3/npm-kraken-api to adjust timeout time
+const Kraken = require('kraken-api')
+const _ = require('lodash/fp')
+
```
RafaelTaranto: another one

RafaelTaranto (changes requested):

lib/plugins/erc20/abi/usdt.json
chaotixkilla: I believe you're right. My reasoning for this approach is that each token can select which functions from the erc-20 ABI to use, which leads to, for example USDT's ABI being different than, for example, BNB's ABI. I can find the most generic ERC-20 ABI and simply use that here

lib/commission-math.js
```diff
@@ -1,12 +1,12 @@
...
 const configManager = require('./new-config-manager')
-const coinUtils = require('./coin-utils')
+const { utils: coinUtils } = require('lamassu-coins')
 
 function truncateCrypto (cryptoAtoms, cryptoCode) {
   const DECIMAL_PLACES = 3
   if (cryptoAtoms.eq(0)) return cryptoAtoms
 
-  const scale = coinUtils.getCryptoCurrency(cryptoCode).displayScale
+  const scale = utils.getCryptoCurrency(cryptoCode).displayScale
```
RafaelTaranto: Forgot this one.

lib/plugins/wallet/geth/base.js
```diff
@@ -63,28 +65,88 @@ function sendCoins (account, tx, settings, operatorId) {
...
+        data: contract.transfer.getData(toAddress, hex(toSend))
+      }
+
+      const tx = new Tx(rawTx)
+      const privateKey = wallet.getPrivateKey()
+
+      tx.sign(privateKey)
+
+      return '0x' + tx.serialize().toString('hex')
+    })
```
RafaelTaranto: Most of this is duplicated code, can't we split off in a couple of functions?

RafaelTaranto (changes requested):

lib/plugins/wallet/geth/base.js
chaotixkilla: I think splitting off in multiple functions could get confusing fast, so I think I came with a more elegant solution. Will need some changes if we decide on more token standards, but for now I think it works well

lib/plugins/erc20/abi/usdt.json
RafaelTaranto: Oh, do you have example of the differences. AFAIK there's only a single ABI for erc-20 and if you get the one from the spec should work for every erc-20 compliant token.

RafaelTaranto (approved):

lib/blockchain-utils.js
```diff
@@ -0,0 +1,7 @@
+const options = require('./options')
+
+function blockchainDir () {
+  return options.blockchainDir
+}
+
+module.exports = { blockchainDir }
```
joshmh: Do we need a whole module just for this?

joshmh (changes requested):

PR #585, opened 2021-02-08
Obfuscate password fields sending to react

josepfo:


josepfo:
[Stop sending passwords on graphql requests](https://trello.com/c/qr9mPWgg/532-stop-sending-passwords-on-graphql-requests)

chaotixkilla (changes requested):
Since you're replacing the `.loadAccounts()` function with `.showAccounts()`, wouldn't it be more intuitive to enhance the `loadAccounts` function to hide the secret fields by default?

You're creating the `showAccounts()` function, but it is pretty much an enhanced `loadAccounts()`. I'll place this on the 'Fix' row, confirm with Rafael if what I said makes sense and if so, fix it. If not, move this to the next lane

chaotixkilla (approved):

lib/new-settings-loader.js
```diff
@@ -4,14 +4,29 @@ const migration = require('./config-migration')
...
 const accountsSql = `update user_config set data = $2, valid = $3, schema_version = $4 where type = $1;
 insert into user_config (type, data, valid, schema_version)
 select $1, $2, $3, $4 where $1 not in (select type from user_config)`
 function saveAccounts (accountsToSave) {
   return loadAccounts()
     .then(currentAccounts => {
-      const newAccounts = _.assign(currentAccounts, accountsToSave)
+      const serviceCode = _.keys(accountsToSave)[0]
+      accountsToSave[serviceCode] = _.omitBy(_.isEmpty, accountsToSave[serviceCode])
+      const newAccounts = _.merge(currentAccounts, accountsToSave)
```
RafaelTaranto: I believe it is/should be possible to save any amount of accounts.

new-lamassu-admin/src/components/inputs/base/SecretInput.js
```diff
@@ -4,11 +4,8 @@ import { TextInput } from '../base'
 
 const SecretInput = memo(({ value, onFocus, onBlur, ...props }) => {
   const [focused, setFocused] = useState(false)
-
+  const isPasswordFilled = props.isPasswordFilled
```
RafaelTaranto: instead of accessing props.isPasswordFilled you can destruct just like we do for `value, onFocus, onBlur, isPasswordFilled, ...props`

new-lamassu-admin/src/components/inputs/base/TextInput.js
```diff
@@ -26,8 +26,9 @@ const TextInput = memo(
     ...props
   }) => {
     const classes = useStyles({ textAlign, width, size })
-    const filled = !error && !R.isNil(value) && !R.isEmpty(value)
-
+    const isPasswordFilled = props.isPasswordFilled
```
RafaelTaranto: same as above

new-lamassu-admin/src/pages/Services/schemas/twilio.js
```diff
@@ -44,5 +47,22 @@ export default {
...
+        .test(secretTest(account?.authToken)),
+      fromNumber: Yup.string()
+        .max(100, 'Too long')
+        .required(),
+      toNumber: Yup.string()
+        .max(100, 'Too long')
+        .required()
+    }
+    return Yup.object().shape(schema)
+  }
```
RafaelTaranto: Why did we keep the validationSchema? Better to replace everywhere to use the function alternative, this goes for every schema

new-lamassu-admin/src/pages/Services/schemas/mailgun.js
```diff
@@ -5,6 +5,7 @@ import TextInputFormik from 'src/components/inputs/formik/TextInput'
 export default {
   code: 'mailgun',
   name: 'Mailgun',
+  hasSecret: false,
```
RafaelTaranto: Wasn't  this deprecated?

new-lamassu-admin/src/pages/Services/schemas/helper.js
```diff
@@ -0,0 +1,12 @@
+import * as R from 'ramda'
+
+const secretTest = secret => ({
+  test(val) {
+    if (R.isNil(secret) && R.isNil(val)) {
```
RafaelTaranto, line 5: shouldn't we test for isNil and isEmpty. I believe that if we fill the field and then empty it it will be a empty string instead of a nil value

new-lamassu-admin/src/pages/Services/Services.js
```diff
@@ -56,6 +57,47 @@ const Services = () => {
...
+      })
+    )(elements)
+  }
+
+  const getAccounts = ({ elements, code }) => {
+    const account = accounts[code]
+    const passwordFields = R.compose(
+      R.reject(R.isNil),
+      R.map(({ component, code }) => (component === SecretInput ? code : null))
+    )(elements)
```
RafaelTaranto: Shouldn't this be a filter?

new-lamassu-admin/src/pages/Services/Services.js
```diff
@@ -56,6 +57,47 @@ const Services = () => {
...
+    return R.compose(
+      R.mapObjIndexed((value, key, obj) =>
+        R.includes(key, passwordFields) ? '' : value
+      )
+    )(account)
```
RafaelTaranto: I don't see a need for this compose here.

new-lamassu-admin/src/pages/Services/Services.js
```diff
@@ -56,6 +57,47 @@ const Services = () => {
     }))(faceElements)
   }
 
+  const getElements = ({ code, elements }) => {
+    return R.compose(
```
RafaelTaranto: Compose is not needed here

new-lamassu-admin/src/pages/Services/Services.js
```diff
@@ -56,6 +57,47 @@ const Services = () => {
...
+                isPasswordFilled:
+                  !R.isNil(accounts[code]) &&
+                  !R.isNil(R.path([elem.code], accounts[code]))
+              },
+              elem
+            )
+          : R.identity(elem)
+      })
+    )(elements)
+  }
```
RafaelTaranto: This is overall pretty hard to parse.

You can do

```
if (elem.component !== SecretInput) return elem
return R.assoc('inputProps',
```

instead of the ternary if.

RafaelTaranto: Also looks like R.assoc does a shallow clone, can't we have nested objects here?

RafaelTaranto (changes requested):

lib/new-settings-loader.js
```diff
@@ -4,14 +4,34 @@ const migration = require('./config-migration')
...
 const accountsSql = `update user_config set data = $2, valid = $3, schema_version = $4 where type = $1;
 insert into user_config (type, data, valid, schema_version)
 select $1, $2, $3, $4 where $1 not in (select type from user_config)`
-function saveAccounts (accountsToSave) {
+function saveAccounts (accounts) {
   return loadAccounts()
     .then(currentAccounts => {
-      const newAccounts = _.assign(currentAccounts, accountsToSave)
+      const serviceCodes = _.keys(accounts)
+      const filledAccountInfo = _.map(code => { return { [code]: _.omitBy(_.isEmpty, accounts[code]) } }, serviceCodes)
```
RafaelTaranto: I think we should only apply this if the code is in the SECRET_FIELDS list, or else ops would never be able to delete any other config.

RafaelTaranto: Also when you need to return a object in a lamda function you can do this:

```js
code => ({ [code]: ..........  })
```

RafaelTaranto: Probably easier to just map over the original array isn't it? I don't see the reason to get the keys and then map over transforming into a object.

lib/new-settings-loader.js
```diff
@@ -4,14 +4,34 @@ const migration = require('./config-migration')
...
+      const accountsToSave = _.reduce((result, item) => {
+        var key = Object.keys(item)[0]
+        result[key] = item[key]
+        return result
+      }, {}, filledAccountInfo)
```
RafaelTaranto: This code is pretty hard to understand. What's the goal here?

RafaelTaranto (changes requested):

lib/new-settings-loader.js
josepfo: Essentially, I wanted to map through each nested field of each account and remove what's empty. There's probably a better way of doing it all at once by accessing the nested fields in a single map but it was not being obvious to me.

lib/new-settings-loader.js
josepfo: Yeah, I agree... Transforms a list of objects into a single object.

RafaelTaranto (approved):

lib/new-settings-loader.js
```diff
@@ -40,6 +53,18 @@ function loadAccounts (schemaVersion) {
...
+    .then(accounts => {
+      const filledSecretPaths = _.compact(_.map(path => {
+        if (!_.isEmpty(_.get(path, accounts))) {
+          return path
+        }
+      }, SECRET_FIELDS))
+      return _.compose(_.map(path => _.assoc(path, PASSWORD_FILLED), filledSecretPaths))(accounts)
+    })
+}
+
```
joshmh, line 67: This looks like a good use case for lenses. Ramda has them but for lodash maybe we could use something like:

https://github.com/jamesmcnamara/shades

new-lamassu-admin/src/pages/Services/schemas/bitgo.js
```diff
@@ -80,47 +82,49 @@ export default {
...
+      token: Yup.string()
+        .max(100, 'Too long')
+        .required(),
+      BTCWalletId: Yup.string().max(100, 'Too long'),
+      BTCWalletPassphrase: Yup.string()
+        .max(100, 'Too long')
+        .when('BTCWalletId', {
+          is: isDefined,
+          then: Yup.string().test(secretTest(account?.BTCWalletPassphrase))
+        }),
```
joshmh: Use a pure function to generate these records to reduce boilerplate. Maybe we can even replace the whole ``.when`` part.

joshmh (changes requested):

lib/new-settings-loader.js
josepfo: Hey, Josh, I've been going through the docs, and I don't seem to find a way to make this return much simpler. Do you have any suggestions? It would be nice if shades had a way to replace all the filledSecretPaths without a map, right?

joshmh: Hey José, the idea of lenses is you can define a path down into the structure and then read and set that deep field easily. I didn't look into how exactly it would change the code here, though. I'll merge this for now but you can play around with it afterwards.

PR #588, opened 2021-02-09
Enable cashbox by default

chaotixkilla:


chaotixkilla:
[Cashbox should always be enabled](https://trello.com/c/WhUb2bzK/559-cashbox-should-always-be-enabled)

chaotixkilla:
[Reject expired IDs](https://trello.com/c/TXOi8uy4/565-reject-expired-ids)

josepfo (approved):

new-lamassu-admin/src/components/editableTable/Row.js
```diff
@@ -192,7 +192,7 @@ const ERow = ({ editing, disabled, lastOfGroup }) => {
 
   const classes = useStyles()
 
-  const shouldStripe = stripeWhen && stripeWhen(values) && !editing
+  const shouldStripe = stripeWhen && stripeWhen(values)
```
RafaelTaranto, line 195: Have you make sure that other places that contain stripe aren't being broken by this change?

new-lamassu-admin/src/pages/Maintenance/CashCassettes.js
```diff
@@ -159,6 +159,8 @@ const CashCassettes = () => {
     }
   ]
 
+  const isAnyStripeFalse = () => !!R.find(R.propEq('stripe', false))(elements)
+
```
RafaelTaranto: Why do we need this?

RafaelTaranto (changes requested):

new-lamassu-admin/src/pages/Maintenance/CashCassettes.js
chaotixkilla: We don't actually. Fixed

RafaelTaranto (approved):

PR #590, opened 2021-02-09
Fix ETH pending balance being equal to the confirmed balance on Funding page

chaotixkilla:


chaotixkilla:
[Check why the pending balance is equal to the wallet balance in 'Funding' page](https://trello.com/c/EAu2R10s/555-check-why-the-pending-balance-is-equal-to-the-wallet-balance-in-funding-page)

josepfo (approved):

RafaelTaranto (approved):

lib/new-admin/funding.js
```diff
@@ -51,7 +51,9 @@ function getSingleCoinFunding (settings, fiatCode, cryptoCode) {
       const rate = (rates.ask.add(rates.bid)).div(2)
       const fundingConfirmedBalance = fundingRec.fundingConfirmedBalance
       const fiatConfirmedBalance = computeFiat(rate, cryptoCode, fundingConfirmedBalance)
-      const pending = fundingRec.fundingPendingBalance
+      const pending = cryptoCode === 'ETH'
+        ? fundingRec.fundingPendingBalance - fundingRec.fundingConfirmedBalance
```
joshmh: I think this should be done at the wallet plugin level. This might also be different for different ETH wallets.

joshmh (changes requested):

lib/new-admin/funding.js
chaotixkilla: I also agree. The main reason I went with this solution, was to avoid an extra request on Infura, but thinking about it, our request cache should help us with this

PR #597, opened 2021-02-12
New spec on empty tables

chaotixkilla:


chaotixkilla:
[Implement the new spec for empty tables](https://trello.com/c/3wSwvpod/573-implement-the-new-spec-for-empty-tables)

PR #602, opened 2021-02-15
Notifications UI fixes

chaotixkilla:


chaotixkilla:
[Small ui fixes](https://trello.com/c/5pJC6s6R/572-small-ui-fixes)

josepfo (approved):

new-lamassu-admin/src/pages/Notifications/sections/FiatBalanceAlerts.styles.js
```diff
@@ -21,5 +23,14 @@ export default {
   },
   col2: {
     width: 136
+  },
+  cashbox: {
+    '& p': {
+      marginRight: 4,
+      fontSize: 20
+    }
+  },
```
RafaelTaranto: isn't this something we can do straight on the cashbox component?

RafaelTaranto (changes requested):

new-lamassu-admin/src/pages/Notifications/sections/FiatBalanceAlerts.styles.js
chaotixkilla: This would make the cashbox element on other pages (CashCassettes) have the larger font size by default, which is not intended. Either the notifications page or the cash cassettes page would need to override the styling for it to work as intended on both pages

RafaelTaranto (approved):

PR #604, opened 2021-02-16
Fix ETH wallet sending the pending balance as available balance

chaotixkilla:


chaotixkilla:
[ETH wallet is sending the real pending balance to machines instead of the real balance](https://trello.com/c/jRBtBS4e/587-eth-wallet-is-sending-the-real-pending-balance-to-machines-instead-of-the-real-balance)

josepfo (approved):

RafaelTaranto (approved):

PR #605, opened 2021-02-17
Validate forms only on submit

josepfo:


josepfo:
[Make the validation run only after submit instead of on focusout](https://trello.com/c/GbjiqSdH/580-make-the-validation-run-only-after-submit-instead-of-on-focusout)

chaotixkilla (approved):

RafaelTaranto (approved):

PR #610, opened 2021-02-18
Prevent new admin from getting indexed on google

josepfo:


josepfo:
[Research and add SEO prevention to our admin project](https://trello.com/c/zIgWM7vP/586-research-and-add-seo-prevention-to-our-admin-project)

new-lamassu-admin/public/index.html
```diff
@@ -7,6 +7,7 @@
       name="viewport"
       content="width=device-width, initial-scale=1, shrink-to-fit=no"
     />
+    <meta name="googlebot" content="noindex"/>
```
hellypi: Do we only want to block google, or do we want to block every search engine?
If we want to block every search engine, change "googlebot" to "robots"
(confirm with Rafael or whoever created the card)

hellypi (changes requested):

new-lamassu-admin/public/index.html
josepfo: Card only mentioned google indexing, hence the googlebot. But you are right, it's better to generalize

hellypi (approved):

RafaelTaranto (approved):

PR #611, opened 2021-02-18
Animate operator information title

josepfo:


josepfo:
[Third level routing animation](https://trello.com/c/rHm19lAC/578-third-level-routing-animation)

new-lamassu-admin/src/App.js
```diff
@@ -91,12 +92,26 @@ const Main = () => {
...
+  }
+
   return (
     <div className={classes.root}>
       {!is404 && wizardTested && <Header tree={tree} />}
       <main className={classes.wrapper}>
         {sidebar && !is404 && wizardTested && (
-          <TitleSection title={parent.title}></TitleSection>
+          <Slide
+            {...transitionProps}
```
hellypi: why not just 
```suggestion
            direction="left"
```

Also this makes it so that every time we open this page the title slides to the left. But what I we add a new page after this one?
I accidentally found out this bug: every time we open operator info page, the title slides left. But when we change to a page on the left of this one, the title should slide right, but it still slides left, here's a video demonstrating the issue: https://streamable.com/14a6cc

hellypi (changes requested):

new-lamassu-admin/src/App.js
josepfo: yeah, I talked to Rafa about this. It's a little hard to animate the title on this page because of how it's passed onto app.js. I think this is a fine workaround for now just to be consistent with the other pages.

josepfo: as for the other bug, I'll copy-paste your finding in the card, probably it's better to create a new one for this.

RafaelTaranto (approved):

PR #615, opened 2021-02-19
Fix 7.4 to 7.5 migration config

josepfo:


josepfo:
[7.4 to 7.5 database migration script](https://trello.com/c/d4gBVvGP/593-74-to-75-database-migration-script)

lib/config-migration.js
```diff
@@ -407,7 +411,11 @@ function migrateComplianceTriggers (config) {
       createTrigger(requirements.sanctions, global.sanctionsVerificationThreshold)
     )
   }
-
+  if (global.hardLimitVerificationActive && _.isNumber(global.hardLimitVerificationThreshold)) {
```
chaotixkilla, line 420: Isn't checking for `global.hardLimitVerificationActive` redundant?

chaotixkilla (changes requested):
Just a small detail to simplify a condition

lib/config-migration.js
josepfo: I'm not sure about this, but I believe it's possible that on 7.4 admin the hardLimitVerificationActive is disabled (false), and there's a threshold somewhere in the config, for example, if the hardLimit was ever enabled.

chaotixkilla: Ah sorry, I read the variable wrong. You're correct!

chaotixkilla (approved):

lib/config-migration.js
```diff
@@ -344,24 +344,28 @@ function migrateComplianceTriggers (config) {
...
-      cashDirection: 'both',
+      direction: 'both',
+      suspensionDays: null,
       threshold,
       thresholdDays: 1,
       triggerType: triggerTypes.volume,
       requirement
     }
+    if (requirement === 'suspend') triggerConfig.suspensionDays = 1
+    return triggerConfig
```
RafaelTaranto: Better to add a new param and passDown suspensionDays from that

RafaelTaranto (changes requested):

lib/config-migration.js
```diff
@@ -344,24 +347,29 @@ function migrateComplianceTriggers (config) {
...
-    threshold
+    threshold,
+    suspensionDays = null
   ) {
-    return {
+    const triggerConfig = {
       id: uuid.v4(),
-      cashDirection: 'both',
+      direction: 'both',
+      suspensionDays: suspensionDays,
```
RafaelTaranto: Ok, I still have one nitpick here. We'll be creating suspensionDays for all triggers but that should be done only for suspend. 

```js
if (!triggerType === 'suspend) return triggerConfig

return _.assign(triggerConfig, { suspensionDays })
```

RafaelTaranto (changes requested):

RafaelTaranto (approved):

PR #618, opened 2021-02-22
green background on new Rows on edit tables

brunoNevess:


brunoNevess:
[Background color for new rows on editable tables should be light green](https://trello.com/c/9LTssnBN/585-background-color-for-new-rows-on-editable-tables-should-be-light-green)

RafaelTaranto (approved):

PR #619, opened 2021-02-22
crypto dropdown stays open after selection

brunoNevess:


brunoNevess:
[Leave the dropdown open on the crypto currency dropdown](https://trello.com/c/Hxz4abTL/591-leave-the-dropdown-open-on-the-crypto-currency-dropdown)

new-lamassu-admin/src/components/inputs/base/Autocomplete.js
```diff
@@ -14,6 +14,7 @@ const Autocomplete = ({
   multiple,
   onChange,
   labelProp,
+  keepOpenProp,
```
chaotixkilla, line 17: Shouldn't this `keepOpenProp` be also on the `<MAutoComplete>`  node? It works as it currently is, but other props are being explicitly passed on the node. It's more for consistency sake

new-lamassu-admin/src/pages/Commissions/helper.js
```diff
@@ -85,14 +85,15 @@ const getOverridesFields = (getData, currency, auxElements) => {
     },
     {
       name: 'cryptoCurrencies',
-      width: 280,
+      width: 316,
```
chaotixkilla, line 88: Any specific reason for this change?

chaotixkilla (changes requested):

PR #620, opened 2021-02-22
Remove geth option for ETH

josepfo:


josepfo:
[Remove 'geth' as option when setting up Ethereum](https://trello.com/c/bCRxZfiP/595-remove-geth-as-option-when-setting-up-ethereum)

RafaelTaranto (approved):

PR #622, opened 2021-02-23
refactor: prevent recursive call on zec sends

josepfo:


josepfo:
[Shielded ZEC support](https://trello.com/c/9effiH90/232-shielded-zec-support)

lib/plugins/wallet/zcashd/zcashd.js
```diff
@@ -46,11 +47,36 @@ function balance (account, cryptoCode) {
...
+        .then(res => {
+          const status = _.get('status', res[0])
+          switch (status) {
+            case 'success':
+              resolve(res[0])
+              break
+            case 'failed':
+              throw new pRetry.AbortError(res[0].error)
+            case 'executing':
+              reject(res[0].error)
```
RafaelTaranto:
```suggestion
              reject('operation still executing')
```

josepfo: Should we go for an error instead reject(new Error('operation still executing'))?

RafaelTaranto (approved):

PR #623, opened 2021-02-23
Feat: make queries with schema changing

hellypi:


hellypi:
[Create custom SQL queries for use with schema changing](https://trello.com/c/EW88QE7T/603-create-custom-sql-queries-for-use-with-schema-changing)

lib/extendedQueries.js
```diff
@@ -0,0 +1,56 @@
+const _ = require('lodash/fp')
+
+const concatSchema = (qry, tables) => {
+  const schemaName = 'public' // fetch schema name from Async Local Storage here
```
josepfo: Is this comment necessary?

josepfo (commented):
If you think it's worth having the comment, leave it. Everything else looks good.

lib/extendedQueries.js
hellypi: well not really no :)
it was just so I could remember why that variable exists if I forgot, before we implement ALS

chaotixkilla (approved):

lib/db.js
```diff
@@ -4,9 +4,32 @@ const _ = require('lodash/fp')
 const psqlUrl = require('../lib/options').postgresql
 const logger = require('./logger')
 const eventBus = require('./event-bus')
+const extendedQueries = require('./extendedQueries')
 
 const pgp = Pgp({
   pgNative: true,
+  extend (obj, dbContext) {
+    obj.taskEx = function () {
```
RafaelTaranto: Are we planning on using taskEx outside of the extendedQueries file? if not we might wanna prepend a underscore to it's name and maybe simplify the logic here.

```js
const getSchema() => {}
const getDefaultSchema() => {}

const searchPathWrapper = (t, cb) => {
    return t.none('SET search_path to $1:name', [schema])
      .then(cb.bind(t, t))
      .finally(t.none('set search_path to $1~', [getDefaultSchema()])))
}

const pgp = Pgp({
  pgNative: true,
  obj._taskEx = myTx = function(cb) {
    const schema = getSchema()
    
    if (!schema) {
      return Promise.reject(new Error('No schema selected, cannot complete query'))
    }
    
    return obj.task.call(this, args.options, t => searchPathWrapper(t, cb))
  },
   obj.query = (query, values, qrm) => obj._taskEx(() => obj.query(query, values, qrm))
}
```

RafaelTaranto: This is untested code but I'm pretty sure it would be enough.

I'm also starting to like your idea of not prepending the query names with $, imagine that the query resetting the search_path fails, a query that is not using $ appended version would be able to perform normally and might leak some bugs.

RafaelTaranto (changes requested):

lib/db.js
RafaelTaranto: Looks like this was not changed .

lib/db.js
```diff
@@ -5,8 +5,77 @@ const psqlUrl = require('../lib/options').postgresql
...
+    .then(res => {
+      return t.none('SET search_path TO $1:name', [getDefaultSchema()]).then(() => {
+        return res
+      })
+    })
+    .catch(err => {
+      return t.none('SET search_path TO $1:name', [getDefaultSchema()]).then(() => {
+        throw err
+      })
+    })
```
RafaelTaranto:
```suggestion
      .finally(t.none('set search_path to $1~', [getDefaultSchema()])))
```

RafaelTaranto (changes requested):

RafaelTaranto (approved):

RafaelTaranto (approved):

PR #625, opened 2021-02-24
Handle errors when a wallet isn't configured

josepfo:


josepfo:
[Lack of a configured wallet brings down server and machines](https://trello.com/c/1tuPODPV/300-lack-of-a-configured-wallet-brings-down-server-and-machines)

lib/plugins.js
```diff
@@ -31,7 +31,7 @@ const mapValuesWithKey = _.mapValues.convert({
 
 const TRADE_TTL = 2 * T.minutes
 const STALE_TICKER = 3 * T.minutes
-const STALE_BALANCE = 3 * T.minutes
+const STALE_BALANCE = 15 * T.seconds
```
RafaelTaranto: We don't really want to change this.

RafaelTaranto (changes requested):

lib/plugins.js
josepfo: Hmm ok, I wasn't 100% sure about this either. Should I try to prevent the coin from staying too long on the screen without messing with the cache time?

RafaelTaranto (approved):

PR #630, opened 2021-03-01
Display fiat amount and currency on same line

josepfo:


josepfo:
[Ensure fiat amount and currency display on same line](https://trello.com/c/4wOO84IY/617-ensure-fiat-amount-and-currency-display-on-same-line)

new-lamassu-admin/src/pages/Transactions/Transactions.js
```diff
@@ -117,7 +117,7 @@ const Transactions = () => {
     },
     {
       header: 'Cash',
-      width: 104,
+      width: 110,
```
RafaelTaranto: How many more chars we can fit in this amount of pixels?

RafaelTaranto (changes requested):

new-lamassu-admin/src/pages/Transactions/Transactions.js
josepfo: If the amount reaches 4 digits it goes to a new line. Should I increase the width even more?

RafaelTaranto: Yeah, put date and time together and get as much as you can here and in crypto

RafaelTaranto (approved):

PR #632, opened 2021-03-02
Feat/use new schema changing queries

hellypi:
starts with Fix: revert package* changes


hellypi:
[Make l-s use the new schema-changing queries](https://trello.com/c/fp12IX8Z/605-make-l-s-use-the-new-schema-changing-queries)

hellypi (commented):
.

lib/compute-schema.js
```diff
@@ -0,0 +1,12 @@
...
+const computeSchema = (req, res, next) => {
+  const store = new Map()
+  store.set('schema', 'public')
+  store.set('defaultSchema', 'ERROR_SCHEMA')
+  asyncLocalStorage.run(store, () => {
+    next()
+  })
+}
+
+module.exports = computeSchema
```
josepfo: Seems like the first 3 lines of this function could be replaced by the defaultStore() from ./async-storage. What do you think about using it to create the store here?

lib/db.js
```diff
@@ -23,13 +48,17 @@ eventBus.subscribe('log', args => {
...
-  const msgToSave = message ? message : _.get('message', meta)
+  const msgToSave = message || _.get('message', meta)
 
   const sql = `insert into server_logs
   (id, device_id, message, log_level, meta) values ($1, $2, $3, $4, $5) returning *`
-
-  db.one(sql, [uuid.v4(), '', msgToSave, level, meta])
-    .then(_.mapKeys(_.camelCase))
+  // need to set AsyncLocalStorage (ALS) for this function as well
+  // because this module is imported before ALS is set up on app.js
```
josepfo, line 56: Are these still useful to understand the code?

josepfo (commented):
Left you some comments. But looks good.

lib/compute-schema.js
hellypi: Since this is temporary(?) hard-coded stuff I didn't want to use defaultStore() but I think you're right

josepfo (approved):

PR #636, opened 2021-03-08
Fix: render machine and crypto in commissions page

josepfo:


josepfo:
[Fix rendering commissions overrides breaks new-admin](https://trello.com/c/e7rnEz6k/626-fix-rendering-commissions-overrides-breaks-new-admin)

RafaelTaranto (approved):

PR #638, opened 2021-03-09
Green background on new rows

josepfo:


josepfo:
[Background color for new rows on editable tables should be light green](https://trello.com/c/9LTssnBN/585-background-color-for-new-rows-on-editable-tables-should-be-light-green)

RafaelTaranto (approved):

PR #640, opened 2021-03-12
Fix cassette number on fiat balance overrides

josepfo:


josepfo:
[Fix cassette number on fiat balance overrides](https://trello.com/c/5X529YZo/632-fix-cassette-number-on-fiat-balance-overrides)

chaotixkilla (approved):

RafaelTaranto (approved):

PR #641, opened 2021-03-12
Fix: cashout row stripped when editing

josepfo:


josepfo:
[Cashout row stripped when editing](https://trello.com/c/CGEkUilN/633-cashout-row-stripped-when-editing)

chaotixkilla (approved):

RafaelTaranto (approved):

PR #653, opened 2021-03-20
Fix font size on commissions and cashout

josepfo:


josepfo:
[Font size change changes on commissions, cash-out.](https://trello.com/c/PQOAmrkt/594-font-size-change-changes-on-commissions-cash-out)

chaotixkilla (approved):

new-lamassu-admin/src/components/editableTable/Row.styles.js
```diff
@@ -23,7 +23,8 @@ export default {
     }
   },
   suffix: {
-    marginLeft: 7
+    marginLeft: 7,
+    fontSize: '12px'
```
RafaelTaranto: Is there a reason why this is 12px? I'm pretty sure we're using odd numbers for our fonts. Also this should be using something from the typography module instead of hard coding.

new-lamassu-admin/src/pages/Cashout/helper.js
```diff
@@ -21,6 +21,12 @@ const DenominationsSchema = Yup.object().shape({
     .max(currencyMax)
 })
 
+const boldStyle = () => {
+  return {
+    fontWeight: 'bold'
+  }
+}
+
```
RafaelTaranto: If this is something that has to be common in multiple screen we should add a `boldLabel` option on the element fields and move this logic into the editableTable component

RafaelTaranto (changes requested):

new-lamassu-admin/src/components/editableTable/Row.styles.js
josepfo: According to the card's description and the design, the suffix should be 12px and the number 16px.

RafaelTaranto (approved):

PR #662, opened 2021-03-24
Use arrow button to navigate to machine page on dashboard, instead of having the whole row clickable

chaotixkilla:


chaotixkilla:
[Add round navigation button next to machine names on Dashboard](https://trello.com/c/46caZAo3/666-add-round-navigation-button-next-to-machine-names-on-dashboard)

josepfo (approved):

PR #657, opened 2021-03-23
Fix error messages compliance triggers

josepfo:

PR #663, opened 2021-03-24
Fix validation message on triggers wizard

chaotixkilla:


chaotixkilla:
[Correct validation message in New Compliance Trigger](https://trello.com/c/9Ogvnmx6/664-correct-validation-message-in-new-compliance-trigger)

new-lamassu-admin/src/pages/Triggers/Wizard.js
```diff
@@ -236,21 +272,28 @@ const Wizard = ({ onClose, save, error, currency }) => {
         />
         <Formik
           validateOnBlur={false}
-          validateOnChange={false}
+          validateOnChange={true}
```
josepfo, line 275: Maybe this should remain false?

new-lamassu-admin/src/pages/Triggers/helper.js
```diff
@@ -223,18 +223,48 @@ const typeSchema = Yup.object()
...
+        txAmount: threshold => threshold.threshold > 0,
         txVolume: threshold =>
-          threshold.threshold >= 0 && threshold.thresholdDays >= 0,
+          threshold.threshold >= 0 && threshold.thresholdDays > 0,
         txVelocity: threshold =>
-          threshold.threshold >= 0 && threshold.thresholdDays >= 0,
-        consecutiveDays: threshold => threshold.thresholdDays >= 0
+          threshold.threshold > 0 && threshold.thresholdDays > 0,
+        consecutiveDays: threshold => threshold.thresholdDays > 0
       }
```
josepfo, line 260: I was thinking about doing similar custom messages for the validation when editing triggers, but to be consistent with the rest of the code I validated each field individually and used the default yup messages. You can see that here https://github.com/lamassu/lamassu-server/pull/657 and use this for reference https://github.com/jquense/yup/blob/91ace1e8be3fc23c775ec8117c47b406bf29da4a/src/locale.ts. So maybe we should go towards the same approach. Can you ask Rafael about this? he will know better.

new-lamassu-admin/src/pages/Triggers/Wizard.js
chaotixkilla: Having this on false, would make the form much less responsive to user interaction

josepfo: yeah but I once worked on a card where I was supposed to force the forms validation only after submit. Check this with Rafael if you can.

PR #664, opened 2021-03-25
Feat cashbox batch table and update bills

josepfo:


josepfo:
[Cashbox batches](https://trello.com/c/MsXqmj1l/445-cashbox-batches)

lib/cashbox-batches.js
```diff
@@ -0,0 +1,18 @@
...
+    FROM cash_in_txs
+    WHERE bills.cash_in_txs_id = cash_in_txs.id AND
+          cash_in_txs.device_id = $2 AND 
+          bills.cashbox_batch_id IS NULL
+  `
+  return db.oneOrNone(sql, [rec.deviceId])
+    .then(res => {
+      return db.oneOrNone(sql2, [res.id, res.device_id])
+    })
+}
```
hellypi: Queries executed in sequence that depend on the returned values of the previous, should be done either inside db.tx or db.task
My opinion is that this should be done inside a transaction (db.tx), because in case of error on the second query, a rollback happens and no change caused by the first query occurs
https://vitaly-t.github.io/pg-promise/Database.html#tx

hellypi (changes requested):

hellypi (approved):

lib/cashbox-batches.js
```diff
@@ -0,0 +1,20 @@
...
+    WHERE bills.cash_in_txs_id = cash_in_txs.id AND
+          cash_in_txs.device_id = $2 AND 
+          bills.cashbox_batch_id IS NULL
+  `
+  return db.tx(async t => {
+    const newBatch = await t.oneOrNone(sql, [rec.deviceId])
+    return t.oneOrNone(sql2, [newBatch.id, newBatch.device_id])
+  })
+    .then(data => data)
+    .catch(error => error)
```
RafaelTaranto: This feels redundant

migrations/1616528363530-add_cashbox_batches.js
```diff
@@ -0,0 +1,26 @@
+var db = require('./db')
+
+exports.up = function (next) {
+  var sqls = [
+    `create table cashbox_batches (
+      id serial PRIMARY KEY,
```
RafaelTaranto: I don't think we're using serial IDs on any table, uuid.v4() is probably better here.

RafaelTaranto (changes requested):

migrations/1616528363530-add_cashbox_batches.js
```diff
@@ -0,0 +1,26 @@
...
+  var sqls = [
+    `create table cashbox_batches (
+      id uuid PRIMARY KEY,
+      device_id text REFERENCES devices (device_id),
+      created timestamptz NOT NULL default now()
+    )`,
+
+    `ALTER TABLE bills ADD COLUMN legacy boolean DEFAULT false`,
+
+    `ALTER TABLE bills ADD COLUMN cashbox_batch_id int`,
```
RafaelTaranto: `Int` here should be UUID

RafaelTaranto (changes requested):

migrations/1616528363530-add_cashbox_batches.js
josepfo: I think that was fixed here https://github.com/lamassu/lamassu-server/pull/679/files

RafaelTaranto (approved):

PR #665, opened 2021-03-25
Reuse bill selector component to commission and cash out screens

chaotixkilla:


chaotixkilla:
[Implement denomination picker on Minimum TX and Cash Out panels](https://trello.com/c/bAKl1kwB/657-implement-denomination-picker-on-minimum-tx-and-cash-out-panels)

hellypi (approved):
Looks good

new-lamassu-admin/src/pages/Commissions/components/CommissionsDetails.js
```diff
@@ -6,15 +6,15 @@ import Section from 'src/components/layout/Section'
...
+  getSchema,
   getOverridesSchema,
   defaults,
   overridesDefaults,
   getOrder
 } from 'src/pages/Commissions/helper'
 
 const CommissionsDetails = memo(
-  ({ config, currency, data, error, save, saveOverrides }) => {
+  ({ config, locale, currency, data, error, save, saveOverrides, classes }) => {
```
RafaelTaranto: I'm confused over this classes here, doesn't seem to be used.

new-lamassu-admin/src/pages/Commissions/Commissions.js
```diff
@@ -16,6 +16,13 @@ import CommissionsList from './components/CommissionsList'
...
   listViewButton: {
     marginLeft: 4
+  },
+  autoComplete: {
+    width: '100%'
+  },
+  bold: {
+    fontWeight: 'bold',
+    height: 200
   }
```
RafaelTaranto: Don't seem to be used

new-lamassu-admin/src/utils/bill-options.js
```diff
@@ -0,0 +1,12 @@
+import * as R from 'ramda'
+
+const getBillOptions = R.curry((locale, denomiations) => {
+  const currency = R.prop('fiatCurrency')(locale)
+  return R.compose(
+    R.map(code => ({ code: parseInt(code), display: code })),
+    R.keys,
+    R.path([currency])
```
RafaelTaranto, line 8: I don't think ramda is null safe, does this work even when picking a fiat that has no bill definition?

RafaelTaranto (changes requested):

new-lamassu-admin/src/utils/bill-options.js
chaotixkilla: I tested with USD and COP, with and without bill definitions, respectively, and everything seems to work correctly. When there's no bill definitions, it returns an empty array and allows the user to insert a number

RafaelTaranto (approved):

PR #666, opened 2021-03-26
Feat/pazuz wallet plugin

hellypi:


hellypi:
[Create the pazuz wallet plugin on the lamassu codebase](https://trello.com/c/FY8G0Rgx/621-create-the-pazuz-wallet-plugin-on-the-lamassu-codebase)

lib/plugins/wallet/pazuz-wallet/pazuz-wallet.js
```diff
@@ -0,0 +1,108 @@
+const https = require('https')
+const BN = require('../../../bn')
+const E = require('../../../error')
+
+const axios = require('axios').create({
+  // TODO: get rejectUnauthorized true to work
```
chaotixkilla, line 10: Does this TODO still apply?

lib/plugins/wallet/pazuz-wallet/pazuz-wallet.js
```diff
@@ -0,0 +1,108 @@
...
+  baseURL: 'https://localhost:5555/api/',
```
chaotixkilla: This might depend on an .env var in the future

lib/plugins/wallet/pazuz-wallet/pazuz-wallet.js
```diff
@@ -0,0 +1,108 @@
...
+})
+
+const SUPPORTED_COINS = ['BTC', 'ZEC', 'LTC', 'BCH', 'DASH', 'ETH']
+
+function checkCryptoCode (cryptoCode) {
+  if (!SUPPORTED_COINS.includes(cryptoCode)) {
+    return Promise.reject(new Error('Unsupported crypto: ' + cryptoCode))
+  }
+
+  return Promise.resolve()
```
chaotixkilla: Could be solved with a ternary. Plus, something like lodash should be used, instead of the vanilla JS library. Rafael told me that a while ago, so you should check if that still applies

lib/plugins/wallet/pazuz-wallet/pazuz-wallet.js
```diff
@@ -0,0 +1,108 @@
...
+  throw new E.NotImplementedError()
+}
+
+function sweep (account, cryptoCode, hdIndex, settings, operatorId) {
+  throw new E.NotImplementedError()
+}
+
+function isStrictAddress (cryptoCode, toAddress, settings, operatorId) {
+  throw new E.NotImplementedError()
+}
```
chaotixkilla, line 95: I like this approach, good job

chaotixkilla (changes requested):
Some very minor stuff, otherwise, looks great

lib/plugins/wallet/pazuz-wallet/pazuz-wallet.js
hellypi: yep

chaotixkilla (approved):

lib/wallet.js
```diff
@@ -21,6 +21,10 @@ function computeSeed (masterSeed) {
   return hkdf(masterSeed, 32, { salt: 'lamassu-server-salt', info: 'wallet-seed' })
 }
 
+function computeOperatorId (masterSeed) {
+  return hkdf(masterSeed, 16, { salt: 'lamassu-server-salt', info: 'operator-id' }).toString('hex')
+}
+
```
RafaelTaranto, line 27: We should have a module for this and use everywhere it applies. Also, this won't really work for pazuz, we need to get operatorID from asyncLocalStorage, no?

RafaelTaranto (changes requested):

lib/wallet.js
RafaelTaranto: To be done after some merges.

RafaelTaranto (approved):

PR #670, opened 2021-03-29
Fix validation message on triggers wizard

chaotixkilla:


chaotixkilla:
[Correct validation message in New Compliance Trigger](https://trello.com/c/9Ogvnmx6/664-correct-validation-message-in-new-compliance-trigger)

new-lamassu-admin/src/pages/Triggers/helper.js
```diff
@@ -390,20 +448,34 @@ const type = currency => ({
...
-        .transform(() => null)
+const requirementSchema = Yup.object()
+  .shape({
+    requirement: Yup.object({
+      requirement: Yup.string().required(),
+      suspensionDays: Yup.number().when('requirement', {
+        is: value => value === 'suspend',
+        then: Yup.number()
+          .required()
+          .min(1),
```
josepfo: I think this validation is not necessary since we're doing it in the test function, right?
Could be replaced by:

 suspensionDays: Yup.number()
   .nullable()
   .transform(transformNumber)

josepfo (approved):

RafaelTaranto (approved):

PR #671, opened 2021-04-05
Feat: add rbf checking to blockcypher

josepfo:


josepfo:
[Add RBF checking to Blockcypher plugin](https://trello.com/c/EeoUhzbE/660-add-rbf-checking-to-blockcypher-plugin)

lib/plugins/zero-conf/blockcypher/blockcypher.js
```diff
@@ -21,17 +24,32 @@ function authorize (account, toAddress, cryptoAtoms, cryptoCode) {
...
+                }, values)
+                const sumTxRefs = txrefs => _.sumBy(txref => highConfidence(confidence, txref), txrefs)
+                const authorizedValue = sumTxRefs(data.txrefs) + sumTxRefs(data.unconfirmed_txrefs)
+                return cryptoAtoms.lte(authorizedValue)
+              })
+          } else {
+            const sumTxRefs = txrefs => _.sumBy(txref => highConfidence(confidence, txref), txrefs)
+            const authorizedValue = sumTxRefs(data.txrefs) + sumTxRefs(data.unconfirmed_txrefs)
+            return cryptoAtoms.lte(authorizedValue)
+          }
```
chaotixkilla: Correct me if I'm wrong, but lines 44-46 and 49-51 are the exact same, and 44-46 have no relation with the other code that is being executed inside the condition

```suggestion
          if (isBitcoindAvailable && isRBFEnabled && data.unconfirmed_txrefs) {
            _.map(unconfirmedTxref => {
              promises.push(new Promise((resolve, reject) => { resolve(fetchRBF(unconfirmedTxref.tx_hash)) }))
            }, data.unconfirmed_txrefs)
            
            Promise.all(promises)
              .then(values => {
                _.map(rbfInfo => {
                  _.map(unconfirmedTxref => {
                    if (rbfInfo.tx_hash === unconfirmedTxref.tx_hash) unconfirmedTxref.rbf = rbfInfo['bip125-replaceable']
                  }, data.unconfirmed_txrefs)
                }, values)                
              })
          }
          
          const sumTxRefs = txrefs => _.sumBy(txref => highConfidence(confidence, txref), txrefs)
          const authorizedValue = sumTxRefs(data.txrefs) + sumTxRefs(data.unconfirmed_txrefs)
          return cryptoAtoms.lte(authorizedValue)
```

lib/plugins/zero-conf/blockcypher/blockcypher.js
```diff
@@ -21,17 +24,32 @@ function authorize (account, toAddress, cryptoAtoms, cryptoCode) {
...
+            _.map(unconfirmedTxref => {
+              promise.push(new Promise((resolve, reject) => { resolve(fetchRBF(unconfirmedTxref.tx_hash)) }))
+            }, data.unconfirmed_txrefs)
+            return Promise.all(promise)
+              .then(values => {
+                _.map(rbfInfo => {
+                  _.map(unconfirmedTxref => {
+                    if (rbfInfo.tx_hash === unconfirmedTxref.tx_hash) unconfirmedTxref.rbf = rbfInfo['bip125-replaceable']
+                  }, data.unconfirmed_txrefs)
+                }, values)
```
chaotixkilla: A bit of nitpicking, and mostly for my hate of nested looping unless strictly necessary, but can't this be a `zip` between `values` and `data.unconfirmed_txrefs`, which would result in an array of pairs `[rbfInfo.tx_hash, unconfirmedTxref.tx_hash]` to which you can map a comparing function to each pair.

As I said, a bit of nitpicking, you can easily ignore this, as I don't think the effort would win over the performance boost, but something to consider about readability

chaotixkilla (changes requested):

lib/plugins/zero-conf/blockcypher/blockcypher.js
josepfo: I did it this way to keep the highConfidence really simple but yeah that nested loop should be avoided. Doing what you suggested I think it requires the highConfidence to change a lot tho, right?

lib/plugins/zero-conf/blockcypher/blockcypher.js
josepfo: yeah sumTxRefs doesn't but authorizedValue does depend on data

lib/plugins/zero-conf/blockcypher/blockcypher.js
chaotixkilla: Maybe. I'm not the most experienced person with this section of code, so we can pass this and see what Rafael thinks.

Thinking better, you're not using the `highConfidence()` from lines 40-44, why would it be an issue?

chaotixkilla (approved):

lib/plugins/zero-conf/blockcypher/blockcypher.js
```diff
@@ -21,17 +24,28 @@ function authorize (account, toAddress, cryptoAtoms, cryptoCode) {
       })
 
       const confidence = account.confidenceFactor
-
+      const isRBFEnabled = account.rbf
```
RafaelTaranto, line 26: Shouldn't this have a UI change as well? How did you add this rbf option?

lib/plugins/zero-conf/blockcypher/blockcypher.js
```diff
@@ -21,17 +24,28 @@ function authorize (account, toAddress, cryptoAtoms, cryptoCode) {
...
+                const unconfirmedTxsRBF = _.fromPairs(values)
+                const sumTxRefs = txrefs => _.sumBy(txref => highConfidence(confidence, txref, unconfirmedTxsRBF[txref.tx_hash]), txrefs)
+                const authorizedValue = sumTxRefs(data.txrefs) + sumTxRefs(data.unconfirmed_txrefs)
+                return cryptoAtoms.lte(authorizedValue)
+              })
+          } else {
+            const sumTxRefs = txrefs => _.sumBy(txref => highConfidence(confidence, txref), txrefs)
+            const authorizedValue = sumTxRefs(data.txrefs) + sumTxRefs(data.unconfirmed_txrefs)
+            return cryptoAtoms.lte(authorizedValue)
+          }
```
RafaelTaranto:
```js
if (isBitcoindAvailable && isRBFEnabled && data.unconfirmed_txrefs) {
  const promises = _.map(unconfirmedTxref => fetchRBF(unconfirmedTxref.tx_hash)), data.unconfirmed_txrefs)
  return Promise.all(promise)
  //rest of logic
}

const sumTxRefs = txrefs => _.sumBy(txref => highConfidence(confidence, txref), txrefs)
const authorizedValue = sumTxRefs(data.txrefs) + sumTxRefs(data.unconfirmed_txrefs)
return cryptoAtoms.lte(authorizedValue)
```

RafaelTaranto (changes requested):

lib/plugins/zero-conf/blockcypher/blockcypher.js
josepfo: There's a card attached with the UI part

RafaelTaranto (approved):

PR #672, opened 2021-04-05
feat: rbf checkbox in blockcypher settings

josepfo:


josepfo:
[Add RBF toggle to blockcypher in third-party settings](https://trello.com/c/nCajAo7W/674-add-rbf-toggle-to-blockcypher-in-third-party-settings)

chaotixkilla (changes requested):
The checkbox logic changes should only be extended towards supporting the disabled message. Any gql logic, label, wallet logic, etc, etc should only be sent to the Checkbox component as props and not be built in the component, as this will break all other uses of this component.

chaotixkilla (approved):
Much better now, everything seems fine. Good job!

RafaelTaranto (approved):

PR #677, opened 2021-04-12
feat: advanced options screen triggers

josepfo:


josepfo:
[Implement the advanced options screen for the triggers page.](https://trello.com/c/tIAlUyuu/691-implement-the-advanced-options-screen-for-the-triggers-page)

new-lamassu-admin/src/pages/Triggers/Triggers.js
```diff
@@ -11,10 +11,13 @@ import { Table as EditableTable } from 'src/components/editableTable'
 import { Switch } from 'src/components/inputs'
 import TitleSection from 'src/components/layout/TitleSection'
 import { P, Label2, H2 } from 'src/components/typography'
+import { ReactComponent as ReverseListingViewIcon } from 'src/styling/icons/circle buttons/listing-view/white.svg'
+import { ReactComponent as ListingViewIcon } from 'src/styling/icons/circle buttons/listing-view/zodiac.svg'
```
chaotixkilla: According to the spec, the icons to use are the ones at `src/styling/icons/circle buttons/settings`, and not `listing-view`

Change the icon path and names here and lines 85 and 86

```suggestion
import { ReactComponent as ReverseSettingsIcon} from 'src/styling/icons/circle buttons/settings/white.svg'
import { ReactComponent as SettingsIcon } from 'src/styling/icons/circle buttons/settings/zodiac.svg'
```

new-lamassu-admin/src/pages/Triggers/Triggers.js
```diff
@@ -74,72 +78,96 @@ const Triggers = () => {
...
   return (
     <>
-      <TitleSection title="Compliance Triggers" className={classes.tableWidth}>
-        <Box display="flex" alignItems="center">
+      <TitleSection
+        title="Compliance Triggers"
+        button={{
+          text: 'Advanced settings',
+          icon: ListingViewIcon,
+          inverseIcon: ReverseListingViewIcon,
```
chaotixkilla:
```suggestion
          icon: SettingsIcon,
          inverseIcon: ReverseSettingsIcon,
```

new-lamassu-admin/src/pages/Triggers/Triggers.js
```diff
@@ -74,72 +78,96 @@ const Triggers = () => {
...
-              {rejectAddressReuse ? 'On' : 'Off'}
-            </Label2>
-            <Tooltip width={304}>
-              <P>
-                This option requires a user to scan a different cryptocurrency
-                address if they attempt to scan one that had been previously
-                used for a transaction in your network
-              </P>
-            </Tooltip>
+            justifyContent="flex-end">
```
chaotixkilla, line 125: With the `className` defined, couldn't you add the `display="flex"` and `justifyContent="flex-end"` to `classes.tableWidth` and give it a more relevant name?

chaotixkilla: Just noticed this section of the code is not new. You can choose to ignore this comment

chaotixkilla (changes requested):
Just very minor stuff, otherwise, looks great

chaotixkilla (approved):

migrations/1617983833835-add-default-advanced-trigger-settings.js
```diff
@@ -0,0 +1,21 @@
...
+exports.up = function (next) {
+  const triggersDefault = {
+    triggersConfig_expirationTime: 'Forever',
+    triggersConfig_automation: 'Automatic'
+  }
+
+  return saveConfig(triggersDefault)
+    .then(() => next())
+    .catch(err => {
+      if (err.message === 'lamassu-server is not configured') {
```
RafaelTaranto: This should never happen, it's stuff from the old settings-loader. Even if it's possible to fail on the new-settings-loader we want to break the migration here in case of error.

migrations/1617983833835-add-default-advanced-trigger-settings.js
```diff
@@ -0,0 +1,21 @@
...
+
+exports.up = function (next) {
+  const triggersDefault = {
+    triggersConfig_expirationTime: 'Forever',
+    triggersConfig_automation: 'Automatic'
+  }
+
+  return saveConfig(triggersDefault)
+    .then(() => next())
+    .catch(err => {
```
RafaelTaranto, line 11: You can probably remove this catch

RafaelTaranto (changes requested):

migrations/1617983833835-add-default-advanced-trigger-settings.js
RafaelTaranto: Actually I was wrong here, you wanna call next(err)

migrations/1617983833835-add-default-advanced-trigger-settings.js
josepfo: So this if shouldn't exist? I've been doing this for every migration. The only thing missing here is the return next(err) at the end:
```suggestion
    .catch(err => {
      if (err.message === 'lamassu-server is not configured') {
        return next()
      }
      console.log(err.message)
      return next(err)
    })
```

RafaelTaranto: What other migration has this? The `lamassu server is not configured` part is related to the 7.4 config schema so it should only appear on the 7.4 -> 7.5 config migration.

josepfo: https://github.com/lamassu/lamassu-server/pull/704/files This one for instance. I think it needs fixing too.

RafaelTaranto (approved):

PR #680, opened 2021-04-13
fix: remove crypto amount rounding

josepfo:


josepfo:
[Don't round crypto amount in admin's Transactions panel](https://trello.com/c/3ZK0hj15/681-dont-round-crypto-amount-in-admins-transactions-panel)

new-lamassu-admin/src/pages/Transactions/Transactions.js
```diff
@@ -130,7 +130,7 @@ const Transactions = () => {
       textAlign: 'right',
       size: 'sm',
       view: it =>
-        `${toUnit(new BigNumber(it.cryptoAtoms), it.cryptoCode).toFormat(5)} ${
+        `${toUnit(new BigNumber(it.cryptoAtoms), it.cryptoCode)} ${
```
RafaelTaranto, line 133: We don't want to remove the rounding as a whole, we should do the same as the machine does. 

Also is there even enough space here to show the additional digits?

RafaelTaranto (changes requested):

RafaelTaranto (approved):

PR #689, opened 2021-04-15
Chore: migration rename non blockcypher to none

hellypi:


hellypi:
[Rename non-blockCypher props in config to 'none'](https://trello.com/c/Jk0Lucgy/707-rename-non-blockcypher-props-in-config-to-none)

migrations/1618507684019-rename-0-conf.js
```diff
@@ -0,0 +1,24 @@
+const _ = require('lodash/fp')
+const settingsLoader = require('../lib/new-settings-loader')
+const configManager = require('../lib/new-config-manager')
+
+const isNil = val => val == null
```
chaotixkilla: You have this function available in lodash. `_.isNil()` if you want to check for `null` and `undefined`, `_.isNull()` if you want to check for `null` only

migrations/1618507684019-rename-0-conf.js
```diff
@@ -0,0 +1,24 @@
...
+
+exports.up = async function (next) {
+  const settingsPromise = settingsLoader.loadLatest()
+  const [{ config }] = await Promise.all([settingsPromise])
```
chaotixkilla: Isn't the promise redundant? You can just call await with the function which returns the promise, instead of wrapping it around another promise.
Something like this (I'm not sure if the deconstruction remains the same, so test it first):

```suggestion
  const [{ config }] = await settingsLoader.loadLatest()
```

chaotixkilla (changes requested):

chaotixkilla (approved):

migrations/1618507684019-rename-0-conf.js
```diff
@@ -0,0 +1,21 @@
...
+  const cryptoCodes = configManager.getCryptosFromWalletNamespace(config)
+  _.forEach(cryptoCode => {
+    const key = `wallets_${cryptoCode}_zeroConf`
+    const zeroConfSetting = _.get(key, config)
+    if (cryptoCode === 'BTC' && zeroConfSetting === 'blockcypher') return
+    if (!_.isNil(zeroConfSetting) && zeroConfSetting !== 'none') {
+      config[key] = 'none'
+    }
+  }, cryptoCodes)
+  return settingsLoader.saveConfig(config).then(() => next()).catch(err => next(err))
```
RafaelTaranto: You don't need the then and catch here

RafaelTaranto (changes requested):

RafaelTaranto (approved):

PR #693, opened 2021-04-21
Machine model prettifier

chaotixkilla:


chaotixkilla:
[Machine model not being prettified on admin](https://trello.com/c/KtqlZTie/388-machine-model-not-being-prettified-on-admin)

PR #694, opened 2021-04-21
fix: migrate accounts and default cash-out commissions

josepfo:


josepfo:
[v7.4 > v7.5 migration script not preserving third-party fields](https://trello.com/c/LGvWijM3/712-v74-v75-migration-script-not-preserving-third-party-fields)

migrations/1599523522436-migrate-config.js
```diff
@@ -1,34 +1,35 @@
...
-    .then(() => next())
+      .then(() => next())
   }
 
-  settingsLoader.loadLatest(false)
+  loadLatest(OLD_SETTINGS_LOADER_SCHEMA_VERSION)
     .then(async settings => ({
       settings,
       machines: await machineLoader.getMachineNames(settings.config)
     }))
```
hellypi, line 22: this .then mixed with await makes the code seem kinda weird. Can't we change the function to async and use Promise.all() to await all promises here?

something like

```JS
module.exports.up = async function (next) {
	// ...
  const settingsPromise = loadLatest(OLD_SETTINGS_LOADER_SCHEMA_VERSION)
  const machinesPromise = machineLoader.getMachineNames(settings.config)
  try {
  	const [settings, machines] = await Promise.all([settingsPromise, machinesPromise])
    // rest of code
  } catch (err) {
  	// ...
  }
}
```

migrations/1599523522436-migrate-config.js
```diff
@@ -1,34 +1,35 @@
...
-      const sql = machines 
+      const sql = machines
         ? machines.map(m => `update devices set name = '${m.name}' where device_id = '${m.deviceId}'`)
         : []
       return db.multi(sql, () => migrateConfig(settings))
     })
     .catch(err => {
-      if (err.message = 'lamassu-server is not configured')
+      if (err.message === 'lamassu-server is not configured') {
         next()
```
hellypi: *Please check with Rafael*

(according to what we've done in the past) When we catch an error, the callback "next" should be sent with the error object, notice how we call cb(err) in here - https://github.com/lamassu/lamassu-server/blob/dev/migrations/db.js

In this case if the message is 'lamassu-server is not configured' I think we should continue as normal, but if not we should call next(err)

```suggestion
  if (err.message = 'lamassu-server is not configured')
      return next()
  console.log(err.message)
  return next(err)
```

hellypi (changes requested):

hellypi (approved):

RafaelTaranto (approved):

PR #702, opened 2021-04-30
Feat: make poller work for an arbitrary number of schemas (users)

hellypi:
Commits related to this PR start (and include) from a4a77e8
For code review, relevant files changed are app.js and poller.js


hellypi:
[Change the lamassu-server poller to get it ready for pazuz (lamassu-side)](https://trello.com/c/QuZrNDvI/724-change-the-lamassu-server-poller-to-get-it-ready-for-pazuz-lamassu-side)

lib/poller.js
```diff
@@ -150,4 +150,8 @@ function setup (__settings = null, schemasToAdd = [], schemasToRemove = []) {
   return initializeEachSchema(schemasToAdd)
 }
 
-module.exports = { setup, reload }
+function getActiveSchemas () {
+  return Array.from(schemaCallbacks.keys())
```
chaotixkilla: Couldn't this be just `return _.keys(schemaCallbacks)`?

chaotixkilla (changes requested):
Has conflicts

hellypi:
PR replaced by https://github.com/lamassu/lamassu-server/pull/709

lib/poller.js
hellypi: Since it's a Map lodash doesn't work on it. And I want to return an array not a map iterator so I also have to use Array.from

PR #704, opened 2021-05-03
feat: fiat balance alerts to percentages and config migration

josepfo:


josepfo:
[Get the cashbox component up to spec](https://trello.com/c/1j4qBriH/736-get-the-cashbox-component-up-to-spec)

lib/plugins.js
```diff
@@ -570,7 +570,7 @@ function plugins (settings, deviceId) {
       }
       : null
 
-    const cassette1Alert = cashOutEnabled && device.cassette1 < notifications.fiatBalanceCassette1
+    const cassette1Alert = cashOutEnabled && (device.cassette1 / 500) < notifications.fillingPercentageCassette1
```
hellypi: wondering if there's a variable somewhere in the repo for the cash cassette max capacity, instead of just hard coding 500.

lib/plugins.js
```diff
@@ -582,7 +582,7 @@ function plugins (settings, deviceId) {
       }
       : null
 
-    const cassette2Alert = cashOutEnabled && device.cassette2 < notifications.fiatBalanceCassette2
+    const cassette2Alert = cashOutEnabled && (device.cassette2 / 500) < notifications.fillingPercentageCassette2
```
hellypi: same here

migrations/1619968992683-fiat-balance-notification-to-percent.js
```diff
@@ -0,0 +1,52 @@
+const _ = require('lodash/fp')
+const { saveConfig, loadLatest } = require('../lib/new-settings-loader')
+
+exports.up = function (next) {
+  loadLatest()
```
hellypi:
```suggestion
  return loadLatest()
```

migrations/1619968992683-fiat-balance-notification-to-percent.js
```diff
@@ -0,0 +1,52 @@
...
+        config.notifications_fiatBalanceCassette2 = (100 * (fiatBalance2 / 500)).toFixed(0)
+      }
+
+      const {
+        notifications_fiatBalanceCassette1: notifications_fillingPercentageCassette1,
+        notifications_fiatBalanceCassette2: notifications_fillingPercentageCassette2,
+        ...rest
+      } = config
+
+      config = { notifications_fillingPercentageCassette1, notifications_fillingPercentageCassette2, ...rest }
```
hellypi: Just a headsup, you don't need to send the whole config object to saveConfig(), internally the config is merged by that function before getting saved to the DB.

This has some implications:
- if the original config object is { a:1, b:2, c:3 } and you do saveConfig({ a:11, d:4 }) the resulting config is { a:11, b:2, c:3, d:4 }
- there isn't a currently supported way of deleting keys from the config object. Best we can do right now is setting the properties to null

My advice is creating a different variable with all the stuff you want to add or change and then call saveConfig with it instead of copying the whole config object onto itself.

Also, set the old fields to null. It would be something like this:

```
const newConfig = {
  notifications_fillingPercentageCassette1: config.notifications_fiatBalanceCassette1, // or use destructuring
  notifications_fillingPercentageCassette2: config.notifications_fiatBalanceCassette2,
  notifications_fiatBalanceCassette1: null,
  notifications_fiatBalanceCassette2: null,
  // etc
}
```

new-lamassu-admin/src/components/inputs/cashbox/Cashbox.js
```diff
@@ -18,20 +18,26 @@ const Cashbox = ({
...
   const classes = cashboxClasses({ percent, cashOut })
   const threshold = 51
 
   return (
-    <div className={classnames(className, classes.cashbox)}>
+    <div
+      className={
+        inFiatBalanceAlerts
+          ? classnames(className, classes.fiatBalanceAlertCashbox)
+          : classnames(className, classes.cashbox)
```
hellypi: classnames supports a better way of doing this, check this file

[Line 31 showErrorColor const and line 56 where I use it]( https://github.com/lamassu/lamassu-server/blob/9802e9505f4370be9b593b85d1592def8529c5be/new-lamassu-admin/src/pages/CustomInfoRequests/Forms/TypeFieldsSetup/ChoiceList.js)

new-lamassu-admin/src/pages/Notifications/sections/FiatBalanceAlerts.js
```diff
@@ -32,33 +27,28 @@ const FiatBalance = ({
...
-
   return (
     <Formik
       validateOnBlur={false}
       validateOnChange={false}
       enableReinitialize
       initialValues={{
-        fiatBalanceCassette1: data?.fiatBalanceCassette1 ?? '',
-        fiatBalanceCassette2: data?.fiatBalanceCassette2 ?? ''
+        fillingPercentageCassette1: data?.fillingPercentageCassette1 ?? '',
```
hellypi: This is correct but Ramda is great for this. Also it helps us avoid problems with accessing undefined variable properties when network goes down, I had a bug like that on the dashboard page
So before the return statement you can define a variable like this
```suggestion
        const fillingPercentageCassette1 = R.path(['fillingPercentageCassette1'])(data) ?? ''
```
And now in initialValues it becomes { fillingPercentageCassette1, ... }

hellypi (changes requested):
.

migrations/1619968992683-fiat-balance-notification-to-percent.js
josepfo: Good point!

hellypi (approved):

migrations/1619968992683-fiat-balance-notification-to-percent.js
```diff
@@ -0,0 +1,50 @@
...
+            override.fiatBalanceCassette1 = (100 * (override.fiatBalanceCassette1 / cassetteMaxCapacity)).toFixed(0)
+          }
+          if (override.fiatBalanceCassette2) {
+            override.fiatBalanceCassette2 = (100 * (override.fiatBalanceCassette2 / cassetteMaxCapacity)).toFixed(0)
+          }
+          const {
+            fiatBalanceCassette1: fillingPercentageCassette1,
+            fiatBalanceCassette2: fillingPercentageCassette2,
+            ...rest } = override
+          return { fillingPercentageCassette1, fillingPercentageCassette2, ...rest }
```
RafaelTaranto: Why not add the fillingPercentage straight to the override instead of replacing fiatBalance? If you do that you can just pick the fields you want to return.

RafaelTaranto (changes requested):

RafaelTaranto (approved):

RafaelTaranto (approved):

PR #706, opened 2021-05-04
fix: Record outgoing transaction fees to logs

joao-cleal:
Seems to be a problem with the current version of the 'json2csv' package, which is omiting some fields coming from the txCSV log query.
Tried to specify to include empty fields, but to no avail. 
Only worked if I specified all field to be parsed.

### DIFF
--- a/lib/new-admin/graphql/resolvers/transaction.resolver.js
+++ b/lib/new-admin/graphql/resolvers/transaction.resolver.js
@@ -5,6 +5,17 @@ const transactions = require('../../services/transactions')
 const anonymous = require('../../../constants').anonymousCustomer
 
 const transactionsLoader = new DataLoader(ids => transactions.getCustomerTransactionsBatch(ids))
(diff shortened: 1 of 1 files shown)


joao-cleal:
[Record transaction fees spent on outgoing transactions (non ETH)](https://trello.com/c/tLpM8lhx/703-record-transaction-fees-spent-on-outgoing-transactions-non-eth)

joao-cleal:
[Record transaction fees spent on outgoing transactions (ETH)](https://trello.com/c/fZlDWBvf/719-record-transaction-fees-spent-on-outgoing-transactions-eth)

chaotixkilla (approved):

RafaelTaranto (approved):

joshmh:
> Seems to be a problem with the current version of the 'json2csv' package, which is omiting some fields coming from the txCSV log query.
> Tried to specify to include empty fields, but to no avail.
> Only worked if I specified all field to be parsed.

Maybe we should replace the library then, sounds pretty annoying.

PR #708, opened 2021-05-06
fix: big number subtraction

josepfo:


josepfo:
[pending.gte is not a function](https://trello.com/c/OyzHoArA/742-pendinggte-is-not-a-function)

PR #709, opened 2021-05-06
Feat/poller multiple schemas

hellypi:
Replaces https://github.com/lamassu/lamassu-server/pull/702
Files changed in this PR: lib/poller.js and lib/app.js


hellypi:
[Change the lamassu-server poller to get it ready for pazuz (lamassu-side)](https://trello.com/c/QuZrNDvI/724-change-the-lamassu-server-poller-to-get-it-ready-for-pazuz-lamassu-side)

lib/poller.js
```diff
@@ -24,24 +28,66 @@ const LOGS_CLEAR_INTERVAL = 1 * T.day
 const SANCTIONS_INITIAL_DOWNLOAD_INTERVAL = 5 * T.minutes
 const SANCTIONS_UPDATE_INTERVAL = 1 * T.week
 const RADAR_UPDATE_INTERVAL = 5 * T.minutes
-
 const CHECK_NOTIFICATION_INTERVAL = 20 * T.seconds
-
 const PENDING_INTERVAL = 10 * T.seconds
+const CACHE_ENTRY_TTL = 3600 // seconds
```
chaotixkilla, line 33: Should just use the available `T` functions here
```suggestion
const CACHE_ENTRY_TTL = 1 * T.hour
```

lib/poller.js
```diff
@@ -69,9 +115,40 @@ function updateCoinAtmRadar () {
...
-  reload(__settings)
+function initializeEachSchema (schemas = ['public']) {
+  // for each schema set "thread variables" and do polling
+  return _.forEach(schema => {
+    const store = defaultStore()
+    store.set('schema', schema)
+    return asyncLocalStorage.run(store, () => {
+      return settingsLoader.loadLatest().then(settings => {
+        // prevent inadvertedly clearing the array without clearing timeouts
+        if (schemaCallbacks.has(schema)) throw new Error(`The shema "${schema}" cannot be initialized twice on poller`)
```
chaotixkilla: You have `schema` misspelled

```suggestion
        if (schemaCallbacks.has(schema)) throw new Error(`The schema "${schema}" cannot be initialized twice on poller`)
```

lib/poller.js
```diff
@@ -86,22 +163,36 @@ function start (__settings) {
...
+    const callbacks = schemaCallbacks.get(schema)
+    _.forEach(clearInterval, callbacks)
+    schemaCallbacks.delete(schema)
+    cachedVariables.del(schema)
+  }, schemasToRemove)
+
+  return initializeEachSchema(schemasToAdd)
+}
+
+const getActiveSchemas = () => Array.from(schemaCallbacks.keys())
```
chaotixkilla, line 196: Is there any nuance which requires the array creation to be like this, instead of something like `_.keys(schemaCallbacks)`?

lib/token-manager.js
```diff
@@ -0,0 +1,13 @@
...
+  const sql = `select * from user_tokens`
+  return db.$any(sql)
+}
+
+function revokeToken (token) {
+  const sql = `delete from user_tokens where token = $1`
+  return db.$none(sql, [token])
+}
+
+module.exports = { getTokenList, revokeToken }
```
chaotixkilla: This file is removed in dev after the authentication merge, most likely reappeared on a rebase or something. You can remove it

chaotixkilla (changes requested):
Some very minor stuff

lib/poller.js
hellypi: This library uses seconds as the base unit, not milliseconds so using 1 * T.hour here would actually make the TTL 1000 * 3600 seconds long

lib/poller.js
hellypi: Yes, the fact I'm using a Map instead of an object. I think we can change it to an object since the functionality I wanted with Map isn't necessary anymore

chaotixkilla (approved):

hellypi:
Moved to #791 791

PR #713, opened 2021-05-07
Chore: make populateSettings middleware use node-cache

hellypi:


hellypi:
[Use node-cache on lib/middlewares](https://trello.com/c/E4yaS9It/752-use-node-cache-on-lib-middlewares)

RafaelTaranto (approved):

PR #720, opened 2021-05-17
Fix: cashbox denomination value reversed

josepfo:


josepfo:
[Cashbox denomination value is reversed](https://trello.com/c/Xy9sD895/758-cashbox-denomination-value-is-reversed)

RafaelTaranto (approved):

PR #723, opened 2021-05-19
feat: apply fee discount for outgoing tx

josepfo:


josepfo:
[Add a fee priority selector for outgoing BTC transactions](https://trello.com/c/7zt6lWRX/745-add-a-fee-priority-selector-for-outgoing-btc-transactions)

lib/plugins/wallet/bitcoind/bitcoind.js
```diff
@@ -49,10 +49,45 @@ function balance (account, cryptoCode) {
   return accountBalance(cryptoCode)
 }
 
-function sendCoins (account, address, cryptoAtoms, cryptoCode) {
-  const coins = cryptoAtoms.shift(-unitScale).toFixed(8)
+function estimateFee () {
+  return fetch('estimatesmartfee', [6, 'unset'])
+    .then(result => BN(result.feerate))
+    .catch(() => {})
```
hellypi, line 57: why not console.error?

migrations/1620954224627-add-fee-priority.js
```diff
@@ -0,0 +1,20 @@
...
+        .catch(err => {
+          console.log(err.message)
+          return next(err)
+        })
+    })
+}
+
+module.exports.down = function (next) {
+  next()
+}
```
hellypi: need newline

migrations/1620954224627-add-fee-priority.js
```diff
@@ -0,0 +1,20 @@
+const { saveConfig, loadLatest } = require('../lib/new-settings-loader')
+
+exports.up = function (next) {
+  const newConfig = {
+    wallets_BTC_feeDiscount: 'Default'
+  }
+  return loadLatest()
+    .then(config => {
+      return saveConfig(newConfig)
+        .then(() => next())
```
hellypi:
```suggestion
        .then(next)
```

migrations/1620954224627-add-fee-priority.js
```diff
@@ -0,0 +1,20 @@
...
+        .catch(err => {
+          console.log(err.message)
```
hellypi: no need for this, error gets logged anyway

lib/plugins/wallet/bitcoind/bitcoind.js
```diff
@@ -49,10 +49,45 @@ function balance (account, cryptoCode) {
...
+}
 
+function calculateFeeDiscount (feeDiscount) {
+  if (!feeDiscount || feeDiscount === 'Default') return
+  return estimateFee()
+    .then(estimatedFee => {
+      if (estimatedFee) {
+        const operation = _.head(feeDiscount)
+        const percent = BN(_.toNumber(_.join('', _.tail(feeDiscount))) / 100)
```
hellypi: I think this can be simplified, you don't need to convert to number to use BigNumber
If feeDiscount is "-20" we can just do `BN(feeDiscount.slice(1)).div(100)`
When working with bignumber it is good to always use the library own math funcs to avoid imprecision from javascript and only convert to number at the end with .toNumber

lib/plugins/wallet/bitcoind/bitcoind.js
```diff
@@ -49,10 +49,45 @@ function balance (account, cryptoCode) {
...
+        if (_.isEqual(operation, '-')) {
+          const newFee = estimatedFee.minus(percentValue)
+          if (newFee.lt(0.00001)) return
+          return newFee
+        }
+        if (_.isEqual(operation, '+')) {
+          const newFee = estimatedFee.add(percentValue)
+          if (newFee.gt(0.1)) return
+          return newFee
+        }
```
hellypi: I think this is just personal preference but I like to use ternary operators in stuff like this

for the outer ifs:
`const newFee = _.isEqual(operation, '-') ? estimatedFee.minus(percentValue) : estimatedFee.add(percentValue)`
for the inner ifs:
`return newFee.gt(0.00001) || newFee.lt(0.1) ? newFee : null`

hellypi (changes requested):

lib/plugins/wallet/bitcoind/bitcoind.js
josepfo: Looks cleaner for sure, although I wanted to do it this way just to be extra safe that the first character is '+' and we don't add erroneous fees.

lib/plugins/wallet/bitcoind/bitcoind.js
josepfo: I don't think there's a reason to sound the alarm as the tx will proceed normally. I'll ask Rafael

lib/plugins/wallet/bitcoind/bitcoind.js
josepfo: I've rewritten most of the code, please take a look.

lib/wallet.js
```diff
@@ -54,6 +55,17 @@ function _balance (settings, cryptoCode) {
...
+        return loadLatestConfig()
+          .then(config => {
+            const feeDiscount = config.wallets_BTC_feeDiscount
+            return r.wallet.sendCoins(r.account, toAddress, cryptoAtoms, cryptoCode, feeDiscount)
+              .then(res => {
+                mem.clear(module.exports.balance)
+                return res
+              })
+          })
+      }
```
hellypi: Can you check if this works? I pulled the .then outside
```suggestion
      if (cryptoCode === 'BTC') {
        return loadLatestConfig()
          .then(config => {
            const feeDiscount = config.wallets_BTC_feeDiscount
            return r.wallet.sendCoins(r.account, toAddress, cryptoAtoms, cryptoCode, feeDiscount)
          })
          .then(res => {
            mem.clear(module.exports.balance)
            return res
          })
      }
```

migrations/1620954224627-add-fee-priority.js
```diff
@@ -0,0 +1,19 @@
...
+    wallets_BTC_feeDiscount: '1'
+  }
+  return loadLatest()
+    .then(config => {
+      return saveConfig(newConfig)
+        .then(next)
+        .catch(err => {
+          return next(err)
+        })
+    })
```
hellypi: same here, each then() can return a value that will be used as the resolved value for the next then.
So instead of
```JS
return somepromise()
  .then(response => {
    return newpromise().then( innerResponse => {
      ...
    })
} )
```
we should have
```JS
return somepromise()
  .then(response => {
    return newpromise()
  })
  .then(innerResponse => {
    ...
  })
```

hellypi (changes requested):

hellypi:
has conflicts

hellypi (changes requested):
unlucky, I think recent merges caused conflicts again

migrations/1620954224627-add-fee-priority.js
```diff
@@ -0,0 +1,19 @@
+const { saveConfig, loadLatest } = require('../lib/new-settings-loader')
+
+exports.up = function (next) {
+  const newConfig = {
+    wallets_BTC_feeDiscount: '1'
```
RafaelTaranto: Can we rename this to feeMultiplier? Also what happens when we need to add discount into another coin? 

It will require a new migration and a new coin in that big if, let's generalize this. It should add the fee multiplier to every configured coin and automatically set that to 1 every time the operator adds a new coin.

lib/wallet.js
```diff
@@ -58,6 +59,17 @@ function _balance (settings, cryptoCode) {
 function sendCoins (settings, tx) {
   return fetchWallet(settings, tx.cryptoCode)
     .then(r => {
+      if (tx.cryptoCode === 'BTC') {
```
RafaelTaranto: This can be generic, we just get the feeDiscount for the specific coin

RafaelTaranto (changes requested):

migrations/1620954224627-add-fee-priority.js
josepfo: Hmm, didn't know that this would work for other coins too. The card is very specific about BTC. But I think I understand what you are saying, if in the future we want to add this feat to another coin, the config is already ready for it.

new-lamassu-admin/src/pages/Wallet/Wallet.js
```diff
@@ -140,6 +185,33 @@ const Wallet = ({ name: SCREEN_KEY }) => {
...
+          <DialogActions className={classes.actions}>
+            <Button
+              onClick={() =>
+                saveFeeDiscount({ BTC_feeMultiplier: selectedDiscount })
+              }>
+              Confirm
+            </Button>
+          </DialogActions>
+        </Modal>
+      )}
```
chaotixkilla, line 214: Is this only specific to BTC? If it is, ignore this comment

chaotixkilla (approved):

lib/wallet.js
```diff
@@ -58,7 +59,11 @@ function _balance (settings, cryptoCode) {
 function sendCoins (settings, tx) {
   return fetchWallet(settings, tx.cryptoCode)
     .then(r => {
-      return r.wallet.sendCoins(r.account, tx, settings, r.operatorId)
+      return loadLatestConfig()
```
RafaelTaranto: Is `settings` different than loadLatestConfig?

RafaelTaranto (changes requested):

lib/plugins/wallet/bitcoind/bitcoind.js
```diff
@@ -50,11 +50,32 @@ function balance (account, cryptoCode, settings, operatorId) {
...
+function estimateFee () {
+  return fetch('estimatesmartfee', [6, 'unset'])
+    .then(result => BN(result.feerate))
+    .catch(() => {})
+}
+
+function calculateFeeDiscount (feeMultiplier) {
+  // 0 makes bitcoind do automatic fee selection
+  const AUTOMATIC_FEE = 0
+  if (!feeMultiplier) return AUTOMATIC_FEE
```
RafaelTaranto: Shouldn't we default to automatic_fee if the multiplier is 1 as well?

RafaelTaranto (changes requested):

lib/plugins/wallet/bitcoind/bitcoind.js
josepfo: Hmm, that is a good point. Probably wouldn't make much of a difference since we use do  `estimatedFee.times(feeMultiplier)`, and this estimate should be the same similar to what bitcoind does automatically under the hood (at least, that's what I've tried to do). But anyway, we better force an automatic fee calculation from the bitcoind, right?

lib/wallet.js
josepfo: Can't say for sure, I'll verify this

RafaelTaranto (approved):

PR #724, opened 2021-05-20
Notify cashbox removal

chaotixkilla:


chaotixkilla:
[Notify operator on cashbox removed](https://trello.com/c/fqyEXyFB/765-notify-operator-on-cashbox-removed)

migrations/1621430588944-notify-cashbox-removal.js
```diff
@@ -0,0 +1,35 @@
...
+      if(config.notifications_notificationCenter_active) {
+        newConfig.notifications_notificationCenter_security = true
+      }
+
+      return saveConfig(newConfig)
+        .then(() => db.multi(sql, next))
+        .catch(err => {
+          if (err.message === 'lamassu-server is not configured') {
+            return next()
+          }
```
josepfo: This condition only makes sense when migrating something from v7.4. I did the same mistake hehe
```suggestion
```

josepfo (changes requested):

migrations/1621430588944-notify-cashbox-removal.js
```diff
@@ -0,0 +1,32 @@
...
+      if(config.notifications_email_active) {
+        newConfig.notifications_email_security = true
+      }
+      if(config.notifications_sms_active) {
+        newConfig.notifications_sms_security = true
+      }
+      if(config.notifications_notificationCenter_active) {
+        newConfig.notifications_notificationCenter_security = true
+      }
+
```
josepfo, line 21: Don't we want to add these fields when the value is false as well?

RafaelTaranto (approved):

lib/notifier/index.js
```diff
@@ -246,6 +246,41 @@ function sendTransactionMessage (rec, isHighValueTx) {
...
+
+      const emailActive =
+        notifications.email.active &&
+        notifications.email.security
+
+      const smsActive =
+        notifications.sms.active &&
+        notifications.sms.security
+      
+      if (emailActive) promises.push(emailFuncs.sendMessage(settings, rec))
```
joshmh, line 276: This notification logic should be general purpose for all notifications. @chaotixkilla can you create a new Trello card to generalize the logic?

PR #725, opened 2021-05-21
Monero support

chaotixkilla:


chaotixkilla:
[Monero - Add support](https://trello.com/c/LnyIVlBL/716-monero-add-support)

lib/plugins/common/json-rpc.js
```diff
@@ -3,8 +3,9 @@ const axios = require('axios')
 const uuid = require('uuid')
 const fs = require('fs')
 const _ = require('lodash/fp')
+const request = require('request-promise')
```
josepfo, line 6: I think this module is deprecated and we should look for alternatives, can you please check this? btw, can't we use axios for this as well?

lib/blockchain/common.js
```diff
@@ -65,11 +70,24 @@ function es (cmd) {
...
 
+  if (!_.isNil(coinRec.wallet)) {
+    const supervisorConfigWallet = `[program:${blockchain}-wallet]
+command=nice ${walletCmd}
+autostart=true
+autorestart=true
+stderr_logfile=/var/log/supervisor/${blockchain}-wallet.err.log
+stdout_logfile=/var/log/supervisor/${blockchain}-wallet.out.log
+environment=HOME="/root"
+`
```
josepfo: Seems good but I think we could move this config generation inside a function, it's pretty much the same string as the regular supervisor config. What do you think?

lib/blockchain/common.js
```diff
@@ -97,7 +118,9 @@ function fetchAndInstall (coinRec) {
   const binDir = binaries.dir
 
   es(`wget -q ${url}`)
-  es(`tar -xzf ${downloadFile}`)
+  coinRec.cryptoCode === 'XMR'
+    ? es(`tar -xf ${downloadFile}`)
+    : es(`tar -xzf ${downloadFile}`)
```
josepfo: It doesn't feel right to have this XMR condition inside a 'common' file. The alternative here is to check whether we have a bzip2 or gzip and use the appropriate option. 

```suggestion
es(`file ${downloadFile} | awk '{print $2}'`) === 'bzip2'
  ? es(`tar -xf ${downloadFile}`)
  :  es(`tar -xzf ${downloadFile}`)
```

josepfo: Probably it's even worth to have the bzip2 and gzip as a key-value inside the BINARIES object

lib/plugins/wallet/monerod/monerod.js
```diff
@@ -0,0 +1,196 @@
...
+const configPath = utils.configPath(cryptoRec, blockchainUtils.blockchainDir())
+const walletDir = path.resolve(utils.cryptoDir(cryptoRec, blockchainUtils.blockchainDir()), 'wallets')
+const unitScale = cryptoRec.unitScale
+const config = jsonRpc.parseConf(configPath)
+
+const rpcConfig = {
+  username: config['rpc-login'].split(':')[0],
+  password: config['rpc-login'].split(':')[1],
+  port: cryptoRec.walletPort || cryptoRec.defaultPort
+}
```
josepfo: I remember fixing an issue where the 'parseConf' break the server in case of a not configured wallet. Probably you should follow the same pattern as the other coins and wrap this inside a function alongside a try-catch.

josepfo (changes requested):

lib/plugins/common/json-rpc.js
chaotixkilla: This is related to what we spoke during the call, going to mark it as resolved

lib/blockchain/common.js
chaotixkilla: Had to make some slight changes as that command responds with stuff like `Zip` for `.zip`, `Debian` for `.deb`, etc.

Used `echo ${downloadFile} | awk -F. '{print $NF}'`, which will retrieve `gz` or `bz2`, depending on the file. The key/value would be great if we had more than 2 different file extensions being used. As we currently only have XMR using a different extension, I'd believe `.gz` is the most common one

lib/plugins/wallet/monerod/monerod.js
```diff
@@ -0,0 +1,202 @@
...
+
+function fetch (method, params) {
+  return jsonRpc.fetchDigest(rpcConfig(), method, params)
+}
+
+function handleError (error) {
+  switch(error.code) {
+    case -13:
+      {
+        if (fs.existsSync(path.resolve(walletDir, 'Wallet')) && fs.existsSync(path.resolve(walletDir, 'Wallet.keys')) && fs.existsSync(path.resolve(walletDir, 'Wallet.address.txt'))) {
```
josepfo: this line seems too long, probably better to break it down

josepfo (changes requested):
just a small detail. Looks noice, good job!

josepfo (approved):

lib/blockchain/common.js
```diff
@@ -65,25 +70,36 @@ function es (cmd) {
...
+  const supervisorConfig = generateSupervisorConfig(blockchain, cmd)
   writeFile(`/etc/supervisor/conf.d/${coinRec.code}.conf`, supervisorConfig)
 }
 
 function isInstalledSoftware (coinRec) {
-  return fs.existsSync(`/etc/supervisor/conf.d/${coinRec.code}.conf`)
+  return !_.isNil(coinRec.wallet)
+    ? fs.existsSync(`/etc/supervisor/conf.d/${coinRec.code}.conf`)
+      && fs.existsSync(`/etc/supervisor/conf.d/${coinRec.code}-wallet.conf`)
+    : fs.existsSync(`/etc/supervisor/conf.d/${coinRec.code}.conf`)
```
RafaelTaranto:
```suggestion
  const nodeInstalled = fs.existsSync(`/etc/supervisor/conf.d/${coinRec.code}.conf`)
  const walletInstalled = _.isNil(coinRec.wallet) ? true : fs.existsSync(`/etc/supervisor/conf.d/${coinRec.code}-wallet.conf`)
  return nodeInstalled && walletInstalled
```

lib/plugins/wallet/monerod/monerod.js
```diff
@@ -0,0 +1,206 @@
...
+const jsonRpc = require('../../common/json-rpc')
+
+const { utils } = require('lamassu-coins')
+
+const blockchainUtils = require('../../../coin-utils')
+const BN = require('../../../bn')
+const E = require('../../../error')
+const { logger } = require('../../../blockchain/common')
+
+const cryptoRec = utils.getCryptoCurrency('XMR')
```
RafaelTaranto: Don't we have XMR as a constant on lamassu-coins? A bit nitpicky but it helps to remove strings from the code.

lib/plugins/wallet/monerod/monerod.js
```diff
@@ -0,0 +1,206 @@
...
+}
+
+function balance (account, cryptoCode) {
+  return accountBalance(cryptoCode)
+}
+
+function sendCoins (account, address, cryptoAtoms, cryptoCode) {
+  return checkCryptoCode(cryptoCode)
+    .then(() => refreshWallet())
+    .then(() => fetch('transfer_split', {
```
RafaelTaranto, line 103: is there a reason to use `transfer_split` instead of `transfer`

RafaelTaranto (changes requested):

lib/plugins/wallet/monerod/monerod.js
chaotixkilla: From what I read, both endpoints do the same, but `transfer_split` splits a transaction into multiple smaller ones, if needed. I believe `transfer_split` is similar to a `transfer` 2.0, and act the same on regular, single-destination transactions and has improvements on multi-destination transactions, which we could implement in the future.
So yeah, for those reasons, I chose `transfer_split`

PR #727, opened 2021-05-24
BTC transaction batching

chaotixkilla:


chaotixkilla:
[BTC batching transactions](https://trello.com/c/9m7k0p9h/764-btc-batching-transactions)

lib/wallet.js
```diff
@@ -202,6 +220,11 @@ function isStrictAddress (settings, cryptoCode, toAddress) {
     })
 }
 
+function supportsBatching (settings, cryptoCode) {
+  return fetchWallet(settings, cryptoCode)
+    .then(r => r.wallet.supportsBatching(settings, cryptoCode))
+}
+
```
josepfo: I noticed that every coin has a supportsBatching() that uses checkCryptoCode() and checks if sendCoinsBatch() exists, we can move that logic to this function instead of having a single implementation on each coin.

josepfo (changes requested):

lib/wallet.js
chaotixkilla: I believe I had exactly like that, but the mock-wallet does not fit with that interface, so I had to change it to this

lib/plugins.js
```diff
@@ -251,15 +261,27 @@ function plugins (settings, deviceId) {
...
+          coins: _.zipWith(_.assign, _.zipWith(_.assign, coinsWithoutRate, tickers), batchableCoins),
           configVersion,
           areThereAvailablePromoCodes
         }
       })
   }
 
   function sendCoins (tx) {
-    return wallet.sendCoins(settings, tx)
+    return wallet.supportsBatching(settings, tx.cryptoCode)
```
josepfo, line 285: supportsBatching doesn't need this accounts settings

josepfo: also, you're still using a function for each coin.
```suggestion
    return wallet.checkCryptoCode(tx.cryptoCode)
          .then(() => wallet.SUPPORTS_BATCHING)
```
but if you prefer the way it is now, just leave it, it's ok :)

josepfo (changes requested):

lib/plugins.js
chaotixkilla: Hmm, the generic wallet `supportsBatching` method needs the `settings` parameter, as it calls `fetchWallet(settings, cryptoCode)` before calling the specific wallet `supportsBatching` function. This last one might not need the `settings` parameter, at least at this point in time, I agree, as only bitcoind is having batching enabled. Once 3rd party services have batching, they might require the `settings` parameter.

About the 2nd comment, I decided to keep the individual wallet's `supportsBatching` as a function, despite having the `SUPPORTS_BATCHING` constant, as the function still needs to confirm the cryptoCode with `checkCryptoCode`, as all functions from the individual wallets have. The end result will still be the `SUPPORTS_BATCHING` constant, it's just that it need to confirm the cryptocode first. BTW, I don't think the suggestion on the comment makes sense, as `checkCryptoCode` is a promise and making `Promise && Boolean` does nothing I believe, could you please elaborate?

josepfo: You're right the suggestion doesn't work. Given a second thought it's more readable as it is.

josepfo (approved):

lib/wallet.js
```diff
@@ -73,6 +73,24 @@ function sendCoins (settings, tx) {
...
+    .then(r => {
+      return r.wallet.sendCoinsBatch(r.account, txs, cryptoCode)
+        .then(res => {
+          mem.clear(module.exports.balance)
+          return res
+        })
+    })
+    .catch(err => {
+      if (err.name === INSUFFICIENT_FUNDS_NAME) {
+        throw httpError(INSUFFICIENT_FUNDS_NAME, INSUFFICIENT_FUNDS_CODE)
```
RafaelTaranto, line 93: I don't think this is useful anymore. Isn't this something that we need for the machine ui?

lib/tx-batching.js
```diff
@@ -0,0 +1,88 @@
...
+    .then(batches => {
+      _.each(batch => {
+        const elapsedMS = batch.time_elapsed * 1000
+
+        if (elapsedMS >= lifecycle) {
+          return closeTransactionBatch(batch)
+            .then(() => submitBatch(settings, batch))
+        }
+      }, batches)
+    })
```
RafaelTaranto: Shouldn't this be a atomic operation? We need a tx for this.

lib/tx-batching.js
```diff
@@ -0,0 +1,88 @@
...
+  return db.manyOrNone(sql, [_.map(pgp.as.text, statuses).join(',')])
+}
+
+function submitBatch (settings, batch) {
+  getBatchTransactions(batch)
+    .then(txs => {
+      wallet.sendCoinsBatch(settings, txs, batch.crypto_code)
+        .then(res => confirmSentBatch(batch, res))
+        .catch(err => setErroredBatch(batch, err.message))
+    })
```
RafaelTaranto: I think that it's safe enough if the parent function is already inside a transaction, but we need to make sure here as well. This should be part of the transaction in the parent

lib/tx-batching.js
```diff
@@ -0,0 +1,88 @@
...
+
+function processBatches (settings, lifecycle) {
+  getBatchesByStatus(['open', 'failed'])
+    .then(batches => {
+      _.each(batch => {
+        const elapsedMS = batch.time_elapsed * 1000
+
+        if (elapsedMS >= lifecycle) {
+          return closeTransactionBatch(batch)
+            .then(() => submitBatch(settings, batch))
```
RafaelTaranto: Probably better to do a map -> filter and make sure the "close transacitonBatch" and "submitBatch" functions accept multiple values. (do rename them to plural if you decide to do so.

RafaelTaranto (changes requested):

RafaelTaranto (approved):

PR #733, opened 2021-05-25
Fix missing prop on formik checkbox

chaotixkilla:


chaotixkilla:
[Settings is not defined on login page checkbox](https://trello.com/c/RRuaQEeS/789-settings-is-not-defined-on-login-page-checkbox)

new-lamassu-admin/src/components/inputs/formik/Checkbox.js
```diff
@@ -5,7 +5,21 @@ import { Checkbox } from '../base'
 const CheckboxInput = memo(({ label, textAlign, fullWidth, ...props }) => {
   const { name, onChange, value } = props.field
 
-  return <Checkbox name={name} onChange={onChange} value={value} {...props} />
+  const settings = {
+    enabled: props.enabled ?? true,
```
josepfo: Should we do
```suggestion
    enabled: props.enabled ?? false,
```
instead?

new-lamassu-admin/src/components/inputs/formik/Checkbox.js
```diff
@@ -5,7 +5,21 @@ import { Checkbox } from '../base'
...
+    label: label,
+    disabledMessage: props.disabledMessage ?? ''
+  }
```
josepfo: Since this is overriding the settings that comes with props, I think we need to consider all the fields here. It's missing at least the 'requirement' field.

josepfo (changes requested):

new-lamassu-admin/src/components/inputs/formik/Checkbox.js
josepfo: This is perfectly correct, ignore my suggestion!

new-lamassu-admin/src/components/inputs/formik/Checkbox.js
josepfo: This is correct as well. The requirement field is something that I've added to blockcypher but it's not needed in this Checkbox component.

josepfo (approved):

new-lamassu-admin/src/components/inputs/formik/Checkbox.js
```diff
@@ -5,7 +5,21 @@ import { Checkbox } from '../base'
 const CheckboxInput = memo(({ label, textAlign, fullWidth, ...props }) => {
```
RafaelTaranto:
```suggestion
const CheckboxInput = memo(({ label, textAlign, fullWidth, enabled = true, disableMessage = '', ...props }) => {
```

new-lamassu-admin/src/components/inputs/formik/Checkbox.js
```diff
@@ -5,7 +5,21 @@ import { Checkbox } from '../base'
...
   const { name, onChange, value } = props.field
 
-  return <Checkbox name={name} onChange={onChange} value={value} {...props} />
+  const settings = {
+    enabled: props.enabled ?? true,
+    label: label,
+    disabledMessage: props.disabledMessage ?? ''
```
RafaelTaranto:
```suggestion
    enabled,
    label,
    disabledMessage
```

RafaelTaranto (changes requested):

PR #735, opened 2021-05-26
feat: create new block storage volume and migrate data

josepfo:


josepfo:
[Migrate data to a object storage](https://trello.com/c/Zyv7FRv5/720-migrate-data-to-a-object-storage)

digitalocean/createblockstorage.sh
```diff
@@ -0,0 +1,80 @@
...
+echo
+echo "New block storage volume successfully installed!"
+echo "Can be accessed at: /mnt/lamassu-data"
+echo
+
+
+
+
+
+
```
chaotixkilla: There's a couple of extra lines at the end of file

digitalocean/migratedata.sh
```diff
@@ -0,0 +1,134 @@
...
+
+
+
+
+
+
+
+
+
+
```
chaotixkilla: Same here

chaotixkilla (changes requested):
Just some unneeded extra lines at the EOF, otherwise, great job!

chaotixkilla (approved):

digitalocean/migratedata.sh
```diff
@@ -0,0 +1,124 @@
...
+ID_PHOTO_CARD_DIR=/opt/lamassu-server/idphotocard
+FRONTCAMERA_DIR=/opt/lamassu-server/frontcamera
+
+mkdir -p $NEW_VOLUME_MOUNTING_POINT$CERT_DIR
+mkdir -p $NEW_VOLUME_MOUNTING_POINT$KEY_DIR
+mkdir -p $NEW_VOLUME_MOUNTING_POINT$MNEMONIC_DIR
+mkdir -p $NEW_VOLUME_MOUNTING_POINT/var/backups
+mkdir -p $NEW_VOLUME_MOUNTING_POINT/mnt
+mkdir -p $NEW_VOLUME_MOUNTING_POINT/var/lamassu
+mkdir -p $NEW_VOLUME_MOUNTING_POINT/opt/lamassu-server
```
RafaelTaranto: Probably better to organize this instead of prepending the folder everywhere.

stuff like:

`$NEW_VOLUME_MOUNTING_POINT/lamassu-server` instead of `$NEW_VOLUME_MOUNTING_POINT/opt/lamassu-server`

RafaelTaranto (changes requested):

digitalocean/migratedata.sh
josepfo: You mean move everything to $NEW_VOLUME_MOUNTING_POINT/lamassu-server, won't that be a little confusing to know what is what?

digitalocean/createblockstorage.sh
```diff
@@ -0,0 +1,74 @@
...
+sudo snap install doctl
+
+#Create config folder as suggested in doctl documentation
+
+sudo mkdir -p ~/.config
+
+#Access doctl account
+
+echo "Accessing digital ocean account..."
+doctl auth init --context data-migration
```
josepfo: Btw, this cmd will prompt the user to enter the token. As we spoke before this might be an issue.

I tried to use pipe to automatically do this:

```suggestion
echo $TOKEN | doctl auth init --context data-migration
```

but to no avail. Any suggestions?

RafaelTaranto: Hmm, is the text ok? To pass through we would have to find a non interactive mode flag for this command.

digitalocean/migratedata.sh
RafaelTaranto: Oh, we can still keep different folders. Something like:

/newvolume/lamassu/idphotocard
/newvolume/lamassu/ofac
/newvolume/backups
/newvolume/mnemonic
/newvolume/ssl/certs
/newvolume/ssl/private

RafaelTaranto (changes requested):

digitalocean/createblockstorage.sh
josepfo: Yeah, I tried with --acess-token flag but for some reason it doesn't work alongside with --context flag. I'll remove the second one as I don't think it really matters for our purpose here.

RafaelTaranto (approved):

PR #744, opened 2021-06-01
fix: add missing dependecy

RafaelTaranto:


RafaelTaranto:
[Add missing dependency](https://trello.com/c/Ogcq0GyG/797-add-missing-dependency)

PR #750, opened 2021-06-09
Fix coin installation restarting previously killed supervisor processes

chaotixkilla:
fix: use supervisor reload instead of restart

fix: set blockchain process autorestart to unexpected

fix: use supervisor reload and update instead of restart

fix: stop restarting supervisor service

fix: set blockchain supervisor config autorestart


chaotixkilla:
[Installing a new coin with lamassu-coins restarts previously stopped supervisor processes](https://trello.com/c/2pVxymcg/814-installing-a-new-coin-with-lamassu-coins-restarts-previously-stopped-supervisor-processes)

josepfo (approved):

PR #752, opened 2021-06-10
Compliance notifications through sms and email

chaotixkilla:


chaotixkilla:
[Compliance notifications only works for notification center](https://trello.com/c/NAk8OPJe/816-compliance-notifications-only-works-for-notification-center)

hellypi (approved):

lib/notifier/index.js
```diff
@@ -166,6 +166,51 @@ function transactionNotify (tx, rec) {
...
+function complianceNotify (customer, deviceId, ...args) {
+  return Promise.all([
+    settingsLoader.loadLatest(),
+    queries.getMachineName(deviceId)
+  ])
+    .then(([settings, machineName]) => {
+      const notifications = configManager.getGlobalNotifications(settings.config)
+
+      const action = args[0]
+      const period = args[1]
```
RafaelTaranto: better to just name the params instead of getting from args

RafaelTaranto (changes requested):

lib/notifier/index.js
chaotixkilla: Hmm, I don't think I can name the params specifically, as they come from a spread operator. I can deconstruct the array though, instead of selecting elements one by one

RafaelTaranto: Hmm, the reason it works is because of the rest parameter.

```js
// these functions are equivalent
(a, b, c) => {
  return a + b + c
}

(a, ...args) => {
  return a + args[0] + args[1]
}

```

lib/notifier/index.js
```diff
@@ -166,6 +166,50 @@ function transactionNotify (tx, rec) {
...
+        notifications.email.compliance
+
+      const smsActive =
+        notifications.sms.active &&
+        notifications.sms.compliance
+
+      if (emailActive) promises.push(emailFuncs.sendMessage(settings, rec))
+      if (smsActive) promises.push(smsFuncs.sendMessage(settings, rec))
+
+      notifyIfActive('compliance', 'customerComplianceNotify', customer, deviceId, ...args)
```
RafaelTaranto: Also it's better to just pass what's needed here, this rest param stuff can get complicated pretty fast

RafaelTaranto (changes requested):

RafaelTaranto (approved):

PR #759, opened 2021-06-14
Fix customer actions using wrong cookie

chaotixkilla:


chaotixkilla:
[Customer actions break when there's a 'token' cookie present](https://trello.com/c/ua0xDpAr/827-customer-actions-break-when-theres-a-token-cookie-present)

josepfo (approved):

RafaelTaranto (approved):

PR #762, opened 2021-06-16
Fix/design fixes on blacklist screen

josepfo:


josepfo:
[Design fixes on blacklist screen](https://trello.com/c/V2NcndKR/831-design-fixes-on-blacklist-screen)

new-lamassu-admin/src/pages/Blacklist/BlacklistModal.js
```diff
@@ -74,9 +75,11 @@ const BlackListModal = ({
         </Form>
       </Formik>
       <div className={classes.footer}>
-        <Link type="submit" form="address-form">
-          Blacklist address
-        </Link>
+        <Box display="flex" justifyContent="flex-end">
+          <Link type="submit" form="address-form">
```
chaotixkilla, line 79: I'm not really a fan of inlining styles as props. This is an inconsistency we have thoughout some components in our code base. IMO, we should always keep styles contained in styling files

chaotixkilla (approved):

new-lamassu-admin/src/pages/Blacklist/Blacklist.js
```diff
@@ -120,8 +121,20 @@ const Blacklist = () => {
...
   }
 
+  const validateAddress = (cryptoCode, address) => {
+    try {
+      return !R.isNil(coinUtils.parseUrl(cryptoCode, 'main', address))
+    } catch {
+      return false
+    }
+  }
+
```
RafaelTaranto, line 131: I think you can add this as a test on the yup schema, there's no need for it to be outside the modal.

new-lamassu-admin/src/pages/Blacklist/BlacklistModal.js
RafaelTaranto: Ohh, this box thing is made to simplify flexbox behavior. This is not really inline styling, just props being passed to the component.

RafaelTaranto (changes requested):

RafaelTaranto (approved):

PR #763, opened 2021-06-17
Fix: design fixes on notifications screen

josepfo:


josepfo:
[Design fixes on notifications screen](https://trello.com/c/JacvJ4OV/832-design-fixes-on-notifications-screen)

new-lamassu-admin/src/components/inputs/cashbox/Cashbox.js
```diff
@@ -21,7 +21,7 @@ const Cashbox = ({
   labelClassName,
   inFiatBalanceAlerts
 }) => {
-  const classes = cashboxClasses({ percent, cashOut })
+  const classes = cashboxClasses({ percent, cashOut, inFiatBalanceAlerts })
```
chaotixkilla: I'm not really a fan of having a flag for a specific screen, as we might use that same behavior in the future, and the variable name shouldn't be tied to a specific screen. My suggestion is to call this flag something like `shouldApplyColorVariance`/`shouldApplyColorGradient` (I don't like the gradient one as much, as we're not working with gradients)

new-lamassu-admin/src/components/inputs/cashbox/Cashbox.styles.js
```diff
@@ -11,8 +11,12 @@ const colors = {
   }
 }
 
-const colorPicker = ({ percent, cashOut }) =>
-  colors[cashOut ? 'cashOut' : 'cashIn'][percent >= 50 ? 'full' : 'empty']
+const colorPicker = ({ percent, cashOut, inFiatBalanceAlerts }) => {
```
chaotixkilla: This flag

new-lamassu-admin/src/components/inputs/cashbox/Cashbox.js
chaotixkilla: Also, default value should be `false` when this flag is not specified in props. You can make the default value `true`, but the flag would need to be called something as `dontApplyColorVariance`

new-lamassu-admin/src/pages/Notifications/sections/FiatBalanceAlerts.js
```diff
@@ -58,29 +58,57 @@ const FiatBalance = ({ section, min = 0, max = 100, fieldWidth = 80 }) => {
...
+            <div className={classes.first}>
+              <div className={classes.row}>
+                <Cashbox
+                  labelClassName={classes.cashboxLabel}
+                  emptyPartClassName={classes.cashboxEmptyPart}
+                  percent={
+                    values.fillingPercentageCassette1 ??
+                    data?.fillingPercentageCassette1
+                  }
+                  inFiatBalanceAlerts={true}
```
chaotixkilla: This here. Also, I believe you can get away with omitting `={true}`, and just call `applyColorVariant` or whatever name you decide better fits the flag's purpose

new-lamassu-admin/src/components/inputs/cashbox/Cashbox.js
chaotixkilla: Ah, noticed you're calling a pre-existing flag here. I guess you can ignore the variable name change

chaotixkilla (changes requested):

RafaelTaranto (approved):

PR #764, opened 2021-06-18
Analytics screen

chaotixkilla:


chaotixkilla:
[Analytics PT1](https://trello.com/c/3IyFGuPy/828-analytics-pt1)

new-lamassu-admin/src/pages/Analytics/Analytics.js
```diff
@@ -0,0 +1,398 @@
...
+  mainCurrency
+}) => {
+  const classes = useStyles()
+
+  const isCurrency = !!currencies
+  // const mostUsedCurrency = R.reduce(
+  //   (acc, v) => (acc === '' ? v : currencies[v] > currencies[acc] ? v : acc),
+  //   '',
+  //   R.keys(currencies)
+  // )
```
josepfo: Did you miss this code or intend to use it in the future?
```suggestion
```

josepfo (changes requested):
Great job, looks awesome!
Just a few details, according to the spec the percentage on each OverviewEntry should be bold and without empty space between the icon. Also, the background on the Select drop-down should have a dark color.

josepfo (approved):

new-lamassu-admin/src/pages/Analytics/Analytics.js
```diff
@@ -0,0 +1,387 @@
...
+import { fromNamespace } from 'src/utils/config'
+
+import styles from './Analytics.styles'
+import Graph from './Graph'
+
+const useStyles = makeStyles(styles)
+
+const DAY = 24 * 60 * 60 * 1000
+const WEEK = 7 * 24 * 60 * 60 * 1000
+const MONTH = 30 * 24 * 60 * 60 * 1000
```
RafaelTaranto: Don't we use this elswhere?

new-lamassu-admin/src/pages/Analytics/Graph.js
```diff
@@ -0,0 +1,403 @@
...
+  java,
+  neon,
+  subheaderDarkColor,
+  offColor,
+  fontColor,
+  primaryColor,
+  fontSecondary
+} from 'src/styling/variables'
+
+const MINUTE = 60 * 1000
```
RafaelTaranto: We do hehe, add this to a consts file on utils.

new-lamassu-admin/src/pages/Analytics/Graph.js
```diff
@@ -0,0 +1,403 @@
...
+const DAY = 24 * 60 * 60 * 1000
+const WEEK = 7 * 24 * 60 * 60 * 1000
+const MONTH = 30 * 24 * 60 * 60 * 1000
+
+const Graph = ({ data, representing, period, timezone }) => {
+  const ref = useRef(null)
+
+  const GRAPH_HEIGHT = 401
+  const GRAPH_WIDTH = 1163
+  const GRAPH_MARGIN = useMemo(
```
RafaelTaranto, line 22: Is there a reason for this to be a memoized react component? Why not a simple object?

new-lamassu-admin/src/pages/Analytics/Graph.js
```diff
@@ -0,0 +1,403 @@
...
+      'Apr',
+      'May',
+      'Jun',
+      'Jul',
+      'Aug',
+      'Sep',
+      'Oct',
+      'Nov',
+      'Dec'
+    ]
```
RafaelTaranto: I'm pretty sure we can get this from a lib, is the react project using any date libs?

new-lamassu-admin/src/pages/Analytics/Analytics.js
```diff
@@ -0,0 +1,387 @@
...
+          d => new Date(d.created).getTime() >= Date.now() - MONTH
+        ) ?? [],
+      previous:
+        machineTxs.filter(
+          d =>
+            new Date(d.created).getTime() < Date.now() - MONTH &&
+            new Date(d.created).getTime() >= Date.now() - 2 * MONTH
+        ) ?? []
+    }
+  }
```
RafaelTaranto: This would also get pretty simplified with a lib.

RafaelTaranto (changes requested):
Also has conflicts

new-lamassu-admin/src/pages/Analytics/Graph.js
chaotixkilla: Mostly for optimization purposes, as the `useEffect` hook is calling a function, which "needs" to be wrapped in `useCallback`, which calls these objects, which should be memoized ideally I think. Might not matter as much to require optimization

PR #765, opened 2021-06-20
Fix reject address reuse

chaotixkilla:
feat: migrate old tx addresses to blacklist table
feat: add new addresses to blacklist list

### DIFF
--- a/lib/blacklist.js
+++ b/lib/blacklist.js
@@ -3,11 +3,10 @@ const notifierQueries = require('./notifier/queries')
 
 // Get all blacklist rows from the DB "blacklist" table that were manually inserted by the operator
 const getBlacklist = () => {
-  return db.any(`SELECT * FROM blacklist WHERE created_by_operator = 't'`).then(res =>
+  return db.any(`SELECT * FROM blacklist`).then(res =>
     res.map(item => ({
       cryptoCode: item.crypto_code,
-      address: item.address,
-      createdByOperator: item.created_by_operator
+      address: item.address
     }))
   )
 }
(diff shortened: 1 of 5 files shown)


chaotixkilla:
[Rework the "reject address reuse" functionality](https://trello.com/c/gYJ1FyjY/836-rework-the-reject-address-reuse-functionality)

lib/blacklist.js
```diff
@@ -3,11 +3,10 @@ const notifierQueries = require('./notifier/queries')
 
 // Get all blacklist rows from the DB "blacklist" table that were manually inserted by the operator
 const getBlacklist = () => {
-  return db.any(`SELECT * FROM blacklist WHERE created_by_operator = 't'`).then(res =>
+  return db.any(`SELECT * FROM blacklist`).then(res =>
     res.map(item => ({
       cryptoCode: item.crypto_code,
-      address: item.address,
-      createdByOperator: item.created_by_operator
```
RafaelTaranto, line 10: I'm pretty sure the blacklist type will still have the definition of `createdByOperator`.

RafaelTaranto (changes requested):

lib/cash-in/cash-in-tx.js
```diff
@@ -133,20 +133,18 @@ function postProcess (r, pi, isBlacklisted, addressReuse) {
...
       return logAction(sendRec, r.tx)
     })
 }
 
+function getAddresses () {
+  const sql = `SELECT DISTINCT to_address FROM cash_in_txs`
+
+  return db.any(sql).then(res => res.map(it => it.to_address))
+}
+
```
RafaelTaranto: This will run on every tx update, take a look at how the `checkForBlacklisted` function works and do the same here. Also it might be better to return a boolean from the sql instead of getting this huge list to go through in js.

RafaelTaranto (changes requested):

lib/blacklist.js
chaotixkilla: Correct. Altering the blacklist type should solve this. ATM I'm just removing the createdByOperator without migrating older entries with this field enabled, is this a good approach?

RafaelTaranto (approved):

PR #766, opened 2021-06-21
feat: advanced wallet settings

josepfo:


josepfo:
[Operator choice of machine-displayed crypto units](https://trello.com/c/ytcgzrUz/833-operator-choice-of-machine-displayed-crypto-units)

lib/commission-math.js
```diff
@@ -3,10 +3,10 @@ const configManager = require('./new-config-manager')
 const { utils: coinUtils } = require('lamassu-coins')
 
 function truncateCrypto (cryptoAtoms, cryptoCode) {
-  const DECIMAL_PLACES = 3
+  const DECIMAL_PLACES = 6
```
chaotixkilla, line 6: Doesn't this change affect line 13? I'm not sure, but I'm not understanding how the changes on line 9 compensate this one

new-lamassu-admin/src/pages/Wallet/AdvancedWallet.js
```diff
@@ -0,0 +1,63 @@
...
+import gql from 'graphql-tag'
+import * as R from 'ramda'
+import React from 'react'
+
+import { NamespacedTable as EditableTable } from 'src/components/editableTable'
+import { fromNamespace, toNamespace, namespaces } from 'src/utils/config'
+
+import { AdvancedWalletSchema, getAdvancedWalletElements } from './helper'
+
+const { utils: coinUtils } = require('lamassu-coins')
```
chaotixkilla: Shouldn't this import use ES6 import keyword? Surprised the linter hasn't thrown an error

```suggestion
import { utils as coinUtils } from 'lamassu-coins'
```

new-lamassu-admin/src/pages/Wallet/AdvancedWallet.js
```diff
@@ -0,0 +1,63 @@
...
+        namespaces={R.map(R.path(['code']))(cryptoCurrencies)}
+        data={config}
+        error={error?.message}
+        enableEdit
+        editWidth={174}
+        save={save}
+        validationSchema={AdvancedWalletSchema}
+        elements={getAdvancedWalletElements(cryptoCurrencies, coinUtils)}
+      />
+    </>
```
chaotixkilla: You can remove lines 47 and 59, as the fragment is only encaspulating a single element

new-lamassu-admin/src/pages/Wallet/helper.js
```diff
@@ -20,6 +20,48 @@ const WalletSchema = Yup.object().shape({
...
+    R.compose(
+      R.prop(['display']),
+      R.find(R.propEq('code', it))
+    )(cryptoCurrencies)
+
+  const getOptions = R.curry((coinUtils, it) => {
+    const options = R.keys(coinUtils.getCryptoCurrency(it.id).units)
+    return R.map(option => {
+      return { code: option, display: option }
+    })(options)
```
chaotixkilla, line 46:
```suggestion
    return R.map(option => ({ code: option, display: option }))(options)
```

chaotixkilla (changes requested):

new-lamassu-admin/src/pages/Wallet/AdvancedWallet.js
josepfo: good point, don't know what happened

lib/commission-math.js
josepfo: I got seriously confused when I saw that we use display units to compute crypto values hehe. About your comment: the difference between unitScale and DECIMAL_PLACES is the same as before, so this doesn't change anything in practice, but we have to do it because we now have multiple display scales and they should be used for display purposes only.

new-lamassu-admin/src/pages/Wallet/helper.js
josepfo: hmm, eslint formatted this

migrations/1623975493095-add-crypto-units-to-config.js
```diff
@@ -0,0 +1,22 @@
+const { saveConfig } = require('../lib/new-settings-loader')
+
+exports.up = function (next) {
+  const cryptoUnits = {
+    wallets_BTC_cryptoUnits: 'mili',
+    wallets_ETH_cryptoUnits: 'mili',
+    wallets_LTC_cryptoUnits: 'mili',
+    wallets_ZEC_cryptoUnits: 'mili',
+    wallets_BCH_cryptoUnits: 'mili'
```
RafaelTaranto: We only want to add this if the wallet is configured. Also dash is missing

new-lamassu-admin/src/pages/Wallet/Wallet.js
```diff
@@ -99,47 +103,60 @@ const Wallet = ({ name: SCREEN_KEY }) => {
...
+            <Wizard
+              coin={R.find(R.propEq('code', wizard))(cryptoCurrencies)}
+              onClose={() => setWizard(false)}
+              save={save}
+              error={error?.message}
+              cryptoCurrencies={cryptoCurrencies}
+              userAccounts={data?.config?.accounts}
+              accounts={accounts}
+              accountsConfig={accountsConfig}
+            />
```
RafaelTaranto, line 141: We don't want to create a migration every time we add a coin. We either add the option of display units on the wizard, or grab the first available unit on a coin and save automatically on first config.

RafaelTaranto (changes requested):

migrations/1623975493095-add-crypto-units-to-config.js
josepfo: Hmm, I thought we wanted to force this default value for all wallets. otherwise, we must force a default value throughout the codebase so we don't display undefined somewhere. Is it really worth it?

new-lamassu-admin/src/pages/Wallet/Wallet.js
josepfo: I think I'm not following, this code didn't change

migrations/1623975493095-add-crypto-units-to-config.js
```diff
@@ -0,0 +1,29 @@
...
+      const newSettings = {}
+      const activeCryptos = getAllCryptoCurrencies(settings.config)
+      if (_.head(activeCryptos)) {
+        _.map(crypto => {
+          const defaultUnit = _.head(_.keys(coinUtils.getCryptoCurrency(crypto).units))
+          newSettings[`wallets_${crypto}_cryptoUnits`] = defaultUnit
+          return newSettings
+        }, activeCryptos)
+      }
+      return saveConfig(newSettings)
```
RafaelTaranto:
```suggestion
      if (!activeCryptos.length) return Promise.resolve()
      
      _.map(crypto => {
        const defaultUnit = _.head(_.keys(coinUtils.getCryptoCurrency(crypto).units))
        newSettings[`wallets_${crypto}_cryptoUnits`] = defaultUnit
        return newSettings
      }, activeCryptos)
      
      return saveConfig(newSettings)
```

migrations/1623975493095-add-crypto-units-to-config.js
```diff
@@ -0,0 +1,29 @@
+const { saveConfig, loadLatest } = require('../lib/new-settings-loader')
+const { getAllCryptoCurrencies } = require('../lib/new-config-manager.js')
+const { utils: coinUtils } = require('lamassu-coins')
+const _ = require('lodash/fp')
+
+exports.up = function (next) {
+  loadLatest()
+    .then(settings => {
+      const newSettings = {}
+      const activeCryptos = getAllCryptoCurrencies(settings.config)
```
RafaelTaranto: Shouldn't we use: `getCryptosFromWalletNamespace` here?

RafaelTaranto (changes requested):

migrations/1623975493095-add-crypto-units-to-config.js
josepfo: Oh, that's a good catch. `getAllCryptoCurrencies()` returns all the coins that are configured and added in the locales section. I think `getCryptosFromWalletNamespace()` returns all the coins configured. Probably the second fits better here.

PR #769, opened 2021-06-21
Relocate cert-gen script

chaotixkilla:


chaotixkilla:
[Relocate cert-gen to dev](https://trello.com/c/vzracg2F/845-relocate-cert-gen-to-dev)

josepfo (approved):

shivchawla:
The cert script fails with openssl v1.1 (It can be easily fixed by removing empty fields on subject).
`  -subj "/C=IS/ST=/L=Reykjavik/O=Lamassu CA/CN=lamassu.is" \`
to
 ` -subj "/C=IS/L=Reykjavik/O=Lamassu CA/CN=lamassu.is" \`

May be something worth doing.

PR #770, opened 2021-06-22
fix: properly check old txs dispense confirmation

josepfo:


josepfo:
[Server double-counting dispensed notes](https://trello.com/c/Lh4PoQda/842-server-double-counting-dispensed-notes)

RafaelTaranto (approved):

PR #774, opened 2021-06-23
Fix Zero Conf Limit empty values to fail row validation check

chaotixkilla:


chaotixkilla:
[Zero Conf Limit makes the entire row striped when deleting the value](https://trello.com/c/2Nb0MLjk/851-zero-conf-limit-makes-the-entire-row-striped-when-deleting-the-value)

RafaelTaranto (approved):

new-lamassu-admin/src/pages/Wallet/helper.js
```diff
@@ -15,9 +15,10 @@ const WalletSchema = Yup.object().shape({
   zeroConf: Yup.string().required(),
   zeroConfLimit: Yup.number()
     .integer()
-    .required()
+    .required('Zero Conf Limit is a required field')
     .min(0)
     .max(CURRENCY_MAX)
+    .transform(value => (isNaN(value) ? 0 : value))
```
RafaelTaranto: There should be a `transformNumber` in the codebase. Can this 0 be swapped to null? Also isn't the bug because of how zeroConfLimit used to be optional and is now required? I think we can just change the logic in the strip row function.

RafaelTaranto (changes requested):

RafaelTaranto (approved):

PR #777, opened 2021-06-25
Fix user token revoke

chaotixkilla:


chaotixkilla:
[Revoking admin user errors if user has modified any customer compliance types](https://trello.com/c/tFkbGlrx/855-revoking-admin-user-errors-if-user-has-modified-any-customer-compliance-types)

bin/lamassu-revoke
```diff
@@ -12,9 +14,49 @@ if (argv.length !== 1) {
...
+        idPhotoOverride,
+        cameraOverride,
+        sanctionsOverride,
+        authorizedOverride,
+        usSsnOverride
+      ])
+    })
+  })
+  .then(result => {
+    const count = _.reduce((acc, v) => acc + v.length, 0, result)
```
josepfo:
```suggestion
    const count = _.sumBy(_.size, result)
```

josepfo (changes requested):
I prefer having SQL keywords in uppercase but it's up to you.

josepfo (approved):

bin/lamassu-revoke
```diff
@@ -12,9 +14,49 @@ if (argv.length !== 1) {
...
+      console.log(`User ${name} was not found!`)
+      process.exit(0)
+    }
+
+    return _.map(t => t.token, res.rows)
+  })
+  .then(tokens => {
+    const tokenClause = _.map(pgp.as.text, tokens).join(',')
+
+    return db.task('get-touched-customers', t => {
```
RafaelTaranto: This should be a transaction

bin/lamassu-revoke
```diff
@@ -12,9 +14,49 @@ if (argv.length !== 1) {
...
+  .then(result => {
+    const count = _.sumBy(_.size, result)
+    if (count > 0) {
+      console.log(`Removed ${count} ${count === 1 ? `entry` : `entries`} on the customer table, related with the user ${name}!`)
+    }
+  })
+  .then(() => {
+    const sql = `DELETE FROM user_tokens WHERE name = $1 RETURNING *`
+    return db.result(sql, [name], r => r.rowCount)
+  })
```
RafaelTaranto: This should be inside the transaction

RafaelTaranto (changes requested):

bin/lamassu-revoke
```diff
@@ -12,9 +14,50 @@ if (argv.length !== 1) {
...
+      const smsOverride = t.any(`UPDATE customers SET sms_override_by = null WHERE sms_override_by IN ($1^) RETURNING *`, tokenClause)
+      const idDataOverride = t.any(`UPDATE customers SET id_card_data_override_by = null WHERE id_card_data_override_by IN ($1^) RETURNING *`, tokenClause)
+      const idPhotoOverride = t.any(`UPDATE customers SET id_card_photo_override_by = null WHERE id_card_photo_override_by IN ($1^) RETURNING *`, tokenClause)
+      const cameraOverride = t.any(`UPDATE customers SET front_camera_override_by = null WHERE front_camera_override_by IN ($1^) RETURNING *`, tokenClause)
+      const sanctionsOverride = t.any(`UPDATE customers SET sanctions_override_by = null WHERE sanctions_override_by IN ($1^) RETURNING *`, tokenClause)
+      const authorizedOverride = t.any(`UPDATE customers SET authorized_override_by = null WHERE authorized_override_by IN ($1^) RETURNING *`, tokenClause)
+      const usSsnOverride = t.any(`UPDATE customers SET us_ssn_override_by = null WHERE us_ssn_override_by IN ($1^)`, tokenClause)
+      const tokenDeletion = t.result(`DELETE FROM user_tokens WHERE name = $1 RETURNING *`, [name], r => r.rowCount)
+
+      return t.batch([
```
RafaelTaranto, line 40: Can't this try to delete the token before cleaning up the customer table? I don't think order is guaranteed here. Probably better to have token deletion part on a inner `then`, keep in mind that line 39 already executes the query, so even that has to be inside the `then`.

RafaelTaranto (changes requested):

RafaelTaranto (approved):

PR #781, opened 2021-06-30
fix: add us_ssn to compliance types

josepfo:


josepfo:
[Add us_ssn to db compliance types](https://trello.com/c/p2vlaNQQ/864-add-usssn-to-db-compliance-types)

PR #782, opened 2021-06-30
fix: undefined data on sidebar

josepfo:


josepfo:
[Sidebar component breaks on setup wizard](https://trello.com/c/y2zj5nTL/865-sidebar-component-breaks-on-setup-wizard)

PR #783, opened 2021-07-02
Feat: create the balance sheet screen

josepfo:


josepfo:
[Create the balance sheet screen](https://trello.com/c/LnaeTGqg/863-create-the-balance-sheet-screen)

new-lamassu-admin/src/pages/Assets/Assets.js
```diff
@@ -0,0 +1,213 @@
...
+      direction: 'in'
+    },
+    {
+      id: 'unhedgedWalletAssets',
+      display: 'Unhedged wallet assets',
+      amount: 3978,
+      currency: 'USD',
+      direction: 'in'
+    }
+  ]
```
chaotixkilla: You can move this outside the component. Hmm, if you're kind of simulating the query response and want to confine this data to the component's lifespan, I'm not sure what the correct procedure is. Better confirm with Rafael on this one

new-lamassu-admin/src/pages/Assets/Assets.js
```diff
@@ -0,0 +1,213 @@
...
+
+  const filterByClass = x =>
+    R.filter(it => R.path(['class'])(it) === x)(mockData)
+
+  return (
+    <>
+      <TitleSection title="Balance sheet" />
+      <div className={classes.root}>
+        <Grid container>
```
chaotixkilla, line 174: Isn't this Grid element redundant? As you already have the wrapping div

new-lamassu-admin/src/pages/Assets/Assets.js
```diff
@@ -0,0 +1,213 @@
...
+
+const HeaderCell = withStyles({
+  root: {
+    borderBottom: '4px solid white',
+    padding: 0,
+    paddingLeft: 20,
+    paddingRight: 20,
+    backgroundColor: 'white'
+  }
+})(TableCell)
```
chaotixkilla: You can move up the object containing
```
{
  borderBottom: '4px solid white',
  padding: 0,
  paddingLeft: 20,
  paddingRight: 20
}
```

And do something like this:
```suggestion
const Cell = withStyles({
  root: refactoredObj
})(TableCell)

const HeaderCell = withStyles({
  root: {
    ...refactoredObj
    backgroundColor: 'white'
  }
})(TableCell)
```

new-lamassu-admin/src/pages/Assets/Assets.js
```diff
@@ -0,0 +1,213 @@
...
+                      <Cell align="right">
+                        <P>{`${selectAmountPrefix(asset)}
+                         ${formatCurrency(Math.abs(asset.amount))} ${
+                          asset.currency
+                        }`}</P>
+                      </Cell>
+                    </TableRow>
+                  )
+                }
+                return null
```
chaotixkilla: I think the reverse is more readable. Pardon if formatting goes wrong :D
```suggestion
                if (!(idx < numToRender)) return <></>
                return (
                  <TableRow className={classes.row} key={idx}>
                    <Cell align="left">
                      <P>{asset.display}</P>
                    </Cell>
                    <Cell align="right">
                      <P>{`${selectAmountPrefix(asset)}
                       ${formatCurrency(Math.abs(asset.amount))} ${
                        asset.currency
                      }`}</P>
                    </Cell>
                  </TableRow>
                )
```

chaotixkilla (changes requested):

new-lamassu-admin/src/pages/Assets/Assets.js
josepfo: Hmm, I think not, if removed the tables will be displayed in a single column.

RafaelTaranto (approved):

PR #784, opened 2021-07-02
fix: align threshold title to the right

josepfo:


josepfo:
[Fix threshold value alignment](https://trello.com/c/DGenuDdq/866-fix-threshold-value-alignment)

RafaelTaranto (approved):

PR #790, opened 2021-07-09
Fix: search filter section according to spec

josepfo:


josepfo:
[Update the clean filters button on the search component to match the spec](https://trello.com/c/3PiaN6eY/390-update-the-clean-filters-button-on-the-search-component-to-match-the-spec)

new-lamassu-admin/src/components/SearchFilter.js
```diff
@@ -1,38 +1,56 @@
...
 import { chipStyles, styles } from './SearchFilter.styles'
 
 const useChipStyles = makeStyles(chipStyles)
 const useStyles = makeStyles(styles)
 
-const SearchFilter = ({ filters, onFilterDelete, setFilters }) => {
+const capitalize = R.converge(R.concat(), [
+  R.compose(R.toUpper, R.head),
+  R.tail
+])
```
chaotixkilla: Like this approach, gj!

new-lamassu-admin/src/components/SearchFilter.js
```diff
@@ -1,38 +1,56 @@
...
+              classes={chipClasses}
+              label={`${capitalize(f.type)}: ${f.value}`}
+              onDelete={() => onFilterDelete(f)}
+              deleteIcon={<CloseIcon className={classes.button} />}
+            />
+          ))}
+        </div>
+        <div className={classes.deleteWrapper}>
+          {entries && (
+            <Label2 className={classes.entries}>{`${entries} entries`}</Label2>
```
chaotixkilla: Quick nitpicking here: On the spec, this label seems to be regular weight, but on your dev prints, it seems it is bold, which leads me to believe that this may be a Label3, instead of the Label2. Please check this

chaotixkilla (changes requested):

new-lamassu-admin/src/components/SearchFilter.js
josepfo: I thought it was clean too, but I just realized we already have a fun to do this in the codebase. 🥲

chaotixkilla (approved):

RafaelTaranto (approved):

PR #800, opened 2021-07-22
feat: use new schema changing queries

RafaelTaranto:


RafaelTaranto:
[Make l-s use the new schema-changing queries](https://trello.com/c/fp12IX8Z/605-make-l-s-use-the-new-schema-changing-queries)

PR #801, opened 2021-07-22
feat: filter by txs with fiat lt 0

josepfo:


josepfo:
[Do not take into account failed/canceled transactions for compliance purposes](https://trello.com/c/bgLnwHCD/886-do-not-take-into-account-failed-canceled-transactions-for-compliance-purposes)

lib/tx.js
```diff
@@ -74,7 +74,7 @@ function customerHistory (customerId, thresholdDays) {
     FROM cash_out_txs txOut
     WHERE txOut.customer_id = $1
     AND txOut.created > now() - interval $2
-    AND (timedout = true OR error_code != 'operatorCancel')
+    AND (timedout = true OR error_code != 'operatorCancel' OR fiat > 0)
```
chaotixkilla: Shouldn't this filter also be present on the cash_in_txs side of the query?

chaotixkilla (changes requested):

lib/tx.js
josepfo: Hmm, I think it doesn't matter since we force the union but not sure how this works for performance.

PR #805, opened 2021-07-28
Fix infinite loop on id and front camera photo rejection

chaotixkilla:


chaotixkilla:
[Rejecting an ID photo in the admin causes customer to be repeatedly prompted for ID in an unsuccessful loop](https://trello.com/c/fDAaMn4y/897-rejecting-an-id-photo-in-the-admin-causes-customer-to-be-repeatedly-prompted-for-id-in-an-unsuccessful-loop)

lib/customers.js
```diff
@@ -263,16 +263,20 @@ function getComplianceTypes () {
...
+    'front_camera_path',
     'authorized',
     'us_ssn'
   ]
 
   const updatedFields = _.intersection(updateableFields, _.keys(fields))
-  const atFields = _.fromPairs(_.map(f => [`${f}_override`, 'automatic'], updatedFields))
+  const overrideFields = _.fromPairs(_.map(f =>
+    _.endsWith('_path', f)
+      ? [`${_.replace('_path', '', f)}_override`, 'automatic']
```
josepfo:
```suggestion
      ? [`${_.replace('_path', '_override', f)}`, 'automatic']
      
```

josepfo: I think we can do this instead, right?

josepfo (changes requested):

lib/customers.js
chaotixkilla: Nice catch!

josepfo (approved):

lib/customers.js
```diff
@@ -263,16 +263,20 @@ function getComplianceTypes () {
...
     'us_ssn'
   ]
 
   const updatedFields = _.intersection(updateableFields, _.keys(fields))
-  const atFields = _.fromPairs(_.map(f => [`${f}_override`, 'automatic'], updatedFields))
+  const overrideFields = _.fromPairs(_.map(f =>
+    _.endsWith('_path', f)
+      ? [`${_.replace('_path', '_override', f)}`, 'automatic']
+      : [`${f}_override`, 'automatic']
+    , updatedFields))
```
RafaelTaranto: Can't we just remove the _path at the end of every string and get rid of the ternary if here?

RafaelTaranto (changes requested):

lib/customers.js
chaotixkilla: I took another look at the code base, and I don't think so. I contextualized this fix in more detail on a comment in the respective card, but this problem happens because the DB column for `id_card_photo` and `front_camera` have the `_path` suffix, which caused the strings without the `_path` to not match with any DB column, and not update the `_override` column to `automatic`.

Simply putting it, the issue is because all other columns are structured using the logic `<column_name><suffix>` which leads to something like `id_card_data`, and with the suffix, `id_card_data_override`. With `id_card_photo` and `front_camera`, the respective columns are  `id_card_photo_path` and `id_card_photo_override`. As the string array only has `id_card_photo`, and not `id_card_photo_path`, the `_.intersection()` call does not match with the appropriate DB column, which leads to the `_override` lines to not be executed for these two columns, which leads to the `id_card_photo_override` not being updated to `automatic` and staying in `blocked`, which leads to the infinite loop.

I think the options to fix this are either this option I chose (make the columns with `_path` being appropriately matched) or rename the columns in the database to be consistent with the `<column_name><suffix>` nomenclature.

RafaelTaranto: I just meant something like:

```js
  const atFields = _.fromPairs(_.map(f => {
    const noPath = _.replace(f, '_path$', '');
    return [`${noPath}_override`, 'automatic'], updatedFields)
  })
```

RafaelTaranto: To get it even more readable you can probably do something like:

```js
    const removePathSuffix = _.replace('_path$', '')
    const getPairs = _.map(f => [`${f}_override`, 'automatic'])
    const overrideFields = _.compose(_.fromPairs, getPairs, removePathSuffix)(updatedFields)
```

RafaelTaranto (approved):

PR #808, opened 2021-08-03
feat: lamassu eth recovery script

josepfo:


josepfo:
[Restoring ETH funds](https://trello.com/c/reiD6Fy0/885-restoring-eth-funds)

chaotixkilla (approved):

bin/lamassu-eth-recovery.js
```diff
@@ -0,0 +1,54 @@
...
+  return key.derivePath(paymentPrefixPath)
+}
+
+function defaultHdNode (masterSeed) {
+  if (!masterSeed) throw new Error('No master seed!')
+  const key = hdkey.fromMasterSeed(masterSeed)
+  return key.derivePath(defaultPrefixPath)
+}
+
+run(address)
```
RafaelTaranto: Can we add a bit of a friendly UI in terms of error handling and how to use the script? Take a look at the `lamassu-register` file

We also wanna follow the same standards as other bins in terms of naming and first line of the file, loose the js extension and add the `#!/usr/bin/env node`

RafaelTaranto (changes requested):

bin/lamassu-eth-recovery.js
RafaelTaranto: We also need to add this to package.json, under `bin`

RafaelTaranto (approved):

PR #811, opened 2021-08-04
Customer custom notes

chaotixkilla:


chaotixkilla:
[Add editable notes section to customer profile page](https://trello.com/c/vWAOREVw/900-add-editable-notes-section-to-customer-profile-page)

lib/customers.js
```diff
@@ -515,6 +518,9 @@ function getCustomerById (id) {
       from cash_in_txs where send_confirmed = true union
       select 'cashOut' as tx_class, id, fiat, fiat_code, created, customer_id 
       from cash_out_txs where confirmed_at is not null) t on c.id = t.customer_id
+      left outer join (
+        select customer_id, content from customer_notes
+      ) cn on c.id = cn.customer_id
```
josepfo:
```suggestion
      LEFT OUTER JOIN (
        SELECT customer_id, content FROM customer_notes
      ) cn ON c.id = cn.customer_id
```

josepfo (changes requested):

josepfo (approved):

migrations/1627868356883-customer-custom-notes.js
```diff
@@ -0,0 +1,20 @@
...
+
+exports.up = function (next) {
+  var sql = [
+    `CREATE TABLE customer_notes (
+      id UUID PRIMARY KEY,
+      customer_id UUID NOT NULL REFERENCES customers(id),
+      created TIMESTAMPTZ NOT NULL DEFAULT now(),
+      last_edited_at TIMESTAMPTZ,
+      last_edited_by UUID REFERENCES users(id),
+      content TEXT
```
RafaelTaranto: Any reason for content allowed as null?

lib/new-admin/graphql/resolvers/customer.resolver.js
```diff
@@ -17,6 +17,10 @@ const resolvers = {
       const token = !!context.req.cookies.lid && context.req.session.user.id
       if (customerId === anonymous.uuid) return customers.getCustomerById(customerId)
       return customers.updateCustomer(customerId, customerInput, token)
+    },
+    setCustomerNotes: (...[, { customerId, newContent }, context]) => {
+      const token = !!context.req.cookies.lid && context.req.session.user.id
```
RafaelTaranto: Can't token be false here? And in other places we're doing the same? I would rather have a get token function that returns null and the null is properly handled. I know this probably won't happen but I think it's still the better option here.

RafaelTaranto (approved):

PR #813, opened 2021-08-04
Custom SMS

chaotixkilla:


chaotixkilla:
[Custom SMS notices to customers](https://trello.com/c/igDUi5Ui/899-custom-sms-notices-to-customers)

lib/plugins.js
```diff
@@ -352,19 +352,19 @@ function plugins (settings, deviceId) {
...
-        const values = [true, tx.id]
+    return sms.getCashOutReadySms(phone, timestamp)
+      .then(msg => {
+        const rec = {
+          sms: msg
+        }
+    
+        return sms.sendMessage(settings, rec)
+          .then(() => {
+            const sql = 'update cash_out_txs set notified=$1 where id=$2'
```
josepfo: Good chance to uppercase this

lib/sms.js
```diff
@@ -1,5 +1,44 @@
...
+          body: messageContent
+        }
+      }
+
+      return {
+        toNumber: phone,
+        body: `Your cash is waiting! Go to the Cryptomat and press Redeem within 24 hours. [${timestamp}]`
+      }
+    })
+}
```
josepfo: Looks good but these functions seem similar enough so we can define a getCustomSms(phone, event, content) that changes according to the use case (cash out ready or code).

lib/plugins.js
```diff
@@ -352,19 +352,19 @@ function plugins (settings, deviceId) {
...
-
-    return sms.sendMessage(settings, rec)
-      .then(() => {
-        const sql = 'update cash_out_txs set notified=$1 where id=$2'
-        const values = [true, tx.id]
+    return sms.getCashOutReadySms(phone, timestamp)
+      .then(msg => {
+        const rec = {
+          sms: msg
+        }
```
josepfo: Maybe we should change 'msg' to something like 'smsStructure' since this is an object, what do you think?

lib/plugins.js
```diff
@@ -713,15 +713,15 @@ function plugins (settings, deviceId) {
...
-      }
-    }
-
-    return sms.sendMessage(settings, rec)
-      .then(() => code)
+    return sms.getPhoneCodeSms(phone, code)
+      .then(msg => {
+        const rec = {
+          sms: msg
+        }
```
josepfo: Here as well

josepfo (changes requested):
Looking good!

josepfo (approved):

RafaelTaranto (approved):

lib/plugins.js
```diff
@@ -352,19 +352,19 @@ function plugins (settings, deviceId) {
...
-        toNumber: phone,
-        body: `Your cash is waiting! Go to the Cryptomat and press Redeem within 24 hours. [${timestamp}]`
-      }
-    }
-
-    return sms.sendMessage(settings, rec)
-      .then(() => {
-        const sql = 'update cash_out_txs set notified=$1 where id=$2'
-        const values = [true, tx.id]
+    return sms.getSms('cash_out_dispense_ready', phone, { timestamp })
```
joshmh: I think we should make these keys constants so we'll have them all in one place and guard against typos.

lib/sms.js
```diff
@@ -1,5 +1,33 @@
...
+  cashOutDispenseReady: `Your cash is waiting! Go to the Cryptomat and press Redeem within 24 hours. [${content.timestamp}]`
+})
+
+function getSms (event, phone, content) {
+  return customSms.getCustomMessage(event)
+    .then(msg => {
+      if (!_.isNil(msg)) {
+        var accMsg = msg.message
+        const contentKeys = _.keys(content)
+        const messageContent = _.reduce((acc, it) => _.replace(`#${it}`, content[it], acc), accMsg, contentKeys)
```
joshmh, line 19: Will this work for different languages? What if we are inserting numbers and we have to adjust plurals?

joshmh (changes requested):

PR #822, opened 2021-08-18
feat: simplified tx exports

josepfo:


josepfo:
[Simplified transaction exports](https://trello.com/c/ngVMjP3e/914-simplified-transaction-exports)

new-lamassu-admin/src/components/LogsDownloaderPopper.js
```diff
@@ -209,7 +232,9 @@ const LogsDownloaderPopover = ({ name, query, args, title, getLogs }) => {
 
   const radioButtonOptions = [
     { display: 'All logs', code: ALL },
-    { display: 'Date range', code: RANGE }
+    { display: 'Date range', code: RANGE },
+    { display: 'Advanced logs', code: ADVANCED },
+    { display: 'Simplified logs', code: SIMPLIFIED }
   ]
```
RafaelTaranto: It's better to create two different arrays for this, no reason to keep this together if we're always slicing it.

lib/new-admin/transactions.js
```diff
@@ -76,6 +78,46 @@ function batch (from = new Date(0).toISOString(), until = new Date().toISOString
...
+  return batch(from, until, limit, offset)
+    .then(res => {
+      const fields = ['txClass', 'id', 'created', 'machineName',
+        'cryptoCode', 'fiat', 'fiatCode', 'phone', 'toAddress',
+        'txHash', 'dispense', 'error', 'status', 'fiatProfit', 'cryptoAmount']
+      return _.map(_.compose(_.pick(fields), getStatus, getProfit, getCryptoAmount))(res)
+    })
+}
+
+const getCryptoAmount = it => { return { ...it, cryptoAmount: coinUtils.toUnit(BN(it.cryptoAtoms), it.cryptoCode).toString() } }
```
RafaelTaranto:
```suggestion
const getCryptoAmount = it => ({ ...it, cryptoAmount: coinUtils.toUnit(BN(it.cryptoAtoms), it.cryptoCode).toString() })
```

lib/new-admin/transactions.js
```diff
@@ -76,6 +78,46 @@ function batch (from = new Date(0).toISOString(), until = new Date().toISOString
...
+  if (it.expired) return { ...it, status: 'Expired' }
+  return { ...it, status: 'Pending' }
+}
+
+const getStatus = it => {
+  if (it.txClass === 'cashOut') {
+    return getCashOutStatus(it)
+  }
+  return getCashInStatus(it)
+}
```
RafaelTaranto: These functions are misleading, we should only return what the name is telling us. e.g. `getStatus` should return only the status, not the entire object

lib/new-admin/transactions.js
```diff
@@ -76,6 +78,46 @@ function batch (from = new Date(0).toISOString(), until = new Date().toISOString
   ]).then(packager)
 }
 
+function simplifiedBatch (from = new Date(0).toISOString(), until = new Date().toISOString(), limit = null, offset = 0) {
+  return batch(from, until, limit, offset)
+    .then(res => {
+      const fields = ['txClass', 'id', 'created', 'machineName',
+        'cryptoCode', 'fiat', 'fiatCode', 'phone', 'toAddress',
+        'txHash', 'dispense', 'error', 'status', 'fiatProfit', 'cryptoAmount']
+      return _.map(_.compose(_.pick(fields), getStatus, getProfit, getCryptoAmount))(res)
```
RafaelTaranto:
```js
const addSimplifiedFields = _.map(it => ({
  ...it,
  status: getStatus(it),
  profit: getProfit(it),
}))

return _.compose(_.pick(fields), addSimplifiedFields)(res)

```

lib/new-admin/transactions.js
```diff
@@ -76,6 +78,46 @@ function batch (from = new Date(0).toISOString(), until = new Date().toISOString
...
+    })
+}
+
+const getCryptoAmount = it => { return { ...it, cryptoAmount: coinUtils.toUnit(BN(it.cryptoAtoms), it.cryptoCode).toString() } }
+
+const getProfit = it => {
+  const getCommissionFee = it => Number.parseFloat(it.commissionPercentage) * Number.parseFloat(it.fiat)
```
RafaelTaranto: It's probably better to use BigNumber here

lib/new-admin/transactions.js
```diff
@@ -76,6 +78,46 @@ function batch (from = new Date(0).toISOString(), until = new Date().toISOString
...
+  if (!it.cashInFee) return { ...it, fiatProfit: getCommissionFee(it) }
+  return { ...it, fiatProfit: getCommissionFee(it) + Number.parseFloat(it.cashInFee) }
```
RafaelTaranto: BigNumber here as well

lib/new-admin/graphql/schema.js
```diff
@@ -394,6 +395,8 @@ const resolvers = {
       transactions.batch(from, until, limit, offset, deviceId),
     transactionsCsv: (...[, { from, until, limit, offset }]) =>
       transactions.batch(from, until, limit, offset).then(parseAsync),
+    simplifiedTransactionsCsv: (...[, { from, until, limit, offset }]) =>
+      transactions.simplifiedBatch(from, until, limit, offset).then(parseAsync),
```
RafaelTaranto: Can't we pass down a "simplified" arg to decide if we want the entire or the simplified version? This will probably get extended in the future and we'll have to add a bunch of different filters.

new-lamassu-admin/src/pages/Transactions/Transactions.js
```diff
@@ -167,8 +173,14 @@ const Transactions = () => {
               <LogsDowloaderPopover
                 title="Download logs"
                 name="transactions"
-                query={GET_TRANSACTIONS_CSV}
+                queries={[
+                  GET_TRANSACTIONS_CSV,
+                  GET_SIMPLIFIED_TRANSACTIONS_CSV
+                ]}
```
RafaelTaranto: With the "simplified" arg we don't need two queries.

new-lamassu-admin/src/components/LogsDownloaderPopper.js
```diff
@@ -256,10 +281,18 @@ const LogsDownloaderPopover = ({ name, query, args, title, getLogs }) => {
...
+            <div className={classes.radioButtonsContainer}>
+              <RadioGroup
+                name="simplified-tx-logs"
+                value={selectedAdvancedRadio}
+                options={radioButtonOptions.slice(2, 4)}
+                ariaLabel="simplified-tx-logs"
+                onChange={handleAdvancedRadioButtons}
+                className={classes.radioButtons}
+              />
+            </div>
```
RafaelTaranto: Can't this be handled by the external component? Not every log download will have simplified and advanced options

RafaelTaranto (changes requested):

lib/new-admin/transactions.js
josepfo: hmm, we're doing similar calculations at the front-end level (for the system performance panel I believe) and we use Number. we better set a standard for this, right?

RafaelTaranto: Yeah, everything financial related should be BigNumber

RafaelTaranto (approved):

PR #830, opened 2021-09-01
Add support for multiple cassettes

chaotixkilla:


chaotixkilla:
[Work on having multiple cassette boxes](https://trello.com/c/7MTrqh00/925-work-on-having-multiple-cassette-boxes)

lib/plugins.js
```diff
@@ -119,53 +125,56 @@ function plugins (settings, deviceId) {
       return _.map(r => r[0] + r[1].provisioned, _.zip(sum, tx.bills))
     }
 
-    const provisioned = _.reduce(sumTxs, [0, 0], redeemableTxs)
-    const zipped = _.zip(_.map('count', cassettes), provisioned)
+    const provisioned = _.reduce(sumTxs, Array(_.size(orderedCassettes)).fill(0), redeemableTxs)
```
josepfo: what do you think about doing it this way just to follow the pattern you've been using? 
```suggestion
    const provisioned = _.reduce(sumTxs, _.times(_.constant(0), _.size(orderedCassettes)), redeemableTxs)
```

lib/plugins.js
```diff
@@ -119,53 +125,56 @@ function plugins (settings, deviceId) {
...
-            count: parseInt(counts[0], 10)
-          },
-          {
-            denomination: parseInt(denominations[1], 10),
-            count: parseInt(counts[1], 10)
-          }
-        ]
+        if (rec.counts.length !== denominations.length) {
+          throw new Error('Denominations and respective counts do not match!')
+        }
```
josepfo, line 165: Can you explain why do we need to throw this?

lib/plugins/wallet/mock-wallet/mock-wallet.js
```diff
@@ -7,7 +7,7 @@ const NAME = 'FakeWallet'
 const SECONDS = 1000
 const PUBLISH_TIME = 3 * SECONDS
 const AUTHORIZE_TIME = PUBLISH_TIME + 5 * SECONDS
-const CONFIRM_TIME = AUTHORIZE_TIME + 120 * SECONDS
+const CONFIRM_TIME = AUTHORIZE_TIME + 10 * SECONDS
```
josepfo: Is this change  intended?

josepfo (changes requested):

lib/plugins.js
chaotixkilla: As we're moving away from 2 hardcoded cassettes to a variable number, it's just a sanity check to ensure that we're not using mismatched cassette configs

josepfo (approved):

PR #839, opened 2021-09-07
feat: add script to update nodes

josepfo:


josepfo:
[Update wallet nodes upon update of lamassu-server](https://trello.com/c/3VeVACyF/936-update-wallet-nodes-upon-update-of-lamassu-server)

bin/lamassu-update-wallet-nodes
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env node
+
+const _ = require('lodash/fp')
+
+const { utils: coinUtils } = require('lamassu-coins')
+const cryptos = coinUtils.cryptoCurrencies()
+
+const common = require('../lib/blockchain/common')
```
chaotixkilla: Group this `require(...)` with the other ones

bin/lamassu-update-wallet-nodes
```diff
@@ -0,0 +1,37 @@
...
+
+const PLUGINS = {
+  BTC: require('../lib/blockchain/bitcoin.js'),
+  LTC: require('../lib/blockchain/litecoin.js'),
+  ETH: require('../lib/blockchain/ethereum.js'),
+  DASH: require('../lib/blockchain/dash.js'),
+  ZEC: require('../lib/blockchain/zcash.js'),
+  BCH: require('../lib/blockchain/bitcoincash.js')
+}
```
chaotixkilla, line 16: While this works for current coins, the script will need manual maintenance for every new coin added to the system. Is something like this possible?

```suggestion
const PLUGINS = {}
_.forEach(it => PLUGINS[it.cryptoCode] = require(`../lib/blockchain/${it.code}.js`), cryptos)
```

chaotixkilla: I guess that eventually manual maintenance will be needed, as the coin script will need to be developed, so you can ignore this I think

chaotixkilla (changes requested):

bin/lamassu-update-wallet-nodes
josepfo: Yeah, I think we can't run from it other than placing this in a common file

chaotixkilla (approved):

PR #842, opened 2021-09-09
Fix customer history retrieval

chaotixkilla:


chaotixkilla:
[Bug - Compliance trigger transaction volume](https://trello.com/c/Xq7OxmVl/941-bug-compliance-trigger-transaction-volume)

josepfo (approved):

lib/tx.js
```diff
@@ -65,20 +68,23 @@ function cancel (txId) {
 }
 
 function customerHistory (customerId, thresholdDays) {
-  const sql = ` SELECT txIn.id, txIn.created, txIn.fiat, 'cashIn' AS direction
-    FROM cash_in_txs txIn
-    WHERE txIn.customer_id = $1
-    AND txIn.created > now() - interval $2
+  const sql = `SELECT * FROM (
+    SELECT txIn.id, txIn.created, txIn.fiat, 'cashIn' AS direction,
+      ((NOT txIn.send_confirmed) AND (txIn.created <= now() - interval $3)) AS expired
```
RafaelTaranto, line 73: Do we really expire cashIn transatcion? Won't the operator already have the customer money at that point and be required to send the coins anyways?

RafaelTaranto (changes requested):

lib/tx.js
RafaelTaranto: Just talked about this with Melanie and apparently it expires 0 fiat value cash-in txs. This should be fine!

RafaelTaranto (approved):

PR #850, opened 2021-09-17
Add operator ID to local app endpoints

chaotixkilla:


chaotixkilla:
[Update the l-s localApp to receive operatorID as well](https://trello.com/c/SpmvUILT/954-update-the-l-s-localapp-to-receive-operatorid-as-well)

PR #853, opened 2021-09-21
fix: locales override language limit

siiky:


siiky:
[Inconsistency when adding languages in Locales](https://trello.com/c/P15cd05F/960-inconsistency-when-adding-languages-in-locales)

chaotixkilla (approved):

RafaelTaranto (approved):

PR #854, opened 2021-09-21
docs: update installation process

siiky:


siiky:
[Update l-s INSTALL*.md](https://trello.com/c/egHXHumO/965-update-l-s-installmd)

ubavic (changes requested):
Changes made in this PR significantly improve the quality of instructions.

However, instruction for generating certificates should be also updated because in 57e3b5b `cert-gen.sh` was moved to `tools/`.

Some things that should be also considered:
 - The line about openssl version should be removed (at least for Nix instructions). Installation works with versions bigger than 1.1.
 - Maybe we should explicitly instruct the user to add the certificate exception to graphql.

siiky:
> However, instruction for generating certificates should be also updated because in 57e3b5b cert-gen.sh was moved to tools/.

Fixed -- just merged dev into this branch.

> The line about openssl version should be removed (at least for Nix instructions). Installation works with versions bigger than 1.1.

I mentioned that to @RafaelTaranto once, but didn't get to change it yet. And I'm not sure what the right change is, anyway. We know that it works with, for example, OpenSSL 1.1.1g (the version installed with Nix), but doesn't it work with other versions? Is there still an OpenSSL version restriction?

> Maybe we should explicitly instruct the user to add the certificate exception to graphql.

That's a good idea!

PR #849, opened 2021-09-16
Reintroduce /dbchange

chaotixkilla:

PR #848, opened 2021-09-16
Allow polling of multiple db schemas

chaotixkilla:

PR #847, opened 2021-09-16
Changes to better support asynclocalstorage on l-a-s-p

chaotixkilla:
Based of https://github.com/lamassu/lamassu-server/pull/846

Relevant changes starting on commit https://github.com/lamassu/lamassu-server/commit/2b54ad5fd4b28499d3ac2effe3d5aab4ecb0f37e

PR #846, opened 2021-09-15
Add asynclocalstorage as middleware to l-a-s

chaotixkilla:

PR #856, opened 2021-09-23
Pazuz-ready changes

chaotixkilla:
Absorbs the following:
https://github.com/lamassu/lamassu-server/pull/850
https://github.com/lamassu/lamassu-server/pull/849
https://github.com/lamassu/lamassu-server/pull/848
https://github.com/lamassu/lamassu-server/pull/847
https://github.com/lamassu/lamassu-server/pull/846


chaotixkilla:
[Get the project running on a droplet (stage 1)](https://trello.com/c/4Jvjm70n/957-get-the-project-running-on-a-droplet-stage-1)

lib/db.js
```diff
@@ -36,7 +37,10 @@ const _task = (obj, opts, cb) => {
   })
 }
 
-const getSchema = () => 'public'
+const getSchema = () => {
+  const store = asyncLocalStorage.getStore() ?? defaultStore()
```
RafaelTaranto, line 42: I don't think we should go to defaultStore here. If the storage isn't set at this point i think the query should fail fast on lamassu so we can spot it before we get silent issues in pazuz

lib/new-admin/middlewares/session.js
```diff
@@ -19,12 +20,16 @@ const getSecret = () => {
 
 const hostname = options.hostname
 
+const lamaDb = {
+  query: (query, values, qrm) => db.query(query, values, qrm, false)
+}
```
RafaelTaranto: What's this doing?

new-lamassu-admin/src/pages/Authentication/Input2FAState.js
```diff
@@ -56,7 +58,21 @@ const Input2FAState = ({ state, dispatch }) => {
...
+          context: {
+            headers: {
+              'Pazuz-Operator-Identifier': base64.encode(state.clientField)
+            }
+          }
+        }
+        return getUserData(
+          process.env.REACT_APP_BUILD_TARGET === 'LAMASSU'
+            ? R.omit(['context'], options)
+            : options
```
RafaelTaranto: Is there harm sending in the context?

new-lamassu-admin/src/pages/UserManagement/modals/CreateUserModal.js
```diff
@@ -74,7 +75,11 @@ const CreateUserModal = ({ state, dispatch }) => {
 
   const [createUser, { error }] = useMutation(CREATE_USER, {
     onCompleted: ({ createRegisterToken: token }) => {
-      setCreateUserURL(urlResolver(`/register?t=${token.token}`))
+      const queryParams =
+        process.env.REACT_APP_BUILD_TARGET === 'LAMASSU'
+          ? `t=${token.token}`
+          : `t=${token.token}&id=${base64.encode(usernameField)}`
```
RafaelTaranto, line 82: What does the ID does on the server side for this one? Might be a good idea to leave a comment on this

RafaelTaranto (changes requested):

lib/new-admin/middlewares/session.js
chaotixkilla: I believe it's older code from this PR that I haven't reverted

PR #857, opened 2021-09-29
chore: builded react files

RafaelTaranto:


RafaelTaranto:
[Chore: builded react files](https://trello.com/c/OrcudWqO/977-chore-builded-react-files)

PR #866, opened 2021-10-07
Fix: getCustomerList SQL query

ubavic:


ubavic:
[List of customers in compliance is empty](https://trello.com/c/bDahXuuO/993-list-of-customers-in-compliance-is-empty)

RafaelTaranto (approved):

PR #876, opened 2021-10-12
fix: machine unpair breaks the UI

ubavic:
Purging db tables upon unpairing from the server will prevent later UI crashes caused by working with a NULL value.

### DIFF
--- a/lib/pairing.js
+++ b/lib/pairing.js
@@ -13,12 +13,13 @@ function pullToken (token) {
 }
 
 function unpair (deviceId) {
-  const sql = 'delete from devices where device_id=$1'
-  const deleteMachinePings = 'delete from machine_pings where device_id=$1'
-  const deleteMachineHeartbeat = 'delete from machine_network_heartbeat where device_id=$1'
-
   // TODO new-admin: We should remove all configs related to that device. This can get tricky.
-  return Promise.all([db.none(sql, [deviceId]), db.none(deleteMachinePings, [deviceId]), db.none(deleteMachineHeartbeat, [deviceId])])
+  return db.tx(async t => {
+    await t.none('DELETE FROM devices WHERE device_id=$1', [deviceId])
+    await t.none('DELETE FROM machine_pings WHERE device_id=$1', [deviceId])
+    await t.none('DELETE FROM machine_network_heartbeat WHERE device_id=$1', [deviceId])
+    await t.none('DELETE FROM machine_network_performance WHERE device_id=$1', [deviceId])
+  })
 }
 
 function pair (token, deviceId, machineModel) {
--- a/lib/pairing.js
+++ b/lib/pairing.js
@@ -15,9 +15,10 @@ function pullToken (token) {
 function unpair (deviceId) {
   const sql = 'delete from devices where device_id=$1'
   const deleteMachinePings = 'delete from machine_pings where device_id=$1'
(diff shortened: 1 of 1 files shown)


ubavic:
[Unpair of a machine from server is causing admin page to go blank](https://trello.com/c/VQyrV7kq/1007-unpair-of-a-machine-from-server-is-causing-admin-page-to-go-blank)

lib/pairing.js
```diff
@@ -15,9 +15,10 @@ function pullToken (token) {
 function unpair (deviceId) {
   const sql = 'delete from devices where device_id=$1'
   const deleteMachinePings = 'delete from machine_pings where device_id=$1'
+  const deleteMachineHeartbeat = 'delete from machine_network_heartbeat where device_id=$1'
 
   // TODO new-admin: We should remove all configs related to that device. This can get tricky.
-  return Promise.all([db.none(sql, [deviceId]), db.none(deleteMachinePings, [deviceId])])
+  return Promise.all([db.none(sql, [deviceId]), db.none(deleteMachinePings, [deviceId]), db.none(deleteMachineHeartbeat, [deviceId])])
```
RafaelTaranto: This should either be inside a transacion or we should join the 3 deletes into a single query.

RafaelTaranto (changes requested):

RafaelTaranto (approved):

lib/pairing.js
```diff
@@ -13,11 +13,13 @@ function pullToken (token) {
...
-  const deleteMachinePings = 'delete from machine_pings where device_id=$1'
-
   // TODO new-admin: We should remove all configs related to that device. This can get tricky.
-  return Promise.all([db.none(sql, [deviceId]), db.none(deleteMachinePings, [deviceId])])
+  return db.tx(async t => {
+    await t.none('DELETE FROM devices WHERE device_id=$1', [deviceId])
+    await t.none('DELETE FROM machine_pings WHERE device_id=$1', [deviceId])
+    await t.none('DELETE FROM machine_network_heartbeat WHERE device_id=$1', [deviceId])
+    await t.none('DELETE FROM machine_network_performance WHERE device_id=$1', [deviceId])
+  })
```
RafaelTaranto: This is doing serialized queries, I don't think it's exactly what we want.

You can unserialize by doing the following:

```js
  return db.tx(t => {
    const q1 = t.none('DELETE FROM devices WHERE device_id=$1', [deviceId])
    const q2 = t.none('DELETE FROM machine_pings WHERE device_id=$1', [deviceId])
    const q3 = t.none('DELETE FROM machine_network_heartbeat WHERE device_id=$1', [deviceId])
    const q4 = t.none('DELETE FROM machine_network_performance WHERE device_id=$1', [deviceId])
    return Promise.all([q1, q2, q3, q4])
  })
```

or


```js
  return db.tx(t => {
    t.none('DELETE FROM devices WHERE device_id=$1', [deviceId])
    t.none('DELETE FROM machine_pings WHERE device_id=$1', [deviceId])
    t.none('DELETE FROM machine_network_heartbeat WHERE device_id=$1', [deviceId])
    t.none('DELETE FROM machine_network_performance WHERE device_id=$1', [deviceId])
  })
```

RafaelTaranto (changes requested):
We should either stick with prom

PR #877, opened 2021-10-13
fix: expired column in customer transactions history

siiky:
I initially had two seperate commits, one to fix the `expired` column, and another to improve FP precision, but while touching up the first commit with `git rebase` I must have accidentally merged the other into it... I started trying to fix it but it was taking too much effort, and thought it wasn't worth it. Let me know if you want two separate commits and I can do it.

### DIFF
--- a/lib/admin/transactions.js
+++ b/lib/admin/transactions.js
@@ -33,7 +33,7 @@ function batch () {
   order by created desc limit $2`
 
   const cashOutSql = `select 'cashOut' as tx_class, cash_out_txs.*,
-  (extract(epoch from (now() - greatest(created, confirmed_at))) * 1000) >= $2 as expired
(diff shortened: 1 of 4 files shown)


siiky:
[Compliance - consecutive day trigger not triggering correctly when set for more than 1 day](https://trello.com/c/9KDH1LMm/1012-compliance-consecutive-day-trigger-not-triggering-correctly-when-set-for-more-than-1-day)

ubavic:
This looks good to me.
However I have a suggestion: Instead of using the epoch time and calculating with FP we could use Postgres `EXTRACT` function like this: 

```pgsql
EXTRACT(day FROM (NOW() - GREATEST(created, confirmed_at))) >= $1
```
where `$1` number of days.

siiky:
We discussed this and decided against it.

RafaelTaranto (approved):

PR #884, opened 2021-10-18
fix: don't notify of transactions if disabled

siiky:


siiky:
[With all notifications disabled, transaction notifications still appearing under notification icon](https://trello.com/c/tVNhoICI/1024-with-all-notifications-disabled-transaction-notifications-still-appearing-under-notification-icon)

PR #886, opened 2021-10-18
Backport database selector flag

chaotixkilla:


chaotixkilla:
[Backport database selector flag](https://trello.com/c/ZEMWK7tw/1026-backport-database-selector-flag)

josepfo (approved):

.env
```diff
@@ -0,0 +1 @@
+LAMASSU_DB=RELEASE
```
RafaelTaranto: Can we rename this .env to .env.dev or .env.sample and add .env to gitignore?

lib/options-loader.js
```diff
@@ -2,8 +2,15 @@ const fs = require('fs')
...
 const argv = require('minimist')(process.argv.slice(2))
+const _ = require('lodash/fp')
 
-const STRESS_TEST_DB = 'psql://postgres:postgres123@localhost/lamassu_stress'
+require('dotenv').config()
+
+const dbMapping = psqlConf => ({
+  STRESS_TEST: _.replace('lamassu', 'lamassu_stress', psqlConf),
+  RELEASE: _.replace('lamassu', 'lamassu_release', psqlConf),
+  DEV: _.replace('lamassu', 'lamassu', psqlConf)
```
RafaelTaranto: Maybe we should add a PROD flag that's exactly the same as dev for now.

lib/options-loader.js
```diff
@@ -32,9 +39,7 @@ function load () {
       opts: JSON.parse(fs.readFileSync(globalConfigPath))
     }
 
-    if (argv.testDB) {
-      config.opts.postgresql = STRESS_TEST_DB
-    }
+    config.opts.postgresql = dbMapping(config.opts.postgresql)[process.env.LAMASSU_DB]
```
RafaelTaranto: Let's default to to PROD.

lib/options-loader.js
```diff
@@ -45,9 +50,7 @@ function load () {
         opts: JSON.parse(fs.readFileSync(homeConfigPath))
       }
 
-      if (argv.testDB) {
-        config.opts.postgresql = STRESS_TEST_DB
-      }
+      config.opts.postgresql = dbMapping(config.opts.postgresql)[process.env.LAMASSU_DB]
```
RafaelTaranto: default to PROD as well

RafaelTaranto (changes requested):

lib/options-loader.js
```diff
@@ -2,8 +2,16 @@ const fs = require('fs')
 const path = require('path')
 const os = require('os')
 const argv = require('minimist')(process.argv.slice(2))
+const _ = require('lodash/fp')
 
-const STRESS_TEST_DB = 'psql://postgres:postgres123@localhost/lamassu_stress'
+require('dotenv').config({ path: `.env${process.env.NODE_ENV !== 'production' ? `.dev` : ``}` })
```
RafaelTaranto: I don't think we should do this either. The idea here is for everyone to just move the .env over and have it not versioned with whatever config they need. Right now I don't think even prod is using NODE_ENV.

lib/options-loader.js
```diff
@@ -32,9 +40,11 @@ function load () {
       opts: JSON.parse(fs.readFileSync(globalConfigPath))
     }
 
-    if (argv.testDB) {
-      config.opts.postgresql = STRESS_TEST_DB
-    }
+    config.opts.postgresql = dbMapping(config.opts.postgresql)[
+      process.env.LAMASSU_DB === '' || !process.env.LAMASSU_DB
```
RafaelTaranto: Isn't `!process.env.LAMASSU_DB` enough?

lib/options-loader.js
```diff
@@ -45,9 +55,17 @@ function load () {
...
+        process.env.LAMASSU_DB === '' || !process.env.LAMASSU_DB
+        ? 'PROD'
+        : process.env.LAMASSU_DB
+      ])
+
+      config.opts.postgresql = dbMapping(config.opts.postgresql)[
+        process.env.LAMASSU_DB === '' || !process.env.LAMASSU_DB
+        ? 'PROD'
+        : process.env.LAMASSU_DB
+      ]
```
RafaelTaranto: If htis is all inside the same function can't we decided on the env on top and reuse a db variable?

RafaelTaranto (approved):

RafaelTaranto (approved):

PR #889, opened 2021-10-20
feat: send ping and speedtest URLs to machine

siiky:

PR #890, opened 2021-10-20
fix: download_speed non-null constraint violation

siiky:
Depends on https://github.com/lamassu/lamassu-server/pull/889


siiky:
[value in column "download_speed" violates not-null constraint](https://trello.com/c/AKDxVbRX/1031-value-in-column-downloadspeed-violates-not-null-constraint)

siiky:
https://github.com/lamassu/lamassu-machine/pull/703 takes care of it on the machine side.

PR #905, opened 2021-10-27
fix: log only a message in case of insufficient funds

ubavic:


ubavic:
[Friendlier error messages for lamassu-send-coins](https://trello.com/c/mFbvHfWs/1019-friendlier-error-messages-for-lamassu-send-coins)

chaotixkilla (approved):

bin/lamassu-send-coins
```diff
@@ -80,7 +80,12 @@ settingsLoader.loadLatest()
                 console.log('Success.')
                 process.exit(0)
               })
-              .catch(console.log)
+              .catch(e => {
+                if (e.name === 'InsufficientFundsError')
+                  console.log(e.message)
+                else
+                  console.log(e)
+              })
```
RafaelTaranto: Shouldn't we always print the message only at all times? 

From the card: "Instead, if there's an error, it's enough to show only the first error line"

RafaelTaranto (changes requested):

bin/lamassu-send-coins
ubavic: Oh damn, I thought that we want only to do this for `InsufficientFundsError`.
Fixing right now

RafaelTaranto (approved):

PR #910, opened 2021-10-28
fix: change color according to notification settings

siiky:


siiky:
[Cash cassette balance alerts not consistent](https://trello.com/c/dCHZvkoj/1052-cash-cassette-balance-alerts-not-consistent)

ubavic:
I am not currently on release server, but I think that second point from my comment on [dev PR](https://github.com/lamassu/lamassu-server/pull/912#issuecomment-959974114) is also applicable here: we need also to update the Dashboard.

ubavic (approved):

PR #912, opened 2021-10-29
fix: change color according to notification settings

siiky:


siiky:
[Cash cassette balance alerts not consistent](https://trello.com/c/dCHZvkoj/1052-cash-cassette-balance-alerts-not-consistent)

ubavic:
Card also addressed the same problem on the Dashboard. We need also to change `MachineComponenents/Casettess/Cassettess.js` 
 
 P.S.  There seems to be much of duplicate code, so we will maybe need to clean little bit in some future card.

ubavic (approved):

ubavic (approved):

PR #919, opened 2021-11-10
fix: confidence checking ui up to spec

josepfo:


josepfo:
[Update confidence checking UI to match the spec.](https://trello.com/c/RxHe0WYu/1058-update-confidence-checking-ui-to-match-the-spec)

chaotixkilla (approved):

new-lamassu-admin/src/pages/Wizard/components/Wallet/ChooseCoin.js
```diff
@@ -38,7 +38,7 @@ const ChooseCoin = ({ addData }) => {
     if (!schema.isValidSync(it)) return setError(true)
 
     if (it.coin !== 'BTC') {
-      return addData({ coin: it.coin, zeroConf: 'all-zero-conf' })
+      return addData({ coin: it.coin, zeroConf: 'none', zeroConfLimit: 0 })
     }
```
josepfo, line 42: We might be able to remove this since we set default the same default values for all coins, am I right? @RafaelTaranto

RafaelTaranto: Isn't this where we're setting the default? 

For new configs I think we still need it.

PR #929, opened 2021-11-15
Fix database initialization without .env file

chaotixkilla:


chaotixkilla:
[Fix lamassu-server db initialization without .env file (via pazuz-server)](https://trello.com/c/QuHawNqE/1083-fix-lamassu-server-db-initialization-without-env-file-via-pazuz-server)

PR #930, opened 2021-11-15
fix: fix bugs in Cashbox history

ubavic:


ubavic:
[Cashbox history isn't displaying any history](https://trello.com/c/l8WjAy3e/1060-cashbox-history-isnt-displaying-any-history)

new-lamassu-admin/src/pages/Maintenance/CashboxHistory.js
```diff
@@ -111,21 +110,27 @@ const CashboxHistory = ({ machines, currency }) => {
...
+    const _performedBy = R.prop(
+      'performedBy',
+      R.find(f => f.id === row.id, fields)
+    )
+
+    const performedBy = _performedBy === '' ? null : _performedBy
+
     schema
-      .isValid(fields)
+      .isValid(performedBy)
```
siiky: Looking at the `schema`, I don't understand why this is working... From my understanding, the variable given to `isValid()` should be an object with the key `"performedBy"`.
But again, it's working on release too, so maybe it's my misunderstanding.

new-lamassu-admin/src/pages/Maintenance/CashboxHistory.js
```diff
@@ -192,16 +197,25 @@ const CashboxHistory = ({ machines, currency }) => {
       width: 180,
       textAlign: 'left',
       view: it => {
-        if (!editing)
+        if (!R.any(R.propEq('id', it.id), fields))
```
siiky: You could define `editing` as a function over `it`, to be able to reuse below.

siiky (commented):
WDYT?

It cleanly applied to release as well, and it worked like a charm there!

siiky (approved):
LGTM! It applied cleanly on release again. Thanks

RafaelTaranto (approved):

PR #935, opened 2021-11-18
Decouple l-s entrypoint

chaotixkilla:


chaotixkilla:
[Decouple l-s entrypoint](https://trello.com/c/8iWl8HjH/1091-decouple-l-s-entrypoint)

tools/build-prod-env.js
```diff
@@ -0,0 +1,47 @@
...
+
+setEnvVariable('STRIKE_BASE_URL', `https://api.strike.acinq.co/api/`)
+setEnvVariable('COIN_ATM_RADAR_URL', `https://coinatmradar.info/api/lamassu/`)
+
+setEnvVariable('OFAC_SOURCES_NAMES', 'sdn_advanced,cons_advanced')
+setEnvVariable('OFAC_SOURCES_URLS', 'https://www.treasury.gov/ofac/downloads/sanctions/1.0/sdn_advanced.xml,https://www.treasury.gov/ofac/downloads/sanctions/1.0/cons_advanced.xml')
+
+setEnvVariable('HOSTNAME', `${argv.hostname}`)
+setEnvVariable('LOG_LEVEL', 'info')
+setEnvVariable('TESTVAR', 'test')
```
ubavic: Is this left behind?

.sample.env
```diff
@@ -0,0 +1,59 @@
...
+HOSTNAME=
+LOG_LEVEL=
+LIGHTNING_NETWORK_DAEMON=
+
+## Deprecated or in deprecation
+
+HTTP=
+DEV_MODE=
+
+## Uncategorized variables
```
ubavic: No newline at the end of file

tools/build-dev-env.js
```diff
@@ -0,0 +1,33 @@
+const fs = require('fs')
+const os = require('os')
+const path = require('path')
+
+const setEnvVariable = require('./set-env-var')
+
+const ENV_SAMPLE = fs.readFileSync(path.resolve(__dirname, '../.sample.env'), 'utf-8').split(os.EOL)
+fs.writeFileSync(path.resolve(__dirname, '../.env'), ENV_SAMPLE.join(os.EOL))
```
ubavic: Why we do this?
Couldn't  `setEnvVariable` just push to the empty `.env`?

tools/build-dev-env.js
```diff
@@ -0,0 +1,33 @@
...
+
+setEnvVariable('LAMASSU_DB', 'DEV')
+setEnvVariable('POSTGRES_USER', 'postgres')
+setEnvVariable('POSTGRES_PASSWORD', 'postgres123')
+setEnvVariable('POSTGRES_HOST', 'localhost')
```
ubavic, line 12: Should this be also `${argv.hostname}`?

tools/build-prod-env.js
```diff
@@ -0,0 +1,47 @@
...
+
+const requiredParams = ['db-password', 'hostname']
+
+if (!_.isEqual(_.intersection(_.keys(argv), requiredParams), requiredParams)) {
+  console.error('Usage: node tools/build-prod-env.js --db-password <DB_PASSWORD> --hostname <IP>')
+  process.exit(2)
+}
+
+const ENV_SAMPLE = fs.readFileSync(path.resolve(__dirname, '../.sample.env'), 'utf-8').split(os.EOL)
+fs.writeFileSync(path.resolve(__dirname, '../.env'), ENV_SAMPLE.join(os.EOL))
```
ubavic: Same as for `build-dev-env.js`

tools/build-prod-env.js
```diff
@@ -0,0 +1,47 @@
...
+
+setEnvVariable('POSTGRES_USER', 'lamassu_pg')
+setEnvVariable('POSTGRES_PASSWORD', `${argv['db-password']}`)
+setEnvVariable('POSTGRES_HOST', 'localhost')
```
ubavic, line 20: Same as for `build-dev-env.js`

bin/old-lamassu-register
```diff
@@ -1,10 +1,12 @@
...
+
 const login = require('../lib/admin/login')
-const options = require('../lib/options')
 
 const name = process.argv[2]
-const domain = options.hostname
+const domain = process.env.HOSTNAME
 
 if (!domain) {
   console.error('No hostname configured in lamassu.json')
```
ubavic: Maybe?
```suggestion
  console.error('No hostname configured in the environment.')
```

lib/ofac/index.js
```diff
@@ -4,23 +4,24 @@ const util = require('util')
...
+const OFAC_DATA_DIR = process.env.OFAC_DATA_DIR
+
 let structs = null
 
 const readdir = util.promisify(fs.readdir)
 
 function load () {
-  if (!options.ofacDataDir) {
+  if (!OFAC_DATA_DIR) {
     const message = 'The ofacDataDir option has not been set in lamassu.json'
```
ubavic: Maybe?

```suggestion
    const message = 'The ofacDataDir option has not been set in the environment.'
```

lib/ofac/update.js
```diff
@@ -97,13 +108,11 @@ const moveToSourcesDir = (srcFile, ofacSourcesDir) => {
 }
 
 function update () {
-  const OFAC_DATA_DIR = options.ofacDataDir
-
   if (!OFAC_DATA_DIR) {
     throw new Error('ofacDataDir must be defined in lamassu.json')
```
ubavic:
```suggestion
    throw new Error('ofacDataDir must be defined in the environment.')
```

lib/ofac/update.js
```diff
@@ -97,13 +108,11 @@ const moveToSourcesDir = (srcFile, ofacSourcesDir) => {
...
   }
 
-  if (!options.ofacSources) {
+  if (!ofacSources) {
     logger.error('ofacSources must be defined in lamassu.json')
```
ubavic:
```suggestion
    logger.error('ofacSources must be defined in the environment.')
```

lib/db.js
```diff
@@ -2,11 +2,18 @@ const Pgp = require('pg-promise')
...
 const { asyncLocalStorage, defaultStore } = require('./async-storage')
 
+const POSTGRES_USER = process.env.POSTGRES_USER
+const POSTGRES_PASSWORD = process.env.POSTGRES_PASSWORD
+const POSTGRES_HOST = process.env.POSTGRES_HOST
+const POSTGRES_PORT = process.env.POSTGRES_PORT
+const POSTGRES_DB = process.env.POSTGRES_DB
+
+const PSQL_URL = `postgres://${POSTGRES_USER}:${POSTGRES_PASSWORD}@${POSTGRES_HOST}:${POSTGRES_PORT}/${POSTGRES_DB}`
+
```
ubavic: We have same code in a few places atround (`notifications`, `logger`). 
Maybe we should add `PSQL_URL` to the environment (similar as was before)?

tools/cert-gen.sh
```diff
@@ -91,32 +91,6 @@ rm /tmp/Lamassu_OP.csr.pem
...
-      "name": "cons_advanced",
-      "url": "https://www.treasury.gov/ofac/downloads/sanctions/1.0/cons_advanced.xml"
-    }
-  ],
-  "idPhotoCardDir": "$IDPHOTOCARD_DIR",
-  "frontCameraDir": "$FRONTCAMERA_DIR",
-  "operatorDataDir": "$OPERATOR_DIR"
-}
-EOF
+node tools/create-dev-env.js
```
ubavic:
```suggestion
node tools/build-dev-env.js
```

ubavic (changes requested):

tools/build-prod-env.js
ubavic: This comment refers to `TESTVAR`

.sample.env
ubavic: Unix files are expected with a newline. Even if this probably isn't a problem, we should respect this to avoid potential problems in the future.

tools/build-prod-env.js
chaotixkilla: oops, nice catch

tools/build-dev-env.js
chaotixkilla: We're building the `.env` based on the sample `.env` and want to preserve the formatting (comments and overall organization)

tools/build-dev-env.js
chaotixkilla: No, at least not for now. The database will always be running on the local machine per default (as it currently is now on both dev and prod environments)

lib/db.js
chaotixkilla: Hmm, I think I prefer the idea to keep the variables separated. Added a constant to the constant files to avoid repeated code

ubavic (approved):

bin/convert-txs.js
```diff
@@ -1,9 +1,13 @@
 #!/usr/bin/env node
 
+const path = require('path')
+require('dotenv').config({ path: path.resolve(__dirname, '../.env') })
+
```
RafaelTaranto: Do we need the path here? Looks like we're importing the var from another file.

bin/lamassu-configure-frontcamera
```diff
@@ -2,11 +2,11 @@
 
 'use strict'
 
-const fs = require('fs')
-const options = require('../lib/options-loader')()
+const path = require('path')
+require('dotenv').config({ path: path.resolve(__dirname, '../.env') })
```
RafaelTaranto: This is a problem for production, isn't there a way of starting a server with the dotenv path pre-configured? 

The path for the .env is ~/.lamassu/.env for prod environments.

RafaelTaranto (changes requested):

PR #943, opened 2021-11-21
Fix transaction search filters

chaotixkilla:


chaotixkilla:
[Transaction filters not picking up cancelled transactions](https://trello.com/c/MVSP7370/1089-transaction-filters-not-picking-up-cancelled-transactions)

new-lamassu-admin/src/pages/Transactions/Transactions.js
```diff
@@ -114,6 +114,14 @@ const GET_TRANSACTIONS = gql`
...
 `
 
+const getFiltersObj = filters =>
+  R.compose(
+    R.mergeAll,
+    R.map(f => ({
+      [f.type]: f.value
+    }))
+  )(filters)
+
```
ubavic: Maybe like this?

```suggestion
const getFiltersObj = R.reduce((s, f) => ({...s, [f.type]: f.value}), {});
```

ubavic (approved):

PR #951, opened 2021-11-25
fix: max nr of cassette value fallback

josepfo:


josepfo:
[Initial wizard breaks upon choosing exchange](https://trello.com/c/NAtTloLY/1108-initial-wizard-breaks-upon-choosing-exchange)

ubavic (approved):

PR #953, opened 2021-11-25
Fix operator-id migration

chaotixkilla:


chaotixkilla:
[Issue with the operator-id migration, causing future queries to be skipped](https://trello.com/c/rdA0FCml/1115-issue-with-the-operator-id-migration-causing-future-queries-to-be-skipped)

PR #962, opened 2021-11-29
fix: list of customers in compliance is empty

ubavic:


ubavic:
[List of customers in compliance is empty](https://trello.com/c/bDahXuuO/993-list-of-customers-in-compliance-is-empty)

siiky (approved):

PR #966, opened 2021-11-30
fix: give up on this (now) noop

siiky:


siiky:
[Error in machine network performance migrations during update](https://trello.com/c/uIJsiz4J/1096-error-in-machine-network-performance-migrations-during-update)

josepfo (approved):

PR #968, opened 2021-12-01
Implement pazuz ticker

chaotixkilla:


chaotixkilla:
[Create the pazuz ticker plugin](https://trello.com/c/6jGKAvp6/671-create-the-pazuz-ticker-plugin)

tools/set-env-var.js
```diff
@@ -0,0 +1,24 @@
+const fs = require('fs')
+const os = require('os')
+const path = require('path')
+
+const setEnvVariable = (key, value) => {
+  const ENV_VARIABLES = fs.readFileSync(path.resolve(__dirname, '../.env'), 'utf-8').split(os.EOL)
+  const target = ENV_VARIABLES.indexOf(ENV_VARIABLES.find(line => line.match(new RegExp(`^${key}`))))
```
siiky: Using the regex `^${key}=` would better future proof this code. With the current regex, `foo` would also match `foobar`, which is wrong, and leads to the wrong result depending on the order of the keys in the file.

dev/notify.js
```diff
@@ -2,7 +2,7 @@ require('es6-promise').polyfill()
 
 var notifier = require('../lib/notifier')
 var db = require('../lib/postgresql_interface')
-var psqlUrl = require('../lib/options').postgres
+const { PSQL_URL } = require('./constants')
```
siiky: Did you mean `'../lib/constants'`?

lib/plugins/wallet/lnd/lnd.js
```diff
@@ -20,7 +21,7 @@ module.exports = {
 }
 
 function connect () {
-  return lnd.connect(options.lnd || {})
+  return lnd.connect(LND || {})
```
siiky: Is this right? `lnd.connect()` seems to expect an object, but values of environment variables are strings.

lib/plugins/wallet/monerod/monerod.js
```diff
@@ -5,14 +5,16 @@ const jsonRpc = require('../../common/json-rpc')
 
 const { COINS, utils } = require('lamassu-coins')
 
-const blockchainUtils = require('../../../coin-utils')
 const BN = require('../../../bn')
 const E = require('../../../error')
 const { logger } = require('../../../blockchain/common')
+const options = require('../../../options')
+
+const blockchainDir = options.blockchainDir
```
siiky: In the file `lib/blockchain/do-volume.js` you read the blockchain dir from the environment. Is that a different variable, or should this be read from the environment as well?

tools/build-prod-env.js
```diff
@@ -0,0 +1,46 @@
...
+
+const requiredParams = ['db-password', 'hostname']
+
+if (!_.isEqual(_.intersection(_.keys(argv), requiredParams), requiredParams)) {
+  console.error('Usage: node tools/build-prod-env.js --db-password <DB_PASSWORD> --hostname <IP>')
+  process.exit(2)
+}
+
+const ENV_SAMPLE = fs.readFileSync(path.resolve(__dirname, '../.sample.env'), 'utf-8').split(os.EOL)
+fs.writeFileSync(path.resolve(__dirname, '../.env'), ENV_SAMPLE.join(os.EOL))
```
siiky: Any special reason to read&write instead of using `fs.copyFile()` or `fs.copyFileSync()`?

siiky (changes requested):

siiky (approved):
LGTM

PR #970, opened 2021-12-03
fix: none 0conf

ubavic:


ubavic:
[Cash-out always resulting in a pending transaction](https://trello.com/c/SlzpsEzR/997-cash-out-always-resulting-in-a-pending-transaction)

siiky (approved):
LGTM

lib/wallet.js
```diff
@@ -175,6 +175,8 @@ function authorizeZeroConf (settings, tx, machineId) {
 
   if (plugin === 'all-zero-conf') return Promise.resolve(true)
 
+  if (plugin === 'none') return Promise.resolve(false)
```
RafaelTaranto: I don't think this is correct, if no zeroConf plugin is configured and tx.fiat < zeroConfLimit why would we wait for confirmation?

Also we don't use 'all-zero-conf' anymore, so I think this PR can be renaming 'all-zero-conf' to 'none'

RafaelTaranto (changes requested):

lib/wallet.js
siiky: Ah I misunderstood `tx.fiat.gt(zeroConfLimit)`... If there's no plugin, then the tx should be always accepted, no matter the limit, no? That's what I thought initially.

PR #971, opened 2021-12-03
fix: password confirmation step

siiky:


siiky:
[Password confirmation step is skipped on registration](https://trello.com/c/D0FE5j3T/1107-password-confirmation-step-is-skipped-on-registration)

new-lamassu-admin/src/pages/Authentication/Register.js
```diff
@@ -42,18 +42,17 @@ const REGISTER = gql`
...
-const validationSchema = Yup.object().shape({
+const validationSchema = Yup.object({
   password: Yup.string()
     .required('A password is required')
     .test(
       'len',
-      'Your password must contain more than 8 characters',
+      'Your password must contain at least 8 characters',
       val => val.length >= 8
     ),
```
ubavic: Suggestion: Instead using `test()` we could use `min()` method
```
.label("Password")
.min(8)
```

This will give *Password must be at least 8 characters* message.

For custom message we could write
```
 .min(8, "Your password must contain at least 8 characters")
```

siiky: Neat, didn't know that! Will change it.

siiky: I tried using `'Your password must contain at least ${min} characters'` as suggested by the docs for `string.min()` but prettier complained about the interpolation in a non interpolatable string.

ubavic (approved):

PR #983, opened 2021-12-07
Fix: confirm dialog title undefined on exit

ubavic:


new-lamassu-admin/src/components/machineActions/MachineActions.js
```diff
@@ -193,8 +192,8 @@ const MachineActions = memo(({ machine, onActionSuccess }) => {
       </div>
       <ConfirmDialog
         disabled={disabled}
-        open={confirmDialogOpen}
```
josepfo, line 196: Better force this as a boolean

josepfo (changes requested):

josepfo (approved):
LGTM

PR #994, opened 2021-12-13
Fix loading flag usage on admin

chaotixkilla:


new-lamassu-admin/src/pages/LoyaltyPanel/IndividualDiscounts.js
```diff
@@ -176,7 +178,7 @@ const IndividualDiscounts = () => {
           </Link>
         </Box>
       )}
-      {!isLoading && !R.isEmpty(discountResponse.individualDiscounts) && (
+      {!loading && !R.isEmpty(discountResponse.individualDiscounts) && (
```
siiky: What's the difference between this one and the one above?

siiky (changes requested):

siiky (approved):
LGTM

PR #1002, opened 2021-12-16
fix: unpairing machine through dashboard shows blank page

ubavic:


chaotixkilla (approved):

new-lamassu-admin/src/pages/Machines/Machines.js
```diff
@@ -50,13 +50,47 @@ const GET_INFO = gql`
...
+  useQuery(GET_MACHINES, {
+    onCompleted: data => {
+      const machines = data.machines
+      const machineFound = machines.map(m => m.deviceId).includes(id)
+
+      if (!machineFound) return history.push('/maintenance/machine-status')
+
+      setLoading(false)
+    }
+  })
```
RafaelTaranto: Doesn't use query return a loading flag already? Also, if you already have a query that fetches a machine, when you unpair it can't you refresh original query and if that returns error/empty you forward to another screen?

RafaelTaranto (changes requested):

new-lamassu-admin/src/pages/Machines/Machines.js
ubavic: It does return `loading`, but that flag is set to `false` before `onCompleted` is finished. This causes the same problems that we had before this fix.

Query `GET_INFO` fails if the machine id doesn’t exist in the database and that introduces the GraphQl error. I thought about using the `onError` hook but that seemed to me more dangerous. Even if I used that hook, this approach  would definitely cause errors on server side as query uses `getMachine` (from ‘machine-loader.js’) which breaks when the query returns an empty result. I could try to refactor function `getMachine` to be a safer.

RafaelTaranto: You can throw a error on the server if the machine doesn't exist and treat that error in specific during "onError". You can see custom errors in the authentication code, I think a "resource doesn't exist" error would be useful.

Maybe take some time searching what's the standard approach here, returning error or empty.

PR #1012, opened 2021-12-20
Add cashbox fiat amount

chaotixkilla:
feat: add cashin bill counter
fix: fiat amount formatting

### DIFF
--- a/lib/new-admin/graphql/resolvers/bill.resolver.js
+++ b/lib/new-admin/graphql/resolvers/bill.resolver.js
@@ -2,7 +2,9 @@ const bills = require('../../services/bills')
 
 const resolvers = {
   Query: {
-    bills: () => bills.getBills()
+    bills: () => bills.getBills(),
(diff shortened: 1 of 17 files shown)


ubavic (approved):

lib/new-admin/graphql/resolvers/bill.resolver.js
```diff
@@ -2,7 +2,9 @@ const bills = require('../../services/bills')
 
 const resolvers = {
   Query: {
-    bills: () => bills.getBills()
+    bills: () => bills.getBills(),
+    looseBills: () => bills.getLooseBills(),
+    looseBillsByMachine: (...[, { deviceId }]) => bills.getLooseBillsByMachine(deviceId)
```
RafaelTaranto: My feeling is that this should be the same graphql endpoint with different filters.

RafaelTaranto (changes requested):

PR #1019, opened 2021-12-23
Fix custom trigger filtering when in presence of older triggers

chaotixkilla:


new-lamassu-admin/src/pages/Triggers/helper.js
```diff
@@ -546,6 +568,12 @@ const Requirement = () => {
     (!values.requirement?.suspensionDays ||
       values.requirement?.suspensionDays < 0)
 
+  const hasCustomRequirementError =
+    !!errors.requirement &&
+    !!touched.requirement?.customInfoRequestId &&
+    (!values.requirement?.customInfoRequestId ||
+      !R.isNil(values.requirement?.customInfoRequestId))
```
josepfo: What do you think about export this and use it on Wizard.js instead of redefining it

josepfo (changes requested):

new-lamassu-admin/src/pages/Triggers/Wizard.js
```diff
@@ -13,6 +13,8 @@ import { singularOrPlural } from 'src/utils/string'
 
 import { type, requirements } from './helper'
 
+console.log('requirements', requirements)
```
josepfo: Forgot this one

josepfo (changes requested):
Looks good!

josepfo (approved):

PR #1029, opened 2022-01-03
Feat: twilio api data UI

josepfo:


new-lamassu-admin/src/pages/Customers/CustomerData.js
```diff
@@ -120,44 +131,56 @@ const CustomerData = ({
     }),
     frontCamera: Yup.object().shape({
       frontCamera: Yup.mixed().required()
-    })
+    }),
+    smsData: Yup.object()
+      .shape({
+        phoneNumber: Yup.string().required()
+      })
+      .required()
```
ubavic: Am I missing something, or this line can be removed?
```suggestion
```

new-lamassu-admin/src/pages/Customers/components/EditableCard.js
```diff
@@ -208,16 +250,26 @@ const EditableCard = ({
...
-                          {`Delete`}
-                        </ActionButton>
+                        {hasAdditionalData ? (
+                          <ActionButton
+                            color="primary"
+                            type="button"
+                            Icon={DeleteIcon}
+                            InverseIcon={DeleteReversedIcon}
+                            onClick={() => deleteEditedData()}>
+                            {`Delete`}
```
ubavic: I don't know if this this was deliberate choice, but this could be simpler.
```suggestion
                            Delete
```
(There are similar lines in this file)

new-lamassu-admin/src/pages/Customers/CustomerData.js
```diff
@@ -374,6 +424,22 @@ const CustomerData = ({
...
+      label: onlyFirstToUpper(it),
+      component: TextInput,
+      editable: true
+    })
+    Yup.object()
+      .shape({
+        [it]: Yup.string()
+      })
+      .required()
+      .concat(schemas.smsData)
```
ubavic: Maybe I am wrong, but this doesn't do anything? (we don't assign created object to variable)

ubavic (changes requested):

new-lamassu-admin/src/pages/Customers/CustomerData.js
josepfo: I believe we need this .required() to use concat() further down, otherwise, I think Yup's type definition will add undefined as an intersection type since we didn't specify the given object would always be not-null.

new-lamassu-admin/src/pages/Customers/CustomerData.js
josepfo: Ah yeah, forgot to assign it to the schemas object

lib/customers.js
```diff
@@ -117,7 +117,8 @@ async function updateCustomer (id, data, userToken) {
       'us_ssn_override',
       'sanctions_override',
       'front_camera_override',
-      'suspended_until'
+      'suspended_until',
+      'subscriber_info_override'
```
RafaelTaranto: We don't really need override for subscriber info. It's not customer submitted data, so there's no point in approving or rejecting it.

RafaelTaranto (changes requested):

ubavic (approved):

migrations/1641482376890-add-overrides-to-subscriber-info.js
```diff
@@ -0,0 +1,22 @@
...
+    `ALTER TABLE customers 
+     ADD COLUMN phone_override VERIFICATION_TYPE NOT NULL DEFAULT 'automatic',
+     ADD COLUMN phone_override_by UUID,
+     ADD COLUMN phone_override_at TIMESTAMPTZ
+    `,
+    `ALTER TABLE edited_customer_data
+     ADD COLUMN phone TEXT,
+     ADD COLUMN phone_at TIMESTAMPTZ,
+     ADD COLUMN phone_by UUID
+    `
```
RafaelTaranto: Isn't phone number the one field we can't edit?

RafaelTaranto (changes requested):

migrations/1641482376890-add-overrides-to-subscriber-info.js
josepfo: yes, but it's disabled on the admin level. I left this logic here since we'll use it in the future, right?

RafaelTaranto: I don't think we will. Also, if we do that for the phone it might have another structure in place, since that's kinda the ID for the customers right now.

PR #1032, opened 2022-01-05
Add testing customer toggle

chaotixkilla:


migrations/1641394367865-testing-customer-toggle.js
```diff
@@ -0,0 +1,13 @@
+var db = require('./db')
+
+exports.up = function (next) {
+  var sql = [
+    `ALTER TABLE customers ADD COLUMN is_test_customer BOOLEAN DEFAULT false`,
```
ubavic: I would add `NOT NULL` constraint.

ubavic (approved):

PR #1043, opened 2022-01-13
fix: ID expiration date in transcations

ubavic:


josepfo (approved):

new-lamassu-admin/src/pages/Transactions/DetailsCard.js
```diff
@@ -123,9 +123,8 @@ const DetailsRow = ({ it: tx, timezone }) => {
     age: differenceInYears(tx.customerIdCardData.dateOfBirth, new Date()),
     country: tx.customerIdCardData.country,
     idCardNumber: tx.customerIdCardData.documentNumber,
-    idCardExpirationDate: format(
-      'dd-MM-yyyy',
-      tx.customerIdCardData.expirationDate
+    idCardExpirationDate: format('yyyy-MM-dd')(
+      parse(new Date(), 'yyyyMMdd', tx.customerIdCardData.expirationDate)
```
RafaelTaranto: Hmm, what happens here is that we default to current time when expirationDate is null? I don't think we should do that, we'll be showing incorrect data to the operator.

RafaelTaranto (changes requested):

PR #1052, opened 2022-01-17
Fix config updates not propagating to machine without restart

chaotixkilla:


josepfo (approved):

lib/poller.js
```diff
@@ -84,52 +84,28 @@ db.connect({ direct: true }).then(sco => {
     switch (parsedData.type) {
       case 'reload':
         return reload(parsedData.schema)
-      case 'machineAction':
-        return machineAction(parsedData.action, parsedData.value)
       default:
         break
     }
   })
```
RafaelTaranto: I don't think we need the type nor the test anymore.

lib/new-settings-loader.js
```diff
@@ -71,12 +72,12 @@ function showAccounts (schemaVersion) {
...
 function saveConfig (config) {
-  return loadLatestConfigOrNone()
-    .then(currentConfig => {
+  return Promise.all([loadLatestConfigOrNone(), getOperatorId('middleware')])
+    .then(([currentConfig, operatorId]) => {
       const newConfig = _.assign(currentConfig, config)
       return db.tx(t => {
         return t.none(configSql, ['config', { config: newConfig }, true, NEW_SETTINGS_LOADER_SCHEMA_VERSION])
-          .then(() => t.none('NOTIFY $1:name, $2', ['poller', JSON.stringify({ type: 'reload', schema: asyncLocalStorage.getStore().get('schema') })]))
+          .then(() => t.none('NOTIFY $1:name, $2', ['reload', JSON.stringify({ type: 'reload', schema: asyncLocalStorage.getStore().get('schema'), operatorId })]))
```
RafaelTaranto: No need for type.

lib/middlewares/populateSettings.js
```diff
@@ -1,11 +1,66 @@
...
+
+db.connect({ direct: true }).then(sco => {
+  sco.client.on('notification', data => {
+    const parsedData = JSON.parse(data.payload)
+    switch (parsedData.type) {
+      case 'machineAction':
+        return machineAction(parsedData.action, parsedData.value)
+      default:
+        break
+    }
```
RafaelTaranto: Same as above, no need for type or the switch anymore.

RafaelTaranto (changes requested):

PR #1057, opened 2022-01-19
feat: set custom info requirement live values

josepfo:


new-lamassu-admin/src/pages/Customers/helper.js
```diff
@@ -271,21 +296,43 @@ const ManualDataEntry = ({ selectedValues, customInfoRequirementOptions }) => {
...
+            options={customInfoRequirementOptions}
+          />
+          {(customInfoRequirementType === 'text' ||
+            customInfoRequirementType === 'numerical') && (
+            <div>{console.log('')}</div>
+          )}
+          {customInfoRequirementType === 'choiceList' && (
+            <div>
+              <div>{console.log('')}</div>
+            </div>
```
RafaelTaranto, line 333: Leftovers?

RafaelTaranto (changes requested):

PR #1063, opened 2022-01-21
fix: save config used on migration

josepfo:


ubavic (approved):

lib/new-settings-loader.js
```diff
@@ -82,6 +82,17 @@ function saveConfig (config) {
     })
 }
 
+function migrationSaveConfig (config) {
+  return loadLatestConfigOrNone()
+    .then(currentConfig => {
+      const newConfig = _.assign(currentConfig, config)
+      return db.tx(t => {
+        return t.none(configSql, ['config', { config: newConfig }, true, NEW_SETTINGS_LOADER_SCHEMA_VERSION])
+          .then(() => t.none('NOTIFY $1:name, $2', ['poller', JSON.stringify({ type: 'reload', schema: asyncLocalStorage.getStore().get('schema') })]))
```
RafaelTaranto: Migrations don't need the notify part of the function

RafaelTaranto (changes requested):

PR #1075, opened 2022-01-27
fix: prevent bitcoind v22 from trodding on bch port

CrypticaScriptura:
https://github.com/lamassu/lamassu-coins/pull/2


siiky (approved):
LGTM

PR #1076, opened 2022-01-27
fix: custom info requests flow

siiky:


lib/customers.js
```diff
@@ -723,15 +721,18 @@ function getCustomerById (id) {
...
+    .then(assignCustomerData)
     .then(populateOverrideUsernames)
     .then(camelize)
     .then(it => ({ ...it, notes: (it.notes ?? []).map(camelize) }))
 }
 
+function assignCustomerData (customer) {
+  return getEditedData(customer.id)
+    .then(customerEditedData => selectLatestData(customer, customerEditedData))
+    .then(_.assign(customer))
```
josepfo: I don't see why this is necessary, selectLatestData should preserve all the data and only update the required fields.

lib/routes/phoneCodeRoutes.js
```diff
@@ -26,8 +26,7 @@ function addOrUpdateCustomer (req) {
 
       return add(req.body)
     })
-    .then(customer => Promise.all([getEditedData(customer.id), getCustomerById(customer.id)]))
-    .then(([customerEditedData, customerOriginalData]) => selectLatestData(customerOriginalData, customerEditedData))
+    .then(customer => getById(customer.id))
     .then(customer => {
```
josepfo, line 30: I added this logic recently, `addOrUpdateCustomer` provides customer info to the machine, the goal  was to update the customer data with edited or manually entered data which is stored in a specific table. Removing this would ignore that data.
However, If this `getById` is part of the solution for this card, maybe it has some computation that we could reuse in `get` and `add`

josepfo (changes requested):

lib/customers.js
siiky: At the time, before I added the `_.assign()`, the data from the `edited_customer_data` table wasn't sent to the machine even though it was fetched from the DB. Everything started working after I added it, but looking again now, I also agree it doesn't look like it's doing anything... I'll check!

lib/routes/phoneCodeRoutes.js
siiky: Yup, the changes I made to `getById()` include getting that manually entered customer data -- it's the `assignCustomerData()`.

Maybe we can refactor and simplify this further? I'll look into it too.

lib/customers.js
siiky: Indeed, looks like the `_.assign()` is unnecessary.

lib/routes/phoneCodeRoutes.js
siiky: Given our discussion, what do you think is the best thing to do?

---

@RafaelTaranto said that `getById()` is the right function to get customer info for the machine. However, we were using `getCustomerById()` previously and the machine was working fine -- except for the custom info requests; `getCustomInfoRequestsData()` should be enough for that.

Plus `getById()` adds info that is seemingly unnecessary to the machine (because `getCustomerById()` doesn't and the machine still worked fine).

Should I try to cut this bit of unnecessary server-machine traffic?

josepfo (approved):

josepfo (approved):

PR #1083, opened 2022-02-01
fix: l-c import on lamassu-send-coins script

josepfo:


siiky (approved):
Just wonder if the interface didn't change after moving to `lamassu-coins`...

PR #1092, opened 2022-02-04
fix: update only if node is stopped

josepfo:


bin/lamassu-update-wallet-nodes
```diff
@@ -28,8 +28,7 @@ function run () {
     const cryptoPlugin = plugin(crypto)
     const status = common.es(`sudo supervisorctl status ${crypto.code} | awk '{ print $2 }'`).trim()
 
-    if (status === 'RUNNING') cryptoPlugin.updateCore(common.getBinaries(crypto.cryptoCode), true)
-    if (status === 'STOPPED') cryptoPlugin.updateCore(common.getBinaries(crypto.cryptoCode), false)
+    cryptoPlugin.updateCore(common.getBinaries(crypto.cryptoCode), status === 'RUNNING')
```
siiky: Do you think it's worth checking also for the `STARTING` status? When I rummaged through this code last time I didn't know about it. WDYT?

http://supervisord.org/subprocess.html#process-states

siiky (approved):

bin/lamassu-update-wallet-nodes
josepfo: yeah, I think we should check for that as well

siiky (approved):

PR #1099, opened 2022-02-10
Get SMS notices closer to spec

chaotixkilla:


new-lamassu-admin/src/pages/OperatorInfo/SMSNotices/SMSNotices.js
```diff
@@ -0,0 +1,271 @@
...
+        />
+      )}
+      <DeleteDialog
+        open={deleteDialog}
+        onDismissed={() => {
+          handleClose()
+        }}
+        onConfirmed={() => {
+          handleClose()
+        }}
```
siiky: Isn't this to delete some custom SMS template? If so, isn't a `disableSMSNotice()` missing?

siiky (changes requested):

new-lamassu-admin/src/pages/OperatorInfo/SMSNotices/SMSNotices.js
chaotixkilla: Oh, leaving this was an oversight on my end. Removed it

About adding a disableSMSNotice() behavior, I don't think that's needed

siiky (approved):

PR #1110, opened 2022-02-14
Fix XMR cash-out address balance lookup

chaotixkilla:


lib/plugins/wallet/monerod/monerod.js
```diff
@@ -175,12 +156,20 @@ function newFunding (account, cryptoCode, settings, operatorId) {
...
-      fundingConfirmedBalance: BN(balanceRes.unlocked_balance),
-      fundingAddress: addressRes.address
-    }))
-    .catch(err => handleError(err))
+    .then(([balanceRes, addressRes]) => Promise.all([
+        balanceRes,
+        addressRes,
+        fetch('get_transfers', { pool: true, account_index: 0 })
+      ]))
+    .then(([balanceRes, addressRes, transferRes]) => {
```
siiky, line 164: Any particular reason to split the 3 calls in 3 different `.then()`s? They look independent of each other to me. Feel free to ignore if you don't think it's worth the effort.

lib/plugins/wallet/monerod/monerod.js
```diff
@@ -175,12 +156,20 @@ function newFunding (account, cryptoCode, settings, operatorId) {
...
+      const memPoolBalance = _.reduce((acc, value) => acc.plus(value.amount), BN(0), transferRes.pool)
+      return {
+        fundingPendingBalance: BN(balanceRes.balance).minus(balanceRes.unlocked_balance).plus(memPoolBalance),
```
siiky, line 167: The docs aren't very clear about this, but doesn't `balance` include the amounts of the txs of the pool? I.e., doesn't this line account for them twice?

Description of the `balance` of the output of [`get_balance`](https://www.getmonero.org/resources/developer-guides/wallet-rpc.html#get_balance):

> The total balance of the current monero-wallet-rpc in session.

And description of `pool` of [`get_transfers`](https://www.getmonero.org/resources/developer-guides/wallet-rpc.html#get_transfers):

> Include transfers from the daemon's transaction pool.

siiky (changes requested):

siiky (approved):

PR #1137, opened 2022-02-28
Fix ERC-20 transfer gas estimation and receiver address

chaotixkilla:


lib/plugins/wallet/geth/base.js
```diff
@@ -140,7 +143,7 @@ function generateTx (_toAddress, wallet, amount, includesFee, cryptoCode) {
       if (isErc20Token) {
         const contract = web3.eth.contract(ABI.ERC20).at(toAddress)
```
siiky: So this contract isn't necessary for anything? I don't know much about anything, but this is weird.

I have to read about the matter some day.

siiky (approved):

lib/plugins/wallet/geth/base.js
chaotixkilla: Just an error on my end, forgot to move the contract creation upwards

PR #1157, opened 2022-03-10
feat: lightning coin support

josepfo:


lib/new-admin/services/funding.js
```diff
@@ -57,7 +57,7 @@ const reflect = p => p.then(value => ({ value, status: 'fulfilled' }), error =>
 
 function getFunding () {
   return settingsLoader.loadLatest().then(settings => {
-    const cryptoCodes = configManager.getAllCryptoCurrencies(settings.config)
+    const cryptoCodes = _.filter(code => coinUtils.getExternalCryptoCode(code) === code)(configManager.getAllCryptoCurrencies(settings.config))
```
siiky: Nitpick: unnecessary parens: `s/)(/, /`

lib/plugins/wallet/galoy/galoy.js
```diff
@@ -0,0 +1,364 @@
...
+  }
+  return Promise.resolve(true)
+    .then(() => {
+      return axios({
+        method: 'post',
+        url: URI,
+        headers: headers,
+        data: graphqlQuery
+      })
+    })
```
siiky: Nitpick: unnecessary `Promise.resolve`. You can just chain the `then()` and `catch()` to the `axios()`:

```js
  return axios({
    method: 'post',
    url: URI,
    headers: headers,
    data: graphqlQuery
  })
  .then(...)
  .catch(...)
```

lib/plugins/wallet/galoy/galoy.js
```diff
@@ -0,0 +1,364 @@
...
+              }
+            }
+          }
+        }
+        id
+        phone
+        twoFAEnabled
+        username
+      }
+    }`,
```
siiky: Why request fields you don't want (`createdAt`, `id`, `phone`, `twoFAEnabled`, `username`)? It happens with a few of the other queries too.

lib/plugins/wallet/galoy/galoy.js
```diff
@@ -0,0 +1,364 @@
...
+    'variables': { 'input': { 'address': `${address}`, 'amount': `${cryptoAtoms}`, 'walletId': `${walletId}` } }
+  }
+  return request(sendOnChain, token)
+    .then(result => {
+      return result.data.onChainPaymentSend
+    })
+    .catch(err => {
+      throw err
+    })
```
siiky: Isn't this the same as not `catch()`ing in the first place? There are some other instances of this.

lib/plugins/wallet/galoy/galoy.js
```diff
@@ -0,0 +1,364 @@
...
+    .then(() => getGaloyAccount(account.apiKey))
+    .then(galoyAccount => {
+      const wallet = _.head(
+        _.filter(wallet => wallet.walletCurrency === externalCryptoCode &&
+          wallet.id === galoyAccount.defaultWalletId &&
+          wallet.id === account.walletId)(galoyAccount.wallets)
+      )
+      const transactions = wallet.transactions.edges
+      if (isLightning(address)) {
+        const paymentHash = invoice.decode(address).paymentHash.toString('hex')
```
siiky, line 282: Doesn't this `invoice.decode()` ever fail?

siiky (changes requested):

lib/plugins/wallet/galoy/galoy.js
josepfo: Some leftovers from testing

lib/plugins/wallet/galoy/galoy.js
josepfo: Initially wanted to throw more informative errors but ended up leaving this as is

lib/plugins/wallet/galoy/galoy.js
josepfo: yes, in that case, the function will throw an error

siiky (approved):

PR #1164, opened 2022-03-16
feat: use GraphQL for all server-machine communication

siiky:
Related to https://github.com/lamassu/lamassu-machine/pull/773


lib/plugins.js
```diff
@@ -365,12 +439,12 @@ function plugins (settings, deviceId) {
 
         const rate = rawRate.div(cashInCommission)
 
-        const lowBalanceMargin = new BN(1.05)
+        const lowBalanceMargin = new BN(0.95)
```
chaotixkilla, line 368: Any specific reason for this change?

lib/routes.js
```diff
@@ -56,10 +60,13 @@ app.use('/', pairingRoutes)
 app.use(findOperatorId)
 app.use(populateDeviceId)
 app.use(computeSchema)
-if (!devMode) app.use(authorize)
+app.use(authorize)
 app.use(configRequiredRoutes, populateSettings)
 app.use(filterOldRequests)
 
+app.use('/graphql', recordPing)
+graphQLServer.applyMiddleware({ app })
```
chaotixkilla: Does the order matter here? I'd personally prefer the routes to be grouped up together, but it's a nitpick

chaotixkilla (approved):

lib/plugins.js
siiky: Yeah, taking 5% off is multiplying by 0.95, not dividing by 1.05.

lib/routes.js
siiky: Hmm not sure, I'll try.

siiky: Seems to work.

chaotixkilla (approved):

lib/graphql/resolvers.js
```diff
@@ -0,0 +1,156 @@
...
+      _.set('restartServices', !!pid && state.restartServicesMap?.[operatorId]?.[deviceId] === pid),
+    ))
+}
+
+
+module.exports = {
+  Query: {
+    staticConfig,
+    dynamicConfig
+  }
```
RafaelTaranto: I think I would rather have a single query that returns both static and dynamic config. If you organize that way it means that you can run pollQueries before and pass it down as a parameter.

RafaelTaranto (changes requested):

PR #1175, opened 2022-03-29
Fix BTC wallet creation

chaotixkilla:


lib/plugins/wallet/bitcoind/bitcoind.js
```diff
@@ -17,21 +17,49 @@ function fetch (method, params) {
...
+    case -5:
+      return logger.error(`${err}`)
+    case -6:
+      throw new E.InsufficientFundsError()
+    case -18:
+      return createWallet()
+    default:
+      throw e
+  }
+}
```
josepfo: Won't this return unexpected results? Let's say we're running `sendCoins` and there's no wallet, when handling the error we will return the `createWallet` output, right?

josepfo (changes requested):

lib/plugins/wallet/bitcoind/bitcoind.js
```diff
@@ -156,6 +180,7 @@ function newFunding (account, cryptoCode, settings, operatorId) {
       fundingConfirmedBalance,
       fundingAddress
     }))
+    .catch(console.log)
```
josepfo:
```suggestion
    .catch(errorHandle)
```
Should we add this on all `newFunding`?

josepfo (approved):

josepfo:
LGTM!

PR #1180, opened 2022-03-31
Fix customer ID not being updated on 0-conf turned 1-conf cash-out transactions

chaotixkilla:


siiky (approved):

lib/cash-out/cash-out-low.js
```diff
@@ -8,7 +8,7 @@ const toObj = helper.toObj
 
 const UPDATEABLE_FIELDS = ['txHash', 'txVersion', 'status', 'dispense', 'dispenseConfirmed',
   'notified', 'redeem', 'phone', 'error', 'swept', 'publishedAt', 'confirmedAt', 'errorCode',
-  'receivedCryptoAtoms', 'walletScore' ]
+  'receivedCryptoAtoms', 'walletScore', 'customerId' ]
```
RafaelTaranto, line 12: If we wanna update the customerID I would prefer if we add a helper function that checks and only allows change if the old ID is equal to anonymous customer.

RafaelTaranto (changes requested):

PR #1186, opened 2022-04-07
Allow for deletion of testing customers

chaotixkilla:


lib/customers.js
```diff
@@ -909,6 +913,11 @@ function disableTestCustomer (customerId) {
   return db.none(sql, [customerId])
 }
 
+function deleteCustomer (customerId) {
+  const sql = `UPDATE customers SET enabled=false WHERE id=$1`
```
siiky: `AND is_test_customer` missing?

migrations/1649353351891-delete-test-customers.js
```diff
@@ -0,0 +1,15 @@
+var db = require('./db')
+
+exports.up = function (next) {
+  var sql = [
+    `ALTER TABLE customers ADD COLUMN enabled BOOLEAN NOT NULL DEFAULT true`,
```
siiky, line 5: Does this set the `enabled` column of existing rows to true too?

lib/customers.js
```diff
@@ -514,10 +515,11 @@ function getCustomersList (phone = null, name = null, address = null, id = null)
       ) cn ON c.id = cn.customer_id
     WHERE c.id != $2
   ) AS cl WHERE rn = 1
+  AND enabled
```
siiky: Can you use `c.enabled` instead? Just so that it's obvious where `enabled` comes from.

siiky (changes requested):

siiky (approved):
LGTM!

PR #1188, opened 2022-04-14
feat: cache the T&C based on their hash

siiky:
To be merged after https://github.com/lamassu/lamassu-server/pull/1164. Related: https://github.com/lamassu/lamassu-machine/pull/790.

### DIFF
--- a/lib/graphql/resolvers.js
+++ b/lib/graphql/resolvers.js
@@ -68,18 +68,6 @@ const buildTriggers = (allTriggers) => {
     })
 }
 
-/*
- * TODO: From `lib/routes/termsAndConditionsRoutes.js` -- remove this after
- * terms are removed from the GraphQL API too.
- */
-const massageTerms = terms => (terms.active && terms.text) ? ({
-  delay: Boolean(terms.delay),
-  title: terms.title,
-  text: nmd(terms.text),
-  accept: terms.acceptButtonText,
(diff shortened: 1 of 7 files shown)


lib/graphql/resolvers.js
```diff
@@ -1,3 +1,5 @@
+const crypto = require('crypto')
```
ubavic: This seems unnecessary

migrations/1649944954805-terms-and-conditions-hash.js
```diff
@@ -0,0 +1,11 @@
+const { saveConfig, loadLatest } = require('../lib/new-settings-loader')
```
ubavic: `loadLatest` is unnecessary

lib/new-settings-loader.js
```diff
@@ -23,6 +26,30 @@ const SECRET_FIELDS = [
...
 ]
 
+/*
+ * JSON.stringify isn't necessarily deterministic so this function may compute
+ * different hashes for the same object.
+ */
+const md5hash = obj =>
+  crypto
+    .createHash('MD5')
+    .update(JSON.stringify(obj))
```
RafaelTaranto: We just need the content hash, so you don't have stringify for the rest of the settings.

RafaelTaranto (changes requested):

PR #1193, opened 2022-04-19
Save CCXT ticker objects

siiky:


chaotixkilla (approved):

lib/plugins/ticker/ccxt.js
```diff
@@ -1,49 +1,57 @@
...
+const tickerObjects = {}
+
+function ticker (fiatCode, cryptoCode, tickerName) {
+  const ticker = tickerObjects[tickerName] ?
+    tickerObjects[tickerName] :
+    tickerObjects[tickerName] = new ccxt[tickerName]({
+      timeout: 3000,
+      enableRateLimit: false,
+    })
+
```
RafaelTaranto: I think the following is a more readable standard for doing this.

```js
  if (!tickerObjects[tickerName]) {
    tickerObjects[tickerName] = new ccxt[tickerName]({
      timeout: 3000,
      enableRateLimit: false,
    })
  }

  const ticker = tickerObjects[tickerName]

```

RafaelTaranto (changes requested):

PR #1201, opened 2022-04-29
Fix: coinatmradar 400 error

josepfo:


siiky (approved):

lib/coinatmradar/coinatmradar.js
```diff
@@ -133,7 +133,7 @@ function sendRadar (data) {
 function mapRecord (rates, settings) {
   const timestamp = new Date().toISOString()
   return Promise.all([getMachines(rates, settings), getOperatorId('coinatmradar')])
-    .then(([machines, operatorId]) => {
+    .then(([machines, { operatorId }]) => {
```
RafaelTaranto: Can't you do this on the "getOperatorId" function? it's unexpected to ask for a ID and receive a object in response.

RafaelTaranto (changes requested):

lib/coinatmradar/coinatmradar.js
RafaelTaranto: Also do take a look at other places using this function to see if they are using it correctly

PR #1247, opened 2022-05-31
Improve Infura network usage

chaotixkilla:


josepfo (commented):
Looks good! Take a look at the wallet `coinFilter` instance, we probably can get rid of it.

josepfo (approved):

lib/plugins/wallet/infura/infura.js
```diff
@@ -12,4 +16,60 @@ function run (account) {
...
+setInterval(() => {
+  _.forEach(it => {
+    // A transaction not being updated for over an hour means that it stopped being fetched on poll,
+    // which means that it probably was cancelled by the operator or something else caused the
+    // transaction to stop being listened to. So, trim the cache to save memory
+    const lastReq = _.isNil(txsCache[it].lastReqTime) ? Date.now() : new Date(txsCache[it].lastReqTime)
+    const timePassedSinceReq = Date.now() - lastReq
+    if (timePassedSinceReq > T.hour) delete txsCache[it]
+  }, _.keys(txsCache))
+}, T.minute)
```
RafaelTaranto: We can use node-cache here. It's already being used in other places of our codebase.

RafaelTaranto (changes requested):

PR #1253, opened 2022-06-09
Add script to sweep balance to a new ETH wallet

chaotixkilla:


bin/lamassu-eth-transfer-pending
```diff
@@ -0,0 +1,128 @@
...
+  })
+  .then(([balance, settings]) => {
+    const tx = {
+      cryptoCode: 'ETH',
+      toAddress: newAddress,
+      cryptoAtoms: balance
+    }
+
+    const opts = {
+      chainId: 3,
```
josepfo:
```suggestion
      chainId: 1,
```

lib/plugins/wallet/geth/base.js
```diff
@@ -115,29 +117,37 @@ function generateTx (_toAddress, wallet, amount, includesFee, cryptoCode) {
...
         : amount
 
+      const maxPriorityFeePerGas = new BN(web3.utils.toWei('2.5', 'gwei')) // web3 default value
+      const maxFeePerGas = new BN(2).times(baseFeePerGas).plus(maxPriorityFeePerGas)
+
       const rawTx = {
-        chainId: 1,
-        nonce: txCount,
-        gasPrice: hex(gasPrice),
+        chainId: _.defaultTo(3, opts?.chainId),
```
josepfo:
```suggestion
        chainId: _.defaultTo(1, opts?.chainId),
```

lib/plugins/wallet/geth/base.js
```diff
@@ -115,29 +117,37 @@ function generateTx (_toAddress, wallet, amount, includesFee, cryptoCode) {
 
   if (isErc20Token) txTemplate.data = contractData
 
+  const common = new Common({ chain: _.defaultTo(Chain.Ropsten, opts?.chainId), hardfork: Hardfork.London })
```
josepfo:
```suggestion
  const common = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.London })
```

Why do we use _.defaultTo here? I don't think the Common constructor will accept anything other than strings or enums on the chain field.

josepfo (changes requested):

bin/lamassu-eth-transfer-pending
```diff
@@ -0,0 +1,128 @@
...
+const fs = pify(require('fs'))
+const _ = require('lodash/fp')
+
+const mnemonicHelpers = require('../lib/mnemonic-helpers')
+const options = require('../lib/options')
+const settingsLoader = require('../lib/new-settings-loader')
+const wallet = require('../lib/wallet')
+
+const defaultPrefixPath = "m/44'/60'/1'/0'"
+const paymentPrefixPath = "m/44'/60'/0'/0'"
```
josepfo:
```suggestion
```

This can be removed, right?

lib/plugins/wallet/geth/base.js
chaotixkilla: It's an enum that corresponds to the chain ID. So far, it's working great, should it be removed?

PR #1254, opened 2022-06-09
Merge release 8.0 into dev 220906

RafaelTaranto:


new-lamassu-admin/src/pages/Wallet/Wallet.js
```diff
@@ -61,6 +61,8 @@ const Wallet = ({ name: SCREEN_KEY }) => {
   const [advancedSettings, setAdvancedSettings] = useState(false)
   const { data } = useQuery(GET_INFO)
 
+  console.log(data)
```
josepfo: There's a console.log here

lib/plugins/wallet/geth/base.js
```diff
@@ -125,11 +124,10 @@ function generateTx (_toAddress, wallet, amount, includesFee, cryptoCode) {
   ]
 
   return Promise.all(promises)
-    .then(([gas, gasPrice, txCount, { baseFeePerGas }]) => [
```
josepfo, line 128: `baseFeePerGas` is removed here but there are more EIP-1559 leftovers. Should we leave it as is and handle this later?

RafaelTaranto: EIP leftovers will come with the latest PR from sergio, I'll merge back again into dev when it happens.

PR #1259, opened 2022-06-14
Refactor getCustomer queries

chaotixkilla:


siiky (approved):

siiky (changes requested):
Almost forgot: I suppose the `console.log`s aren't supposed to stay?

siiky (approved):

lib/customers.js
```diff
@@ -238,13 +238,13 @@ function deleteEditedData (id, data) {
...
   const filteredData = _.pick(defaults, _.mapKeys(_.snakeCase, data))
-  if (_.isEmpty(filteredData)) return getCustomerById(id)
+  if (_.isEmpty(filteredData)) return getCustomers({ id })
 
   const cs = new Pgp.helpers.ColumnSet(_.keys(filteredData),
     { table: 'edited_customer_data' })
   const update = Pgp.helpers.update(filteredData, cs)
   db.none(update)
-  return getCustomerById(id)
+  return getCustomers({ id })
```
RafaelTaranto: Let's leave everything as `getCustomerById()`, move this logic into the function instead of copy pasting everywhere.

RafaelTaranto (changes requested):

PR #1260, opened 2022-06-14
[Release] Refactor getCustomer queries

chaotixkilla:


siiky (changes requested):
I suppose the `console.log`s aren't supposed to stay?

siiky (approved):

PR #1262, opened 2022-06-16
Fix ERC-20 token transaction builder

chaotixkilla:


lib/plugins/wallet/geth/base.js
```diff
@@ -52,6 +52,23 @@ function isStrictAddress (cryptoCode, toAddress, settings, operatorId) {
...
+        return pify(web3.eth.getTransaction)(txid)
+          .then(tx => {
+            if (!tx) return { txid }
+
+            const fee = new BN(tx.gas).times(new BN(tx.gasPrice)).decimalPlaces(0)
+
+            return { txid, fee }
+          })
+      })
+  }
```
siiky: This looks like an exact copy of the code below; I think it would be better to call either `generateErc20Tx()` or `generateTx()` and then run this bit of code on the result of either.

```js
return (isErc20Token ? generateErc20Tx : generateTx)(toAddress, ...)
  .then(...)
```

siiky (changes requested):

siiky (approved):

PR #1263, opened 2022-06-20
Add BTC regtest support

chaotixkilla:


lib/blockchain/bitcoin.js
```diff
@@ -8,47 +8,52 @@ module.exports = { setup, updateCore }
 
 const coinRec = coinUtils.getCryptoCurrency('BTC')
 
+const BLOCKCHAIN_DIR = process.env.BLOCKCHAIN_DIR
+
+const tmpDir = process.env.NODE_ENV === 'development' ? path.resolve(BLOCKCHAIN_DIR, 'tmp') : '/tmp'
```
josepfo: what do you think about encapsulating this condition `process.env.NODE_ENV === 'development'` inside a function `const isDevMode = mode => mode === 'development'` and make it accessible throughout the code?

lib/plugins/wallet/bitcoind/bitcoind.js
```diff
@@ -36,15 +36,15 @@ function checkCryptoCode (cryptoCode) {
...
+    .then(({ mine }) => new BN(mine.trusted).shiftedBy(unitScale).decimalPlaces(0))
     .catch(errorHandle)
 }
 
 function accountUnconfirmedBalance (cryptoCode) {
   return checkCryptoCode(cryptoCode)
-    .then(() => fetch('getwalletinfo'))
-    .then(({ unconfirmed_balance: balance }) => new BN(balance).shiftedBy(unitScale).decimalPlaces(0))
+    .then(() => fetch('getbalances'))
+    .then(({ mine }) => new BN(mine.untrusted_pending).plus(mine.immature).shiftedBy(unitScale).decimalPlaces(0))
```
josepfo, line 48: Are we sure we want to add the immature here?

lib/plugins/wallet/bitcoind/bitcoind.js
```diff
@@ -62,7 +62,7 @@ function estimateFee () {
 
 function calculateFeeDiscount (feeMultiplier) {
   // 0 makes bitcoind do automatic fee selection
-  const AUTOMATIC_FEE = 0
+  const AUTOMATIC_FEE = process.env.NODE_ENV === 'development' ? 0.01 : 0
```
josepfo: Regtest transactions won't go through without this?

josepfo (changes requested):

lib/plugins/wallet/bitcoind/bitcoind.js
chaotixkilla: Nope, just because there's not enough transaction volume to let the blockchain automatically decide how much fee to allocate. So it's easier to just define a static fee value on regtest environments

lib/plugins/wallet/bitcoind/bitcoind.js
chaotixkilla: I believe this was overlooked on our end. Despite no one actually using the nodes for mining if, in the future and for some reason, some operator decides to use a node to operate as the lamassu node and a mining node at the same time, our pending balance was not taking into account balance from coinbase rewards (which is the immature balance there). I can chat about it with Rafael to get a stance on this

josepfo: Okay, tagging @RafaelTaranto

josepfo (approved):

lib/blockchain/install.js
```diff
@@ -160,7 +182,7 @@ function run () {
   const choices = _.flow([
     _.filter(c => c.type !== 'erc-20'),
     _.map(c => {
-      const name = c.code === 'ethereum' ? 'Ethereum and/or USDT' : c.display
+      const name = c.code === 'ethereum' ? 'Ethereum' : c.display
```
RafaelTaranto, line 185: Hmm, should USDT be disabled as well?

tools/build-dev-env.js
```diff
@@ -6,6 +6,8 @@ const setEnvVariable = require('./set-env-var')
 
 fs.copyFileSync(path.resolve(__dirname, '../.sample.env'), path.resolve(__dirname, '../.env'))
 
+setEnvVariable('NODE_ENV', 'development')
```
RafaelTaranto, line 9: This already defaults to development, no?

tools/build-prod-env.js
```diff
@@ -15,6 +15,8 @@ if (!_.isEqual(_.intersection(_.keys(argv), requiredParams), requiredParams)) {
 
 fs.copyFileSync(path.resolve(__dirname, '../.sample.env'), process.env.NODE_ENV === 'production' ? path.resolve('/etc', 'lamassu', '.env') : path.resolve(__dirname, '../.env'))
 
+setEnvVariable('NODE_ENV', 'production')
```
RafaelTaranto, line 18: This is problematic. We're not really updating this file with lamassu-update, also we can't just rewrite everything, we need a "migration" mechanism for this. This is also a risky thing to migrate since the database password gets stored here, so maybe leaving backups in case of migration is a good idea.

RafaelTaranto (changes requested):

tools/build-dev-env.js
chaotixkilla: I'm on WSL2 and it's defaulting to undefined

tools/build-prod-env.js
chaotixkilla: Added the mechanism on a new commit

PR #1280, opened 2022-06-30
Add Electrum wallet for BTC

siiky:


migrations/1655300525155-wallet-indeces.js
```diff
@@ -0,0 +1,17 @@
+const db = require('./db')
+
+exports.up = (next) => {
+  const sql = [
+    `CREATE TABLE IF NOT EXISTS wallet_indeces (
```
chaotixkilla: Shouldn't it be `wallet_indexes`?

lib/plugins/wallet/electrum/electrum.js
```diff
@@ -0,0 +1,504 @@
...
+const { PromiseObject } = require('../helpers')
+const { loadAccounts } = require('../../../new-settings-loader')
+
+
+const electrumConsts = loadAccounts().then(_.get('electrum'))
+
+const electrumPromise = electrumConsts
+  .then(({ host, port, protocol }) => {
+    const electrum = new ElectrumClient(port, host, protocol)
+    electrum.initElectrum({ client: "someclient x.y.z", version: "1.4" })
```
chaotixkilla: Feel like this is a standalone name. We could possibly call it `Lamassu Electrum Client` or something.

What about the version string? Is it the electrum version or something?

BTW, noticed that the package being used is an outdated package (2 years+ unchanged). Is there something specific about this package that newer ones (like https://www.npmjs.com/package/@aguycalled/electrum-client-js) don't do?

lib/plugins/wallet/electrum/electrum.js
```diff
@@ -0,0 +1,504 @@
...
+const wallet_hash = helpers.computeWalletHash(mnemonic)
+
+const GAP_LIMIT = 20
+
+
+const freeAddressIndeces = () =>
+  helpers.freeAddressIndeces('BTC', wallet_hash, ['internal', 'external'])
+
+/* Retrieves the next {in,ex}ternal free index */
+const _nextFreeAddressIndex = (type) => () => freeAddressIndeces().then(_.get(type))
```
chaotixkilla: Am I missing something here?

```suggestion
const _nextFreeAddressIndex = (type) => freeAddressIndeces().then(_.get(type))
```

lib/plugins/wallet/electrum/electrum.js
```diff
@@ -0,0 +1,504 @@
...
+
+  const path = "m/84'/0'/0'"
+  const child = root.derivePath(path).neutered()
+  const xpub = child.toBase58()
+
+  // bitcoinjs does not support zpub yet, so we just convert it from xpub
+  let data = b58.decode(xpub)
+  data = data.slice(4) // Drop the fingerprint?
+  data = Buffer.concat([Buffer.from('04b24746', 'hex'), data])
+  const zpub = b58.encode(data)
```
chaotixkilla: Could possibly just encapsulate this code in a `xpubToZpub` function?

chaotixkilla (changes requested):

migrations/1655300525155-wallet-indeces.js
siiky: Haha @josepfo asked the exact same thing the other day. According to [Merriam-Webster](https://www.merriam-webster.com/dictionary/index), both "indeces" and "indexes" is correct. It doesn't say there, but I believe "indeces" is the brit spelling. TBH, no special reason to have written "indeces" instead of "indexes", it's what came out. If people feel it's weird or w.e. I can change it no problem.

lib/plugins/wallet/electrum/electrum.js
siiky: Ah right, the client name should probably be changed...

The version string is the protocol version. According to [this page](https://electrumx-spesmilo.readthedocs.io/en/latest/protocol-changes.html#version-1-4) (which I think is the official ElectrumX docs site), 1.4 is the latest minor version (with a couple of revisions that don't make a diff to us).

RE the package, "outdated" is not quite the word because it works (:
At the time I tried a couple other packages (neither of which was the one you suggest) but I didn't manage to get them working... Settled for this one because it's what we use on the webstore and it works.

lib/plugins/wallet/electrum/electrum.js
siiky: Yeah, the two lines beneath:

https://github.com/lamassu/lamassu-server/blob/c3a938385d499f05346c8b8e2b92a97a6964f5a5/lib/plugins/wallet/electrum/electrum.js#L72-L73

It's just a "delayed" call. I can easily change it if you think it's not very readable.

lib/plugins/wallet/electrum/electrum.js
siiky: Yup, can do.

siiky: Actually, the zpub isn't used anywhere so...

lib/plugins/wallet/electrum/electrum.js
siiky: Changed the client name and put the "1.4" in a constant so it's more explicit.

Waiting for your (or Rafael's?) reply on the package question.

migrations/1655300525155-wallet-indeces.js
siiky: Changed to `indices`, as decided today.

chaotixkilla (approved):

PR #1283, opened 2022-07-04
Add automatic receipt printing

chaotixkilla:


new-lamassu-admin/src/pages/OperatorInfo/ReceiptPrinting.js
```diff
@@ -69,6 +69,28 @@ const ReceiptPrinting = memo(({ wizard }) => {
           <Label2>{receiptPrintingConfig.active ? 'Yes' : 'No'}</Label2>
         </div>
       </div>
+      <div className={classes.switchRow}>
+        <P>Automatic receipt printing</P>
+        <div className={classes.switch}>
+          <Switch
+            disabled={!receiptPrintingConfig.active}
```
siiky, line 89: Just to confirm: this (`active`) is whether paper printing is enabled, right? The same `active` that is saved in `user_config`.

lib/graphql/resolvers.js
```diff
@@ -33,6 +33,7 @@ const addOperatorInfo = addSmthInfo(
 const addReceiptInfo = addSmthInfo(
   'receiptInfo',
   [
+    'automaticPrint',
```
siiky, line 37: Same comment as I left in Jira: this will conflict with #1282. Depending on which one is merged first, the other has to be changed.

siiky (approved):

PR #1284, opened 2022-07-05
Add l-s pre-packager

chaotixkilla:


siiky (approved):

build/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+RUN apt-get install nodejs -y -q
+
+FROM base as l-s-build
+
+WORKDIR /lamassu
+
+RUN git clone https://github.com/lamassu/lamassu-server -b ${SERVER_VERSION}
+RUN rm -rf /lamassu/lamassu-server/public/*
+RUN cd lamassu-server && npm install --production
+RUN ls -la
```
RafaelTaranto: I don't think this is needed.

build/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+RUN cd lamassu-server/new-lamassu-admin && npm install && npm run build
```
RafaelTaranto, line 30: Doesn't this mean we're also shipping with the new-lamassu-admin node_modules? We can remove that folder after the build.

RafaelTaranto (changes requested):

PR #1287, opened 2022-07-07
Add button to check against OFAC sanction list

chaotixkilla:


lib/new-admin/middlewares/loadSanctionLists.js
```diff
@@ -0,0 +1,28 @@
+const logger = require('../../logger')
+const sanctions = require('../../ofac')
+
+const sanctionStatus = {
+  loaded: false,
+  timestamp: null
```
siiky: The timestamp doesn't seem to be used for anything?

lib/new-admin/admin-server.js
```diff
@@ -48,6 +48,7 @@ app.use(express.static(path.resolve(__dirname, '..', '..', 'public')))
 app.use(cleanUserSessions(USER_SESSIONS_CLEAR_INTERVAL))
 app.use(computeSchema)
 app.use(findOperatorId)
+app.use(loadSanctionLists)
```
siiky: Wouldn't it be easier (maybe better) to run `load()` here, instead of erroring out?

https://github.com/lamassu/lamassu-server/blob/d446b0635af87ab7570cf9c84ef574709f4c04e5/lib/ofac/index.js#L48-L51

No need to run the `loadSanctionLists()` on every request. Only the first `match()` would be slower.

siiky (approved):

siiky (approved):

RafaelTaranto (changes requested):
I don't think we should make it the front-end job to update the customer with sanctions true. Checking for sanctions is already baked into the server. 

Probably better to just send the ID of the customer we want to check against, and do everything automatically on the server, with a response for display purposes nonetheless.

PR #1295, opened 2022-07-13
Add market currency selector

chaotixkilla:


siiky (approved):

siiky (approved):

migrations/1657807941757-market-currency.js
```diff
@@ -0,0 +1,31 @@
...
+
+      const newAccounts = _.reduce(
+        (acc, value) => {
+          if (!_.isNil(accounts[value].currencyMarket)) return acc
+          if (_.includes(localeCurrency, ALL[value].FIAT)) return { ...acc, [value]: { currencyMarket: localeCurrency } }
+          return { ...acc, [value]: { currencyMarket: _.head(ALL[value].FIAT) } }
+        },
+        {},
+        configuredExchanges
+      )
```
RafaelTaranto: Isn't this changing the default behaviour? If we used to default to EUR, why are we not defaulting to EUR here? 

Relying on the FIAT order for default fiat currency might be problematic as well. Maybe create a new variable like DEFAULT_FIAT_MARKET.

new-lamassu-admin/src/utils/constants.js
```diff
@@ -9,6 +9,11 @@ const MANUAL = 'manual'
 const IP_CHECK_REGEX = /^(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$/
 
 const SWEEPABLE_CRYPTOS = ['ETH']
+const WARNING_LEVELS = {
+  CLEAN: 'clean',
+  PARTIAL: 'partial',
+  IMPORTANT: 'important'
+}
```
RafaelTaranto: Looks like this is pretty specific of the Services page. Any reason why is on the generic constant file?

lib/plugins/common/ccxt.js
```diff
@@ -28,15 +28,13 @@ const ALL = {
...
     throw new Error('Unsupported crypto: ' + cryptoCode)
   }
-  const fiatSupported = ALL[serviceName].FIAT
-  if (fiatSupported !== 'ALL_CURRENCIES' && !_.includes(fiatCode, fiatSupported)) {
-    logger.info('Building a market for an unsupported fiat. Defaulting to EUR market')
-    return cryptoCode + '/' + 'EUR'
-  }
+
+  if (_.isNil(_fiatCode)) logger.debug('Missing fiat code information, defaulting to EUR markets')
+  const fiatCode = _fiatCode ?? 'EUR'
```
RafaelTaranto: If we're running that migration, is there a reason for this to exist?

lib/plugins/exchange/ccxt.js
```diff
@@ -49,4 +54,36 @@ function calculatePrice (side, amount, orderBook) {
   throw new Error('Insufficient market depth')
 }
 
-module.exports = { trade }
+function _getMarkets (exchangeName, availableCryptos) {
+  try {
+    const exchange = new ccxt[exchangeName]()
+    const cryptosToQuoteAgainst = ['USDT']
```
RafaelTaranto, line 60: Is there a reason we're quoting against USDT instead of USD? What happens if it depegs?

lib/plugins/exchange/ccxt.js
```diff
@@ -49,4 +54,36 @@ function calculatePrice (side, amount, orderBook) {
...
 
-module.exports = { trade }
+function _getMarkets (exchangeName, availableCryptos) {
+  try {
+    const exchange = new ccxt[exchangeName]()
+    const cryptosToQuoteAgainst = ['USDT']
+    const currencyCodes = _.concat(_.map(it => it.code, currencies), cryptosToQuoteAgainst)
+
+    return exchange.fetchMarkets()
+      .then(_.filter(it => (it.type === 'spot' || it.spot)))
```
RafaelTaranto, line 64: We are using market trading for some exchanges. I think at this point we'll need to filter depending on exchange.

RafaelTaranto (changes requested):

lib/plugins/exchange/ccxt.js
chaotixkilla: We're quoting against both USDT and USD. This variable represents a list of cryptos that operators want to have available to trade with on the exchange (e.g. USDT/BTC pair). USDT comes as a specific request in the comments of the issue card

lib/plugins/exchange/ccxt.js
chaotixkilla: I investigated further and correct me if I'm wrong on this. I believe there's a misunderstanding between order types and market types. We execute trading orders with both market and limit orders, but these orders are executed on the spot market of the respective exchanges. Other options for market are `future` and `swap` (for our supported exchanges), but I don't think we do trades on those markets

PR #1298, opened 2022-07-14
Cash collection receipts

chaotixkilla:


lib/cashbox-batches.js
```diff
@@ -69,32 +70,58 @@ function getBillsByBatchId (id) {
...
+}
+
+function getBatchById (id) {
+  const sql = `
+    SELECT cb.id, cb.device_id, cb.created, cb.operation_type, cb.bill_count_override, cb.performed_by, json_agg(b.*) AS bills
+    FROM cashbox_batches AS cb
+    LEFT JOIN bills AS b ON cb.id = b.cashbox_batch_id
+    WHERE cb.id = $1
+    GROUP BY cb.id
+  `
```
josepfo, line 93: If I got this right, we use this function to extract a bit more info for the receipt, like `operation_type` and the `performed_by` (btw, the latter wasn't used on the machine PR, was that on purpose?). 
My point is: is this really worth it? It seems like we could just use `getMachineUnbatchedBills` in every situation.

josepfo (approved):

PR #1300, opened 2022-07-15
Add machine coin rates screen

chaotixkilla:


lib/plugins.js
```diff
@@ -231,6 +231,7 @@ function plugins (settings, deviceId) {
     const localeConfig = configManager.getLocale(deviceId, settings.config)
     const fiatCode = localeConfig.fiatCurrency
     const cryptoCodes = localeConfig.cryptoCurrencies
+    const machineScreenOpts = configManager.getAllMachineScreenOpts(settings.config)
```
siiky, line 234: `getMachineScreenOpts(screenName, config)` but only the config is given here (and in the wrong order).

lib/graphql/resolvers.js
```diff
@@ -99,6 +106,7 @@ const staticConfig = ({ currentConfigVersion, deviceId, deviceName, pq, settings
     configManager.getLocale(deviceId, settings.config),
     configManager.getOperatorInfo(settings.config),
     configManager.getReceipt(settings.config),
+    configManager.getAllMachineScreenOpts(settings.config),
```
siiky, line 114: `addSmthInfo()` checks if the given `smth` has an `active` value and is truthy, but looking at `getAllMachineScreenOpts()` I don't think there's any.

siiky (changes requested):

lib/graphql/resolvers.js
chaotixkilla: fixed

lib/graphql/resolvers.js
```diff
@@ -131,7 +145,8 @@ const staticConfig = ({ currentConfigVersion, deviceId, deviceName, pq, settings
         }),
         _.update('triggersAutomation', _.mapValues(_.eq('Automatic'))),
         addOperatorInfo(operatorInfo),
-        addReceiptInfo(receiptInfo)
+        addReceiptInfo(receiptInfo),
+        _.update('screenOptions', addMachineScreenOpts(machineScreenOpts))
```
siiky: Isn't this `_.update()` unnecessary here? It's already in `addMachineScreenOpts()`.

siiky (changes requested):

siiky (approved):

PR #1303, opened 2022-07-19
Fix notification center inconsistencies

chaotixkilla:


lib/notifier/index.js
```diff
@@ -136,7 +136,7 @@ function transactionNotify (tx, rec) {
 
     // for notification center
     const directionDisplay = isCashOut ? 'cash-out' : 'cash-in'
-    const readyToNotify = !isCashOut || (tx.direction === 'cashOut' && rec.isRedemption)
+    const readyToNotify = _.defaultTo(false, notifSettings.notificationCenter.transactions) && (!isCashOut || (isCashOut && rec.isRedemption))
```
siiky, line 139: Shouldn't `notifSettings.notificationCenter.active` be checked as well? Or is that done somewhere up the call tree? I just tried disabling the notification center without disabling transaction notifications, and `notifications_notificationCenter_transactions` is still in the DB.

siiky (changes requested):

PR #1320, opened 2022-07-27
Remove coin-specific cryptounits from config

chaotixkilla:


lib/new-config-manager.js
```diff
@@ -152,7 +152,8 @@ const getCryptosFromWalletNamespace = config => {
 const getCashInSettings = config => fromNamespace(namespaces.CASH_IN)(config)
 
 const getCryptoUnits = (crypto, config) => {
-  return getWalletSettings(crypto, config).cryptoUnits
+  const walletSettings = getWalletSettings(crypto, config)
+  return walletSettings.cryptoUnits
```
siiky: Is this necessary for the fix? (no problem with it, I'm just curious to know if there's some stupid JS "feature" that makes a difference here)

siiky (approved):

siiky (approved):

PR #1332, opened 2022-08-01
Add wallet transaction history to the funding page

chaotixkilla:


new-lamassu-admin/src/pages/Funding.js
```diff
@@ -1,43 +1,44 @@
+/* eslint-disable no-unused-vars */
```
siiky: Is this necessary?

siiky (approved):

PR #1338, opened 2022-08-02
[8.1] Fix EIP-1559 fees

chaotixkilla:


siiky (changes requested):
Same as https://github.com/lamassu/lamassu-server/pull/1339

siiky (approved):

PR #1339, opened 2022-08-02
[8.0] Fix EIP-1559 fees

chaotixkilla:


lib/plugins/wallet/geth/base.js
```diff
@@ -10,6 +10,7 @@ const Tx = require('ethereumjs-tx')
 const util = require('ethereumjs-util')
 const coins = require('@lamassu/coins')
 const { default: PQueue } = require('p-queue')
+const { BigNumber } = require('bignumber.js')
```
siiky: `lib/bn.js` is already imported beneath, and `BN.minimum()` seems to work.

siiky (changes requested):

siiky (approved):

PR #1351, opened 2022-08-09
fix: don't add fixed fee twice for profit calculation

siiky:


new-lamassu-admin/src/pages/Transactions/DetailsCard.js
```diff
@@ -89,23 +89,6 @@ const getCryptoAmount = tx =>
   coinUtils.toUnit(new BigNumber(tx.cryptoAtoms), tx.cryptoCode).toNumber()
 
 /* Port of getProfit() from lib/new-admin/services/transactions.js */
-const getCommission = tx => {
```
chaotixkilla: Probably can delete the `/* Port of getProfit() from lib/new-admin/services/transactions.js */` comment, right? Or is the next piece of code also duplicated from there?

chaotixkilla (approved):

new-lamassu-admin/src/pages/Transactions/DetailsCard.js
siiky: Ah, right, thanks! Don't know how I left that...

PR #1352, opened 2022-08-09
fix: update analytics in real time

ubavic:


new-lamassu-admin/src/pages/Dashboard/SystemPerformance/SystemPerformance.js
```diff
@@ -45,25 +60,47 @@ const GET_DATA = gql`
...
+    onCompleted: data => {
+      setTransactions(
+        R.unionWith(
+          R.eqBy(R.prop('id')),
+          data?.transactions || [],
+          transactions
+        )
+      )
+    }
+  })
```
chaotixkilla: Isn't it possible to combine these two queries into a single one?

```
const { data: latestTxs } = useQuery(GET_TRANSACTION, {
  variables: {
    from: R.isEmpty(transactions) ? endOfMinute(subDays(31, NOW)) : new Date(Date.now() - 10000),
    excludeTestingCustomers: true
  },
  pollInterval: 10000,
  notifyOnNetworkStatusChange: true,
  onCompleted: data => setTransactions(R.unionWith(R.eqBy(R.prop('id')), data.transactions || [], transactions))
})
```

If you want to know the initial loading flag to maintain the same behavior, probably best to use `useState`

```
const [initialTxLoading, setInitialTxLoading] = useState(false)

(...)

useQuery(GET_TRANSACTION, {
  variables: {
    from: R.isEmpty(transactions) ? endOfMinute(subDays(31, NOW)) : new Date(Date.now() - 10000),
    excludeTestingCustomers: true
  },
  pollInterval: 10000,
  notifyOnNetworkStatusChange: true,
  onCompleted: data => {
    setInitialTxLoading(true) // initialTxLoading can then be used to ensure that nothing is rendered without any transactions
    setTransactions(R.unionWith(R.eqBy(R.prop('id')), data.transactions || [], transactions))
  }
})
```

chaotixkilla (changes requested):

chaotixkilla (approved):

PR #1364, opened 2022-08-29
Add cash-in notifications settings

siiky:


lib/new-admin/graphql/resolvers/bill.resolver.js
```diff
@@ -2,7 +2,7 @@ const bills = require('../../services/bills')
 
 const resolvers = {
   Query: {
-    bills: (...[, { filters }]) => bills.getBills(filters)
+    bills: (parent, { filters }, ctx, info) => bills.getBills(filters)
```
chaotixkilla: This change isn't needed, right?

chaotixkilla (changes requested):

lib/new-admin/graphql/resolvers/bill.resolver.js
siiky: That's probably true, it just took me a long while to figure out what was happening and where the heck `filters` came from.

chaotixkilla (approved):

PR #1368, opened 2022-09-02
feat: coinbasepro as wallet provider

josepfo:


lib/new-config-manager.js
```diff
@@ -162,6 +162,11 @@ const getCryptoUnits = (crypto, config) =>
 
 const setTermsConditions = toNamespace(namespaces.TERMS_CONDITIONS)
 
+const getCashOutConstraints = (cryptoCode, settings) => {
+  const CASH_IN_ONLY_WALLETS = ['coinbasepro']
+    return _.includes(getWalletSettings(cryptoCode, settings).wallet, CASH_IN_ONLY_WALLETS)
+}
+
```
chaotixkilla: There's a different identation on the return for no specific reason

```suggestion
const getCashOutConstraints = (cryptoCode, settings) => {
  const CASH_IN_ONLY_WALLETS = ['coinbasepro']
  return _.includes(getWalletSettings(cryptoCode, settings).wallet, CASH_IN_ONLY_WALLETS)
}

```

new-lamassu-admin/src/pages/Wizard/components/Wallet/ChooseWallet.js
```diff
@@ -152,6 +152,21 @@ const ChooseWallet = ({ data: currentData, addData }) => {
...
+            save={saveWallet(selected)}
+            elements={schema.coinbasepro.elements}
+            validationSchema={schema.coinbasepro.getValidationSchema(
+              accounts.coinbasepro
+            )}
+            buttonLabel={'Continue'}
+            buttonClass={classes.formButton}
+          />
+        </>
+      )}
```
chaotixkilla, line 169: We need to start generalizing this code. Right now, `bitgo`'s schema takes a coin as parameter, but we could add the coin parameter to every wallet schema to make generalization easier. With the services revamp PR, my suggestion will vary a bit for sure, but I'd go something like this

```suggestion
      {selected && (
        <>
          <H4 noMargin>Enter wallet information</H4>
          <FormRenderer
            value={accounts[selected]}
            save={saveWallet(selected)}
            elements={schema[selected](coin).elements}
            validationSchema={schema[selected](coin).getValidationSchema(
              accounts.[selected]
            )}
            buttonLabel={'Continue'}
            buttonClass={classes.formButton}
          />
        </>
      )}
```

new-lamassu-admin/src/pages/Wallet/Wallet.js
```diff
@@ -177,6 +216,33 @@ const Wallet = ({ name: SCREEN_KEY }) => {
               />
             </Modal>
           )}
+          {cashInOnlyWallet && (
+            <Modal
+              title={'FYI'}
```
chaotixkilla, line 221: Is this the final title?

lib/plugins/wallet/coinbasepro/coinbasepro.js
```diff
@@ -0,0 +1,163 @@
...
+
+function checkBlockchainStatus (cryptoCode) {
+  return checkCryptoCode(cryptoCode)
+    .then(() => Promise.resolve('ready'))
+}
+
+function cryptoNetwork (account, cryptoCode, settings, operatorId) {
+  return checkCryptoCode(cryptoCode)
+    .then('main')
+}
```
chaotixkilla: This will not return `main` and will return whatever the result of the previous promise is (in this case, `undefined`). If the `onFulfilled` is not a function, it will just pass forward a function `(x) => x`.

Better explained in the [Mozilla JS Reference on promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then): `onFulfilled: A function asynchronously called if the Promise is fulfilled. This function has one argument, the fulfillment value. If it is not a function, it is internally replaced with an identity function ((x) => x) which simply passes the fulfillment value forward.`

You can confirm that with this snippet
```
const checkCryptoCode = (code) => Promise.resolve();

const cryptoNetwork = (account, cryptoCode, settings, operatorId) => {
  return checkCryptoCode(cryptoCode).then("main");
};

cryptoNetwork(null, "BTC", null, null).then(console.log);
```

```suggestion
function cryptoNetwork (account, cryptoCode, settings, operatorId) {
  return checkCryptoCode(cryptoCode)
    .then(() => 'main')
}
```

lib/plugins/wallet/coinbasepro/coinbasepro.js
```diff
@@ -0,0 +1,163 @@
...
+    return Promise.reject(new Error('Unsupported crypto: ' + cryptoCode))
+  }
+
+  return Promise.resolve()
+}
+
+function checkBlockchainStatus (cryptoCode) {
+  return checkCryptoCode(cryptoCode)
+    .then(() => Promise.resolve('ready'))
+}
```
chaotixkilla: The `Promise.resolve()` is already implicit

```suggestion
function checkBlockchainStatus (cryptoCode) {
  return checkCryptoCode(cryptoCode)
    .then(() => 'ready')
}
```

lib/plugins/wallet/coinbasepro/coinbasepro.js
```diff
@@ -0,0 +1,163 @@
...
+        getRates(fiatCode, cryptoCode)
+      ]
+      return Promise.all(promises)
+    })
+    .then(([wallet, fundingAddress, ratesRec]) => ({
+      fundingConfirmedBalance: new BN(wallet.balance),
+      fundingPendingBalance: convertPendingBalance(ratesRec, new BN(wallet['hold_balance']), cryptoCode),
+      fundingAddress
+    }))
+}
```
chaotixkilla: Not sure if I like having ticker related logic in a wallet file. How hard would it be to pass the rate as a parameter for `newFunding()`? If you choose this, make sure to rename the other parameters accordingly

```suggestion
function newFunding (account, cryptoCode, fiatRate, settings, operatorId) {
  return checkCryptoCode(cryptoCode)
    .then(() => {
      const promises = [
        getWallet(account, cryptoCode),
        newAddress(account, { cryptoCode }),
        fiatRate
      ]
      return Promise.all(promises)
    })
    .then(([wallet, fundingAddress, ratesRec]) => ({
      fundingConfirmedBalance: new BN(wallet.balance),
      fundingPendingBalance: convertPendingBalance(ratesRec, new BN(wallet['hold_balance']), cryptoCode),
      fundingAddress
    }))
}
```

chaotixkilla (changes requested):

new-lamassu-admin/src/pages/Wizard/components/Wallet/ChooseWallet.js
josepfo: I'm leaving this as it is and we will deal with the conflicts afterwards.

lib/plugins/wallet/coinbasepro/coinbasepro.js
josepfo: @RafaelTaranto any thoughts about this?

chaotixkilla (approved):

PR #1372, opened 2022-09-12
Allow for BTC remote node connection

chaotixkilla:


tools/build-dev-env.js
```diff
@@ -28,5 +31,12 @@ setEnvVariable('OPERATOR_DATA_DIR', `${process.env.HOME}/.lamassu/operatordata`)
 setEnvVariable('OFAC_SOURCES_NAMES', 'sdn_advanced,cons_advanced')
 setEnvVariable('OFAC_SOURCES_URLS', 'https://www.treasury.gov/ofac/downloads/sanctions/1.0/sdn_advanced.xml,https://www.treasury.gov/ofac/downloads/sanctions/1.0/cons_advanced.xml')
 
+setEnvVariable('BTC_NODE_LOCATION', 'remote')
+setEnvVariable('BTC_WALLET_LOCATION', 'local')
+setEnvVariable('BTC_NODE_HOST', '146.190.228.59')
```
siiky: Is this a trusted node? Can we really use it for development? And if it's operated by us, should it be shared here publicly?

lib/blockchain/bitcoin.js
```diff
@@ -1,24 +1,31 @@
...
-  const cmd = `/usr/local/bin/${coinRec.daemon} -datadir=${dataDir}`
-  common.writeSupervisorConfig(coinRec, cmd)
+  const cmd = `${usrBinDir}/${coinRec.daemon} -datadir=${dataDir}`
+  !isDevMode() && common.writeSupervisorConfig(coinRec, cmd)
 }
 
 function updateCore (coinRec, isCurrentlyRunning) {
   common.logger.info('Updating Bitcoin Core. This may take a minute...')
-  common.es(`sudo supervisorctl stop bitcoin`)
+  !isDevMode() && process.env.NODE_ENV !== 'development' && common.es(`sudo supervisorctl stop bitcoin`)
```
siiky: The `process.env.NODE_ENV !== 'development'` is redundant given `!isDevMode()`.

lib/plugins/common/json-rpc.js
```diff
@@ -111,10 +112,12 @@ function rpcConfig (cryptoRec) {
...
     const config = parseConf(configPath)
+    
     return {
-      username: config.rpcuser,
-      password: config.rpcpassword,
-      port: config.rpcport || cryptoRec.defaultPort
+      username: isRemoteWallet(cryptoRec) ? process.env[`${cryptoRec.cryptoCode}_NODE_USER`] : config.rpcuser,
+      password: isRemoteWallet(cryptoRec) ? process.env[`${cryptoRec.cryptoCode}_NODE_PASSWORD`] : config.rpcpassword,
+      host: isRemoteNode(cryptoRec) ? process.env[`${cryptoRec.cryptoCode}_NODE_HOST`] : 'localhost',
+      port: isRemoteNode(cryptoRec) ? process.env[`${cryptoRec.cryptoCode}_NODE_PORT`] : config.rpcport || cryptoRec.defaultPort
```
siiky, line 120: Aren't all these envvars set in the `.env` file, whether it's a dev or prod environment? If so, shouldn't the envvars be respected, instead of silently ignoring them in a dev environment? Having a fallack default is probably still worthwhile.

siiky (changes requested):

tools/build-dev-env.js
chaotixkilla: It is a trusted node, operated by us. Your question is indeed correct, so I rebased that line to redact it

lib/plugins/common/json-rpc.js
chaotixkilla: Not always. as for local node + wallet (the only setup we currently have), those environment variables are not required, hence the check and the fallback values

siiky: Got it, I must have misread the condition because I was thinking `isDev` the whole time.

siiky (approved):

RafaelTaranto (approved):

PR #1374, opened 2022-09-12
Add firehose for webhook notifications

chaotixkilla:


lib/notifier/webhook.js
```diff
@@ -0,0 +1,21 @@
+const axios = require('axios')
+const _ = require('lodash/fp')
+const uuid = require('uuid')
+
+const sendMessage = (settings, rec) => {
+  const webhookUrl = process.env.WEBHOOK_URL
```
siiky: Is the URL expected to change ever after program start? Reading envvars can be expensive, so if possible it should be done once only, not on every call.

lib/notifier/webhook.js
```diff
@@ -0,0 +1,21 @@
...
+
+  if (_.isEmpty(webhookUrl)) return Promise.resolve()
```
siiky: If the above is possible, then there's no need to penalize those who don't use webhooks:

```js
const sendMessage = _.isEmpty(webhookUrl) ?
  () => Promise.resolve() :
  (settings, rec) => /* ... */
```

siiky (changes requested):

siiky (approved):

PR #1379, opened 2022-09-16
Fail fast on lamassu-server default route access and SEO robots.txt

chaotixkilla:


new-lamassu-admin/public/robots.txt
```diff
@@ -0,0 +1,2 @@
+User-agent: *
+Disallow: /
```
siiky, line 2: Is this file used?

siiky (approved):

lib/routes.js
```diff
@@ -53,6 +53,15 @@ app.use(nocache())
 app.use(express.json({ limit: '2mb' }))
 app.use(morgan(':method :url :status :response-time ms - :res[content-length]', { stream: logger.stream }))
 
+app.use('/robots.txt', (req, res) => {
+  res.type('text/plain')
+  res.send("User-agent: *\nDisallow: /")
+})
+
+app.get('/', (req, res) => {
+  res.status(404).json({ error: 'No such route' })
```
RafaelTaranto: No need to return the json here. The more generic the better, so just 404 should be enough.

RafaelTaranto (changes requested):

PR #1381, opened 2022-09-20
Reduce occurrence of concurrent updates on 'devices' table during cash-in transactions

chaotixkilla:


lib/middlewares/populateSettings.js
```diff
@@ -6,6 +6,8 @@ const newSettingsLoader = require('../new-settings-loader')
 const helpers = require('../route-helpers')
 const logger = require('../logger')
 
+let LATEST_MACHINE_CONFIG_VERSION
```
siiky: This is unused now. Intentional?

siiky (changes requested):

PR #1394, opened 2022-09-30
Add machine physical location

chaotixkilla:


lib/new-admin/graphql/resolvers/pairing.resolver.js
```diff
@@ -1,8 +1,10 @@
 const pairing = require('../../services/pairing')
+const machine = require('../../../machine-loader')
 
 const resolvers = {
   Mutation: {
-    createPairingTotem: (...[, { name }]) => pairing.totem(name)
+    createPairingTotem: (...[, { name, location }]) => machine.createMachineLocation(location)
+      .then(() => pairing.totem(name, location))
```
siiky: The type of the `location` parameter in `pairing.totem()` is different from that of `machine.createMachineLocation()`.

While `createMachineLocation()` expects an object with `id`, `label`, etc (which matches the GQL query's type), `totem()` expects the location's ID.

Also note that the `id` is optional for `createMachineLocation()` (a new one is generated if not given), in which case you don't know at this point what the ID of the location is/will be. Changing `createMachineLocation()` to always return the ID of the created location probably is a good enough solution (currently returns nothing as the promise result).

lib/machine-loader.js
```diff
@@ -204,10 +218,17 @@ function setMachine (rec, operatorId) {
...
+    case 'deleteLocation': return deleteMachineLocation(rec.location.id)
+    case 'createLocation': return createMachineLocation(rec.location, rec.deviceId)
     default: throw new Error('No such action: ' + rec.action)
   }
 }
 
+function testLocation (rec) {
+  console.log(rec)
+}
+
```
siiky: Development artifact?

new-lamassu-admin/src/pages/Machines/MachineComponents/Overview.js
```diff
@@ -16,14 +17,31 @@ const Overview = ({ data, onActionSuccess }) => {
...
+          </div>
+          {!R.isNil(data.location) && (
+            <div>
+              <Label1 noMargin>Address</Label1>
+              <P noMargin>
+                {`${data.location.addressLine1}${
+                  data.location.addressLine2
+                    ? `, ${data.location.addressLine2}`
+                    : ``
+                }, ${data.location.zipCode}, ${data.location.country}`}
```
siiky: Hmmm this is a very opinionated address line... Not criticism, just raising this up in case we haven't decided on any one specific addressing format.

It's OK to choose and follow one "standard", but not all countries write down their addresses the same. Some write bottom-up (less general to more general, e.g. PT), others top-down (more general to less general, e.g. JP). Most western countries use bottom-up, but even between those there are slight differences.

https://en.wikipedia.org/wiki/Address#Address_format

siiky (changes requested):

siiky (approved):

PR #1397, opened 2022-10-04
Coin-agnostic address blacklisting

chaotixkilla:


siiky (changes requested):
Has conflicts, otherwise looks good!

new-lamassu-admin/src/pages/Blacklist/Blacklist.js
```diff
@@ -179,21 +161,21 @@ const Blacklist = () => {
     setConfirmDialog(false)
   }
 
-  const validateAddress = (cryptoCode, address) => {
+  const validateAddress = address => {
     try {
-      return !R.isNil(coinUtils.parseUrl(cryptoCode, 'main', address))
+      return !R.isEmpty(addressDetector.detectAddress(address).matches)
```
RafaelTaranto, line 166: Can we get a cleaner API here? I don't really understand what this code does based on the API nor how to use it.

RafaelTaranto (changes requested):

PR #1416, opened 2022-10-18
Fix settings caching

chaotixkilla:


lib/middlewares/populateSettings.js
```diff
@@ -55,30 +55,64 @@ const populateSettings = function (req, res, next) {
...
+
+    if (!versionId && !operatorSettings) {
+      logger.debug('Fetching the latest config version because there\'s no cached value')
+      return newSettingsLoader.loadLatest()
+        .then(settings => {
+          settingsCache.set(`${operatorId}-latest`, settings)
+          req.settings = settings
+        })
+        .then(() => next())
+        .catch(next)
```
siiky: This block looks like the `if (!versionId && !!needsSettingsReload[operatorId)` above (minus the `delete`). Maybe they could be merged?

```js
    if (!versionId && (!!needsSettingsReload[operatorId]) || !operatorSettings) {
    /* ... */
              if (!!needsSettingsReload[operatorId]) delete /* ... */

    /* ... */
}
```

siiky (changes requested):

lib/middlewares/populateSettings.js
```diff
@@ -55,30 +55,56 @@ const populateSettings = function (req, res, next) {
...
+      }
+
+      logger.debug('Fetching and caching a specific config version')
+      req.settings = cachedVersionedSettings
+      return next()
+    }
+
+    const operatorSettings = settingsCache.get(`${operatorId}-latest`)
+
+    if (!versionId && (!!needsSettingsReload[operatorId] || !operatorSettings)) {
```
RafaelTaranto: after exiting safely on top we don't need to check for !versionId anymore.

lib/middlewares/populateSettings.js
```diff
@@ -55,30 +55,56 @@ const populateSettings = function (req, res, next) {
...
+          settingsCache.set(`${operatorId}-latest`, settings)
+          if (!!needsSettingsReload[operatorId]) delete needsSettingsReload[operatorId]
           req.settings = settings
         })
         .then(() => next())
         .catch(next)
     }
-  
+
     if (!versionId && operatorSettings) {
```
RafaelTaranto: I don't think we need this if anymore.

RafaelTaranto (changes requested):

PR #1431, opened 2022-11-01
Fix: mock scoring flag

josepfo:


siiky (approved):
That error message is a bit too generic (we already have many like it) but LGTM!

PR #1439, opened 2022-11-07
fix: table scrollbar offsetting columns

ubavic:


chaotixkilla (approved):

new-lamassu-admin/src/components/tables/DataTable.styles.js
```diff
@@ -30,7 +30,10 @@ export default {
     cursor: 'pointer'
   },
   body: {
-    flex: [[1, 1, 'auto']]
+    flex: [[1, 1, 'auto']],
+    '& > div > div': {
+      overflowY: 'overlay !important'
+    }
```
RafaelTaranto: What is this div, exactly? Any reason why we're using such a specific query and a !important?

RafaelTaranto (changes requested):

new-lamassu-admin/src/components/tables/DataTable.styles.js
ubavic: I managed to do it without `>` selectors (solution was already in our code), but `!important` has to stay because the react-virtualized creates div with inlined CSS.

PR #1443, opened 2022-11-10
fix: standardize caps

ubavic:


new-lamassu-admin/src/pages/Machines/MachineComponents/Commissions/helper.js
```diff
@@ -30,15 +30,13 @@ const getOverridesFields = currency => {
     {
       header: cashInHeader,
       name: 'cashIn',
-      display: 'Cash-in',
```
siiky, line 33: Isn't the hyphen expected in "cash-in" & "cash-out"?

new-lamassu-admin/src/pages/Services/Services.js
```diff
@@ -117,7 +117,7 @@ const Services = () => {
   return (
     !loading && (
       <div className={classes.wrapper}>
-        <TitleSection title="3rd Party Services" />
+        <TitleSection title="3rd Party services" />
```
siiky: I'd say "Party" is the second word here, but this case is dubious...

new-lamassu-admin/src/pages/Wallet/helper.js
```diff
@@ -113,7 +113,7 @@ const getAdvancedWalletElements = () => {
     },
     {
       name: 'feeMultiplier',
-      header: `BTC Miner's Fee`,
+      header: `BTC Miner's fee`,
```
siiky: Here too, I'd say "Miner's" is the second word.

siiky (approved):

new-lamassu-admin/src/pages/Machines/MachineComponents/Commissions/helper.js
ubavic: You are right!

new-lamassu-admin/src/pages/Services/Services.js
ubavic: I agree with you here, but spec says otherwise. Let me check with Carol

new-lamassu-admin/src/pages/Machines/MachineComponents/Commissions/helper.js
ubavic: Actually here `display`doesn't make difference. Header is set by `header` field, and here we use `cashInHeader` component.

siiky (approved):

PR #1466, opened 2022-12-07
LAM-644 Fix: deprecate cash in fee crypto field

josepfo:


lib/tx-batching.js
```diff
@@ -65,7 +65,7 @@ function getOpenBatchCryptoValue (cryptoCode) {
       if (_.isNil(batch)) return Promise.resolve([])
       return db.any(`SELECT * FROM cash_in_txs WHERE batch_id=$1`, [batch.id])
     })
-    .then(txs => _.reduce((acc, tx) => acc.plus(tx.cash_in_fee_crypto).plus(tx.crypto_atoms), BN(0), txs))
+    .then(txs => _.reduce((acc, tx) => acc.plus(tx.crypto_atoms), BN(0), txs))
```
RafaelTaranto, line 68: This is going to break on the new machine + old server combo

Old servers need the cash_in_fee_crypto to perform this (broken) math and bignumber.js does number.plus(null) === NaN. Can you create a backwards compatibility annotated change forcing cashInFreeCrypto to zero on the machine?

RafaelTaranto (changes requested):

PR #1472, opened 2022-12-14
LAM-644 fix: remove `cashInFeeCrypto` from machine-cached transactions

siiky:


lib/cash-in/cash-in-tx.js
```diff
@@ -28,6 +28,7 @@ end`
 module.exports = {post, monitorPending, cancel, PENDING_INTERVAL, TRANSACTION_STATES}
 
 function post (machineTx, pi) {
+  machineTx = _.omit(['cashInFeeCrypto'], machineTx)
```
RafaelTaranto: I think there's a "massage" somewhere that omits fields before saving to the database, you can add it there. Also open this one for the release.

RafaelTaranto (changes requested):

PR #1473, opened 2022-12-14
LAM-730 Add new virtual denomination

siiky:


lib/plugins.js
```diff
@@ -156,28 +156,17 @@ function plugins (settings, deviceId, schema) {
...
         const redeemableTxs = _.reject(_.matchesProperty('id', excludeTxId), _redeemableTxs)
 
-        const denominations = []
-        _.forEach(it => {
-          denominations.push(cashOutConfig[`cassette${it + 1}`])
-        }, _.times(_.identity(), rec.numberOfCassettes))
-    
-        const virtualCassettes = [Math.max(...denominations) * 2]
-
-        const counts = argv.cassettes
```
josepfo, line 166: Why is this variable no longer useful?

siiky: It's a command line argument we don't currently use, that was added ~5 years ago, probably for debugging purposes.

josepfo (approved):

PR #1481, opened 2023-01-05
LAM-784 Encapsulate balance amounts from Web3 into BigNumber constructors

chaotixkilla:


ubavic (approved):

lib/plugins/wallet/geth/base.js
```diff
@@ -276,12 +276,12 @@ function getStatus (account, tx, requested, settings, operatorId) {
   return checkCryptoCode(cryptoCode)
     .then(code => Promise.all([confirmedBalance(toAddress, code), code]))
     .then(([confirmed, code]) => {
-      if (confirmed.gte(requested)) return { receivedCryptoAtoms: confirmed, status: 'confirmed' }
+      if (BN(confirmed).gte(requested)) return { receivedCryptoAtoms: confirmed, status: 'confirmed' }
```
RafaelTaranto: Why do we need to wrap it? Shouldn't the function return BN already?

RafaelTaranto (changes requested):

PR #1485, opened 2023-01-16
LAM-809 Add pazuz sms and email services

chaotixkilla:


lib/plugins.js
```diff
@@ -519,7 +519,7 @@ function plugins (settings, deviceId, schema) {
           return Promise.all(cryptoCodes.map(cryptoCode => {
             return exchange.fetchExchange(settings, cryptoCode)
               .then(exchange => ({
-                fiatCode: exchange.account.currencyMarket,
+                fiatCode: exchange.account?.currencyMarket,
```
RafaelTaranto: This produces a harmless error, right? Allowing for null here is potentially harmful. It's also not related to the PR so you can just undo the change here.

RafaelTaranto (changes requested):

PR #1504, opened 2023-02-01
Fix: remove unsupported languages

ubavic:


migrations/1675278103000-remove-languages.js
```diff
@@ -0,0 +1,59 @@
...
+    "vi-VN"
+]
+
+const changeLanguage = lang =>
+    removedLanguages.includes(lang) ? "en-US" : lang
+
+exports.up = next => {
+    loadLatest()
+    .then(({ config }) => {
+        locale_languages = _.uniq(config.locale_languages.map(changeLanguage))
```
siiky: Why not `_.difference(config.locale_languages, removedLanguages)`?

migrations/1675278103000-remove-languages.js
```diff
@@ -0,0 +1,59 @@
...
+        locale_overrides = config.locale_overrides.map(o => {
```
siiky: Missing `const`.

I guess something like this would be simpler:

```js
migrationSaveConfig({
  locale_languages: ...,
  locale_overrides: ...,
})
```

siiky (changes requested):

migrations/1675278103000-remove-languages.js
ubavic: Ah, I always forgot about these useful methods

siiky (approved):

PR #1525, opened 2023-02-27
feat: add identifiers for Cypress

ubavic:


new-lamassu-admin/src/components/layout/TitleSection.js
```diff
@@ -43,7 +43,8 @@ const TitleSection = ({
                   Icon={button.icon}
                   InverseIcon={button.inverseIcon}
                   toggle={button.toggle}
-                  forceDisable={button.forceDisable}>
+                  forceDisable={button.forceDisable}
+                  dataCy={`subpageButton-${idx}`}>
```
josepfo, line 47: Is there any reason for this to be in the camel case?

josepfo (commented):
LGTM!

new-lamassu-admin/src/components/layout/TitleSection.js
ubavic: I was using same convention everywhere. It is like the name of the component

PR #1552, opened 2023-06-07
[LAM-849] [Release 8.5] Add Aveiro admin UI

chaotixkilla:


lib/plugins.js
```diff
@@ -147,37 +148,75 @@ function plugins (settings, deviceId) {
...
+    _.forEach(it => {
+      computedStackers.push({
+        name: stackers[it].name,
+        denomination: stackers[it].denomination,
+        count: counts[it]
+      })
+    }, _.times(_.identity(), _.size(stackers)))
+
+    return computedStackers
+  }
```
siiky, line 186: This looks like a copy&paste of the above. Can't they be the same function, even if they're exported as 2 different names?

lib/plugins.js
```diff
@@ -194,6 +233,62 @@ function plugins (settings, deviceId) {
       })
   }
 
+  function buildAvailableStackers (excludeTxId) {
```
siiky, line 236: Same as above.

lib/plugins.js
```diff
@@ -645,71 +743,164 @@ function plugins (settings, deviceId) {
...
+        notes: device.cashUnits.cassette4,
         denomination: denomination4,
         fiatCode
       }
       : null
+    
+    const stacker1fAlert = device.numberOfStackers >= 1 && isUnitLow(device.cashUnits.stacker1f, CASH_UNIT_CAPACITY[device.model]['stacker'], notifications.fillingPercentageStacker1f)
+      ? {
+        code: 'LOW_RECYCLER_STACKER',
+        cassette: 4,
```
siiky, line 822: Is this supposed to be `cassette: 4` on all of these?

bin/lamassu-update-stackers
```diff
@@ -0,0 +1,36 @@
...
+  process.exit(3)
+}
+
+if (parseInt(process.argv[3]) > 3 || parseInt(process.argv[3]) < 1) {
+  console.log('Error: <number_of_stackers> is out of range. Should be a number between 1 and 3')
+  process.exit(3)
+}
+
+const deviceId = process.argv[2]
+const numberOfStackers = parseInt(process.argv[3])
```
siiky, line 24: This could be moved before the two `ifs` above.

bin/lamassu-update-stackers
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env node
+
+require('../lib/environment-helper')
+
+const _ = require('lodash')
+const db = require('../lib/db')
+
+if (process.argv.length !== 4) {
+  console.log('Usage: lamassu-update-stackers <device_id> <number_of_stackers>')
```
siiky, line 9: In the usage or one of the other error messages it could be specified what a "stacker" is, lest the operators think they should count F/R. WDYT? Or is this taken care of somewhere else?

lib/cash-out/cash-out-actions.js
```diff
@@ -40,10 +40,11 @@ function mapDispense (tx) {
   const res = {}
 
   _.forEach(it => {
-    res[`provisioned_${it + 1}`] = bills[it].provisioned
-    res[`denomination_${it + 1}`] = bills[it].denomination
-    res[`dispensed_${it + 1}`] = bills[it].dispensed
-    res[`rejected_${it + 1}`] = bills[it].rejected
+    const suffix = bills[it].name.replace(/cassette|stacker/gi, '')
```
siiky, line 43: I think this should be as strict as possible to detect possible stupid errors: `replace(/^(cassette|stacker)/, '')`

lib/cash-out/cash-out-atomic.js
```diff
@@ -107,17 +107,10 @@ function nextHd (t, isHd, tx) {
...
-  where device_id = $${_.size(tx.bills) + 1}
-  returning
-  ${_.size(tx.bills) > 0 ? `cassette1` : ``}
-  ${_.size(tx.bills) > 1 ? `, cassette2`: ``}
-  ${_.size(tx.bills) > 2 ? `, cassette3` : ``}
-  ${_.size(tx.bills) > 3 ? `, cassette4` : ``}`
+  const billsStmt = _.join(', ')(_.map(it => `${tx.bills[it].name} = ${tx.bills[it].name} - $${it + 1}`)(_.range(0, _.size(tx.bills))))
+  const returnStmt = _.join(', ')(_.map(bill => `${bill.name}`)(tx.bills))
+
+  const sql = `UPDATE devices SET ${billsStmt} WHERE device_id = $${_.size(tx.bills) + 1} RETURNING ${returnStmt}`
```
siiky, line 113: I think `pgsql` supports named parameters, which would be better than this `_.size(tx.bills)+1` hack. It looks like it's correct, though.

lib/cash-out/cash-out-helper.js
```diff
@@ -43,23 +69,37 @@ function convertBigNumFields (obj) {
 }
 
 function convertField (key) {
-  return _.snakeCase(key)
+  return _.includes('denomination', key) || _.includes('provisioned', key) ? key : _.snakeCase(key)
```
siiky, line 72: Like the regex comment, I think it would be best to be as strict as possible here: `_.startsWith(key, 'denomination') || _.startsWith(key, 'provisinioned')`

lib/cash-out/cash-out-helper.js
```diff
@@ -43,23 +69,37 @@ function convertBigNumFields (obj) {
...
-    provisioned1: bills[0]?.provisioned ?? 0,
-    provisioned2: bills[1]?.provisioned ?? 0,
-    provisioned3: bills[2]?.provisioned ?? 0,
-    provisioned4: bills[3]?.provisioned ?? 0,
-    denomination1: bills[0]?.denomination ?? 0,
-    denomination2: bills[1]?.denomination ?? 0,
-    denomination3: bills[2]?.denomination ?? 0,
-    denomination4: bills[3]?.denomination ?? 0
-  }
+  const billFields = _.map(it => _.replace(/(denomination|provisioned)/g, '$1_')(it), BILL_FIELDS)
```
siiky, line 79: Strict again:

```js
_.map(_.replace(/^(denomination|provisioned)/, '$1_'), BILL_FIELDS)
```

lib/cash-out/cash-out-helper.js
```diff
@@ -43,23 +69,37 @@ function convertBigNumFields (obj) {
...
+
+  const billsObj = _.flow(
+    _.reduce(
+      (acc, value) => {
+        const suffix = value.name.replace(/cassette|stacker/gi, '')
```
siiky, line 84: Here too:

```js
value.name.replace(/^(cassette|stacker)/, '')
```

lib/cash-out/cash-out-helper.js
```diff
@@ -43,23 +69,37 @@ function convertBigNumFields (obj) {
...
+    ),
+    it => {
+      const missingKeys = _.reduce(
+        (acc, value) => {
+          return _.assign({ [value]: 0 })(acc)
+        },
+        {}
+      )(_.difference(billFields, _.keys(it)))
+      return _.assign(missingKeys, it)
+    }
```
siiky, line 101: Think it would be easier to start with everything set to 0 in the reduce's accumulator object and overwrite with the values of the bills array.

lib/cash-out/cash-out-helper.js
```diff
@@ -78,7 +118,7 @@ function toObj (row) {
   let newObj = {}
 
   keys.forEach(key => {
-    const objKey = _.camelCase(key)
+    const objKey = key.match(/denomination|provisioned/g) ? key.replace(/_/g, '') : _.camelCase(key)
```
siiky, line 121: Strict again

lib/cash-out/cash-out-atomic.js
```diff
@@ -107,17 +107,10 @@ function nextHd (t, isHd, tx) {
...
-  ${_.size(tx.bills) > 2 ? `, cassette3 = cassette3 - $3` : ``}
-  ${_.size(tx.bills) > 3 ? `, cassette4 = cassette4 - $4` : ``}
-  where device_id = $${_.size(tx.bills) + 1}
-  returning
-  ${_.size(tx.bills) > 0 ? `cassette1` : ``}
-  ${_.size(tx.bills) > 1 ? `, cassette2`: ``}
-  ${_.size(tx.bills) > 2 ? `, cassette3` : ``}
-  ${_.size(tx.bills) > 3 ? `, cassette4` : ``}`
+  const billsStmt = _.join(', ')(_.map(it => `${tx.bills[it].name} = ${tx.bills[it].name} - $${it + 1}`)(_.range(0, _.size(tx.bills))))
+  const returnStmt = _.join(', ')(_.map(bill => `${bill.name}`)(tx.bills))
```
RafaelTaranto, line 111: This is not all that legible, maybe naming the map helps, also composing reduces a bit of the clutter. Do compose the return as well.
```js
const mapIncrement = _.map(it => `${tx.bills[it].name} = ${tx.bills[it].name} - $${it + 1}`)
const billsStmt = _.compose(_.join(', '), mapIncrement, _.range(0, _.size(tx.bills)))
```

lib/cash-out/cash-out-tx.js
```diff
@@ -73,8 +74,9 @@ function postProcess (txVector, justAuthorized, pi) {
         const rec = {}
 
         _.forEach(it => {
-          rec[`provisioned_${it + 1}`] = bills[it].provisioned
-          rec[`denomination_${it + 1}`] = bills[it].denomination
+          const suffix = bills[it].name.replace(/cassette|stacker/gi, '')
```
siiky, line 77: strict regex

new-lamassu-admin/src/pages/Cashout/Wizard.js
```diff
@@ -46,37 +47,105 @@ const Wizard = ({ machine, locale, onClose, save, error }) => {
...
+        component: Autocomplete,
+        inputProps: {
+          options: options,
+          labelProp: 'display',
+          valueProp: 'code'
+        }
+      }),
+      R.range(1, machine.numberOfCassettes + 1)
+    ),
+    R.chain(
```
siiky, line 64: Not sure if maybe this function is different in the `fp` and non-`fp` lodash: but shouldn't this be a `flatMap`?

new-lamassu-admin/src/pages/Maintenance/helper.js
```diff
@@ -0,0 +1,141 @@
...
+
+const getElements = (
+  classes,
+  config,
+  bills,
+  setWizard,
+  widths,
+  setMachineId
+) => {
+  console.log(config)
```
siiky, line 17: this should probably be left out

siiky (changes requested):

PR #1560, opened 2023-07-06
fix: concurrency issues caused by poller

RafaelTaranto:


lib/cash-out/cash-out-atomic.js
```diff
@@ -27,6 +27,12 @@ function atomic (tx, pi, fromClient) {
         const isStale = fromClient && oldTx && (oldTx.txVersion >= tx.txVersion)
         if (isStale) throw new E.StaleTxError({ txId: tx.id })
 
+        const isStaleFromServer = !fromClient && oldTx && (oldTx.txVersion > tx.txVersion)
```
joshmh, line 30: Why not ``>=``? Also why not throw an error like isStale? Is the default situation that we have an existing oldTx with the identical version?

lib/cash-out/cash-out-low.js
```diff
@@ -59,6 +59,11 @@ function diff (oldTx, newTx) {
           return updatedTx['customerId'] = newTx['customerId']
         }
         return
+      case 'dispense':
```
joshmh, line 62: Could use a comment here explaining this. I guess we don't want to return a dispense diff which will trigger a dispense if the old record already indicated "dispense". But isn't this already caught on line 51?

lib/poller.js
```diff
@@ -146,19 +145,39 @@ function initializeEachSchema (schemas = ['public']) {
   }, schemas)
 }
 
+function recursiveTimeout (func, timeout, ...vars) {
```
joshmh, line 148: Honestly, I'm lost with both of these versions.

lib/cash-out/cash-out-atomic.js
RafaelTaranto: Yeah, I'll add a comment here to explain this. Only the machine bumps tx version, so any server changes will be at best on the same version as the oldTx. 

We don't throw because the job works a bit differently than the machine. The machine needs to know that it's stale, the job should just keep working instead of stopping because of a single bad record.

lib/cash-out/cash-out-low.js
RafaelTaranto: We don't want dispense to change from `true` to `false`, line 51 only checks for null, so false would still be legal. Will add a comment here as well.

lib/poller.js
RafaelTaranto: Yeah, it used to be a simple setInterval, but that has the same pitfalls as the queue since interval doesn't wait for task completion. Ideally we want a real scheduler here, but for now we're just using recursive setTimeout. This changes the pattern a little bit but it prevents multiple instances of the same job, and we don't really care about real time here.

lib/cash-out/cash-out-low.js
joshmh: line 51 is this:

```
if (oldTx && _.isEqualWith(nilEqual, oldTx[fieldKey], newTx[fieldKey])) return
```

I think it's testing for equality, so if dispense is set in the new record but also in the old record, it should return a diff. Admittedly this code is too complex to be sure.

RafaelTaranto: Right, but it doesn't prevent swapping the flag.

PR #1565, opened 2023-07-19
LAM-921 feat: add USDC (ERC-20) support

CrypticaScriptura:


lib/plugins/common/ccxt.js
```diff
@@ -10,7 +10,7 @@ const bitpay = require('../ticker/bitpay')
 const binance = require('../exchange/binance')
 const logger = require('../../logger')
 
-const { BTC, BCH, DASH, ETH, LTC, ZEC, USDT } = COINS
+const { BTC, BCH, DASH, ETH, LTC, ZEC, XMR, USDT, USDC } = COINS
```
siiky, line 13: `XMR` doesn't seem to be used in the file

siiky (changes requested):

PR #1587, opened 2023-09-04
LAM-938: Add tron and trc-20 usdt 

RafaelTaranto:


lib/plugins/wallet/trongrid/trongrid.js
```diff
@@ -0,0 +1,14 @@
...
+const base = require('../tron/base')
+
+const NAME = 'trongrid'
+
+function run (account) {
+  if (!account.endpoint) throw new Error('Need to configure API endpoint for trongrid')
+
+  const endpoint = 'https://api.trongrid.io'
+
+  base.connect({ ...account, endpoint })
```
siiky: Shouldn't `account.endpoint` be used as the endpoint?

lib/plugins/exchange/binance.js
```diff
@@ -5,7 +5,7 @@ const { ORDER_TYPES } = require('./consts')
 
 const ORDER_TYPE = ORDER_TYPES.MARKET
 const { BTC, BCH, XMR, ETH, LTC, ZEC } = COINS
-const CRYPTO = [BTC, ETH, LTC, ZEC, BCH, XMR]
+const CRYPTO = [BTC, ETH, LTC, ZEC, BCH, XMR ]
```
siiky: `USDT` missing?

lib/plugins/wallet/tron/base.js
```diff
@@ -0,0 +1,190 @@
...
+  if (!mnemonic) throw new Error('No mnemonic seed!')
+
+  return TronWeb.fromMnemonic(mnemonic.replace(/[\r\n]/gm, ' ').trim(), `${DEFAULT_PREFIX_PATH}\/0`)
+}
+
+function paymentWallet (account, index) {
+  const mnemonic = account.mnemonic
+  if (!mnemonic) throw new Error('No mnemonic seed!')
+
+  return TronWeb.fromMnemonic(mnemonic.replace(/[\r\n]/gm, ' ').trim(), `${PAYMENT_PREFIX_PATH}\/${index}`)
```
siiky: Should `\/` after `PAYMENT_PREFIX_PATH` be escaped?

lib/plugins/wallet/tron/base.js
```diff
@@ -0,0 +1,190 @@
...
+}
+
+function getStatus (account, tx, requested, settings, operatorId) {
+  const { toAddress, cryptoCode } = tx
+  return checkCryptoCode(cryptoCode)
+    .then(code => confirmedBalance(toAddress, code))
+    .then((confirmed) => {
+      if (confirmed.gte(requested)) return { receivedCryptoAtoms: confirmed, status: 'confirmed' }
+      return { receivedCryptoAtoms: 0, status: 'notSeen' }
+    })
```
siiky: Isn't it possible to get a transaction of part of the requested value? Like with the other wallets:

https://github.com/lamassu/lamassu-server/blob/e3892e9ac2f111fedf26545441c18419250cb773/lib/plugins/wallet/bitcoind/bitcoind.js#L148

siiky (changes requested):

lib/plugins/wallet/trongrid/trongrid.js
RafaelTaranto: It was decided this is hardcoded, new version doesn't have the check.

lib/plugins/exchange/binance.js
RafaelTaranto: USDT doesn't work for this one, I'll remove the empty space

lib/plugins/wallet/tron/base.js
RafaelTaranto: it's already an escaped forward slash

lib/plugins/wallet/tron/base.js
RafaelTaranto: Good catch, updated.

RafaelTaranto: Actually the problem here is a different one, we don't really have pending amount. Same thing happens with ETH so I'll get confirmation from tech first to see what to do here.

PR #1620, opened 2023-11-28
LAM-996 feat: customer auth via email

RafaelTaranto:


lib/new-config-manager.js
```diff
@@ -1,3 +1,5 @@
+const {AUTH_METHODS} = require('./compliance-triggers')
```
siiky, line 1: `AUTH_METHODS` isn't used?

PR #1667, opened 2024-03-27
LAM-969 feat: pending compliance notifications

siiky:


lib/routes/customerRoutes.js
```diff
@@ -24,15 +25,48 @@ const plugins = require('../plugins')
...
 
-function updateCustomerCustomInfoRequest (customerId, patch, req, res) {
-  if (_.isNil(patch.data)) {
-    return customers.getById(customerId)
-      .then(customer => respond(req, res, { customer }))
+function updateCustomerCustomInfoRequest (customerId, patch) {
+  const promise = _.isNil(patch.data) ?
+    Promise.resolve(null) :
+    customInfoRequestQueries.setCustomerDataViaMachine(customerId, patch.infoRequestId, patch)
+  return promise.then(() => customers.getById(customerId))
```
RafaelTaranto, line 33: Is this missing a `respond` call?

RafaelTaranto (changes requested):

lib/routes/customerRoutes.js
siiky: No, I moved the `respond` outside, to `updateCustomer` after the call to this function. I can move it back inside, but `updateCustomerCustomInfoRequest` is misleading...

PR #1680, opened 2024-05-01
[LAM-1093] feat: scorechain address analysis

RafaelTaranto:


lib/tx.js
```diff
@@ -82,7 +82,7 @@ function customerHistory (customerId, thresholdDays) {
       FROM cash_out_txs txOut
       WHERE txOut.customer_id = $1
       AND txOut.created > now() - interval $2
-      AND (error_code IS NULL OR error_code NOT IN ('operatorCancel', 'scoreThresholdReached', 'ciphertraceError'))
+      AND (error_code IS NULL OR error_code NOT IN ('operatorCancel', 'scoreThresholdReached', 'walletScoringError', 'ciphertraceError'))
```
siiky: If Ciphertrace went down the drain maybe create a migration to `ciphertraceError -> walletScoringError` ?

lib/plugins/wallet-scoring/scorechain/scorechain.js
```diff
@@ -0,0 +1,74 @@
...
+  USDT_TRON: 'TRON'
+}
+
+const TYPE = {
+  TRANSACTION: 'TRANSACTION',
+  ADDRESS: 'ADDRESS'
+}
+
+function rate (account, objectType, cryptoCode, objectId) {
+  if (_.isNil(account) || !account.enabled || !Object.keys(SUPPORTED_COINS).includes(cryptoCode)) return Promise.resolve(null)
```
siiky: This could be replaced by `isWalletScoringEnabled(account, cryptoCode)`

lib/plugins/wallet-scoring/scorechain/scorechain.js
```diff
@@ -0,0 +1,74 @@
...
+
+  return axios.post(`https://api.scorechain.com/v1/scoringAnalysis`, payload, { headers: { 'X-API-KEY': account.apiKey }
+  })
+    .then(res => {
+      const resScore = res.data?.analysis?.assigned?.result?.score
+      if (!resScore) throw new Error('Failed to get score from Scorechain API')
+
+      // normalize score to 0-10 where 0 is the highest risk
+      // use 101 instead of 100 to avoid division by zero
+      return { score: (101 - resScore) / 10 - 0.1, isValid: resScore >= threshold }
```
siiky: Why is 0 the highest risk? I seem to remember 0 being the lowest and 10 the highest with Ciphertrace, and that being colorcoded as green-yellow-red accordingly in l-a; however I don't see changes related to that in this PR.

Also, there's no division by zero possible..? Methinks `score: resScore/10` is fine (or `score: (100-resScore)/10` for the inverse order).

siiky (changes requested):

lib/tx.js
RafaelTaranto: Good point, done.

lib/plugins/wallet-scoring/scorechain/scorechain.js
RafaelTaranto: Nice find.

lib/plugins/wallet-scoring/scorechain/scorechain.js
RafaelTaranto: Idk why I decided to prevent division by zero, maybe I had it the other way around on my mind.

The other one is a mistake on the comment, this normalizes to 0 being lowest risk, which was previously used.

lib/plugins/wallet-scoring/scorechain/scorechain.js
```diff
@@ -0,0 +1,77 @@
...
+  USDT_TRON: 'TRON'
+}
+
+const TYPE = {
+  TRANSACTION: 'TRANSACTION',
+  ADDRESS: 'ADDRESS'
+}
+
+function rate (account, objectType, cryptoCode, objectId) {
+  if (!isWalletScoringEnabled(account, cryptoCode)) return Promise.resolve(null)
```
siiky: `isWalletScoreingEnabled()` returns a promise :/ I could have avoid this by being more explicit...

RafaelTaranto: nah, it's my bad. 

I don't think something called `isWalletScoringEnabled` should be a promise, but it's a bigger change for that one.

siiky (approved):

PR #1681, opened 2024-05-28
LAM-1103 feat: camera diagnostics

RafaelTaranto:


lib/machine-loader.js
```diff
@@ -436,6 +477,44 @@ function getNetworkHeartbeatByDevice (deviceId) {
...
+  }
+
+  const decodedImageData = Buffer.from(data, 'base64')
+  const filename = path.join(dir, name)
+  return fsPromises.writeFile(filename, decodedImageData)
+}
+
+function updatePhotos (scan, front, dir) {
+  const dirname = path.join(dir)
+  _.attempt(() => makeDir.sync(dirname))
```
siiky, line 508: `_.attempt()` will ignore the error if `makeDir` fails, which will then cause `writeFile()` in `createPhoto` to fail too.

PR #1713, opened 2024-08-19
LAM-1145 feat: elliptic wallet scoring

RafaelTaranto:


lib/new-settings-loader.js
```diff
@@ -63,6 +63,16 @@ function saveAccounts (accounts) {
...
       const newAccounts = _.merge(currentAccounts, accounts)
+
+      // Only allow one wallet scoring active at a time
+      if (accounts.elliptic?.enabled && newAccounts.scorechain) {
+        newAccounts.scorechain.enabled = false
+      }
+
+      if (accounts.scorechain?.enabled && newAccounts.elliptic) {
+        newAccounts.elliptic.enabled = false
+      }
```
siiky, line 74: This is dubious but maybe it's ok, not sure...

If the OP has `scorechain` enabled, and then enables `elliptic`, `scorechain` will be disabled. If the OP then tries to enable `scorechain`, `elliptic` will remain enabled, and `scorechain` will remain disabled.

siiky (approved):

lib/new-settings-loader.js
RafaelTaranto: accounts only comes with the different account, the name is misleading in this scenario. In a third party update either scorechain or elliptic will be present, not both.

PR #1721, opened 2024-09-12
LAM-1163 Avoid unnecessary version bumps in l-s

siiky:


lib/graphql/types.js
```diff
@@ -29,7 +29,6 @@ type OperatorInfo {
 }
 
 type MachineInfo {
-  deviceId: String!
```
RafaelTaranto, line 32: Removing this will break backwards compatibility. Add a @deprecated here.

RafaelTaranto (changes requested):

RafaelTaranto (approved):

PR #1767, opened 2024-11-29
LAM-628 LAM-629 backport: blacklist message

RafaelTaranto:


lib/blacklist.js
```diff
@@ -1,48 +1,49 @@
+const _ = require('lodash/fp')
```
siiky, line 1: unused..?

PR #1831, opened 2025-04-22
LAM-1396 chore: update mui

RafaelTaranto:


new-lamassu-admin/src/components/inputs/base/Autocomplete.jsx
```diff
@@ -114,27 +114,30 @@ const Autocomplete = ({
           <Box
             width={18}
             height={18}
-            borderRadius={6}
+            borderRadius="6px"
             bgcolor={warningColors[props.warning]}
           />
         )
 
+        console.log('props.warningMessage: ', props)
```
siiky: Debugging remnants?

new-lamassu-admin/src/components/Popper.jsx
```diff
@@ -120,28 +122,42 @@ const Popover = ({
...
+      name: 'computeStyles',
+      options: {
+        gpuAcceleration: false
+      }
     }
-  })
+  ]
 
   if (props.preventOverflow === false) {
     modifiers.hide = {
```
siiky: This should be changed to a `modifiers.push()` or something?

new-lamassu-admin/src/styling/theme.js
```diff
@@ -41,121 +37,221 @@ export default createTheme({
       default: backgroundColor
     }
   },
-  overrides: {
-    MuiRadio: {
-      colorSecondary: {
-        color: secondaryColor
+})
+
+theme = createTheme(theme, {
```
siiky, line 42: If I understood [the docs](https://mui.com/material-ui/customization/theming/#createtheme-options-args-theme) correctly, this reassignment could be avoided by simply removing these three lines?

siiky (changes requested):

new-lamassu-admin/src/styling/theme.js
RafaelTaranto: The problem here is using theme configuration in the object that I'm re-assigning. I can't reference the theme unless I create a theme first, this is exactly like the docs suggest handling it under "theme composition"

new-lamassu-admin/src/components/Popper.jsx
RafaelTaranto: good catch, removed the option since it was never used anyways.

PR #1870, opened 2025-05-27
LAM-1432 feat: reject address per customer

RafaelTaranto:


packages/server/lib/cash-in/cash-in-tx.js
```diff
@@ -194,14 +195,24 @@ function postProcess(r, pi, isBlacklisted, addressReuse, walletScore) {
...
-  return db.one(sql, [tx.id, tx.toAddress]).then(({ exists }) => exists)
+    SELECT COUNT(*) > 1 as exists
+    FROM (SELECT DISTINCT customer_id
+          FROM cash_in_txs
+          WHERE to_address = $1
+            AND customer_id != $3
+            AND fiat > 0
+          UNION
+          SELECT $2
+          WHERE $2 != $3) t;
```
siiky, line 214: Wouldn't it be enough to find transactions from `customer_id != $2 AND customer_id != $3` and then `COUNT(*) > 0`?

siiky (changes requested):

PR #1875, opened 2025-05-29
LAM-1442 fix: reload table when customer added

RafaelTaranto:


packages/admin-ui/src/pages/Customers/Customers.jsx
```diff
@@ -120,6 +111,8 @@ const Customers = () => {
 
   const byAuthorized = c => (c.authorizedStatus.label === 'Pending' ? 0 : 1)
   const byLastActive = c => new Date(R.prop('lastActive', c) ?? '0')
+
+  const customers = R.path(['customers'])(customersResponse) ?? []
```
siiky: `?? []` is unnecessary in this statement or the following.

packages/admin-ui/src/pages/Customers/Customers.jsx
```diff
@@ -86,23 +86,14 @@ const Customers = () => {
...
-  const [customers, setCustomers] = useState([])
   const [showCreationModal, setShowCreationModal] = useState(false)
 
-  const { data: customersResponse, loading: customerLoading } = useQuery(
-    GET_CUSTOMERS,
-    {
-      onCompleted: data => setCustomers(R.path(['customers'])(data)),
-    },
-  )
+  const { data: customersResponse, loading: customerLoading } =
```
siiky: `s/customerLoading/customersLoading/` for consistency.

siiky (approved):

packages/admin-ui/src/pages/Customers/Customers.jsx
RafaelTaranto: One of the two has to do it. R.map is not null safe.

PR #1877, opened 2025-06-03
LAM-1291 feat: stress testing scripts and some performance improvements

siiky:


packages/server/lib/middlewares/populateDeviceId.js
```diff
@@ -12,7 +14,9 @@ const populateDeviceId = function (req, res, next) {
   const peerCert = req.socket.getPeerCertificate
     ? req.socket.getPeerCertificate()
     : null
-  const deviceId = peerCert?.raw ? sha256(peerCert.raw) : null
+  let deviceId = peerCert?.raw ? sha256(peerCert.raw) : null
+
+  if (!deviceId && IS_STRESS_TESTING) deviceId = req.headers.device_id
```
RafaelTaranto: I know this is far from being ideal, but i would be less worried if we had this commented out by default.

PR #1880, opened 2025-06-10
LAM-1445 feat: model compliance triggers in relational terms

siiky:


packages/server/lib/graphql/resolvers.js
```diff
@@ -74,28 +74,36 @@ const addMachineScreenOpts = smth =>
...
-      if (customTriggers[index].customInfoRequestId !== details.id) return
-      customTriggers[index] = {
-        ...customTriggers[index],
-        customInfoRequest: details,
-      }
-    })
-    return [...normalTriggers, ...customTriggers]
+  allTriggers = allTriggers.map(t =>
+    Object.assign(t, { requirement: t.requirementType }),
+  ) // NOTE: backwards compatibility
```
RafaelTaranto: Make the comment the same as the other backwards compatibility ones, example:

```js
      // BACKWARDS_COMPATIBILITY 10.1
      // bills before 10.2 don't have fixedFee
      if (key === 'fixedFee' && !value) return new BN(0).toString()
      return value.toString()
```

packages/server/migrations/1749551637988-relational-compliance-triggers.js
```diff
@@ -0,0 +1,121 @@
...
+  'external',
+  'custom'
+)
+`
+
+const CREATE_TABLE = `
+CREATE TABLE compliance_triggers (
+  id UUID PRIMARY KEY,
+
+  direction TEXT NOT NULL,
```
RafaelTaranto: why no enum for direction?

packages/server/migrations/settings.js
```diff
@@ -0,0 +1,108 @@
...
+const NEW_SETTINGS_LOADER_SCHEMA_VERSION = 2
+
+const insertConfigRow = (dbOrTx, data) =>
+  dbOrTx.none(
+    "INSERT INTO user_config (type, data, valid, schema_version) VALUES ('config', $1, TRUE, $2)",
+    [data, NEW_SETTINGS_LOADER_SCHEMA_VERSION],
+  )
+
+const notifyReload = (dbOrTx, operatorId) =>
+  dbOrTx.none('NOTIFY $1:name, $2', ['reload', JSON.stringify({ operatorId })])
```
RafaelTaranto: migrations run while the server is down, what are we trying to notify?

packages/typesafe-db/src/user-config.ts
```diff
@@ -0,0 +1,147 @@
...
+
+import db, { inTransaction, DBOrTx } from './db.js'
+import { UserConfig } from './types/types.js'
+
+type UserConfigInsert = Insertable<UserConfig>
+
+const NEW_SETTINGS_LOADER_SCHEMA_VERSION = 2
+
+export async function notifyReload(dbOrTx: DBOrTx, operatorId: string) {
+  // TODO: is this correct?
```
RafaelTaranto: You can test this by adding logs on the event handler.

RafaelTaranto (changes requested):

packages/typesafe-db/src/user-config.ts
siiky: I can confirm the notification is emitted and handled alright, what I don't know is if that `sql.literal` use is correct.

RafaelTaranto (approved):

PR #1889, opened 2025-06-19
LAM-1439 feat: txs new table

RafaelTaranto:


packages/server/lib/new-admin/graphql/types/transaction.type.js
```diff
@@ -83,6 +87,7 @@ const typeDef = gql`
       txClass: String
       deviceId: String
       customerName: String
+      customerId: String
```
siiky: Shouldn't this one be `ID` as well?

packages/typesafe-db/src/interpolled-query-logger.ts
siiky, line 1: `s/inpterpolled/interpolated/`

packages/admin-ui/src/pages/Transactions/Transactions.jsx
```diff
@@ -126,203 +153,294 @@ const GET_TRANSACTIONS = gql`
...
+    variables,
+    notifyOnNetworkStatusChange: true,
+  })
+
+  const [searchCustomersQuery] = useLazyQuery(SEARCH_CUSTOMERS)
+
+  const displayData = useMemo(() => {
+    const formattedData = (data?.transactions ?? []).map(row => ({
+      ...row,
+      toAddress: formatCryptoAddress(row.cryptoCode, row.toAddress),
```
siiky, line 198: [`formatAddress()`](https://github.com/lamassu/lamassu-server/blob/e1855c26f67eefe0d11e564b66e612fe8e6190db/packages/admin-ui/src/utils/string.js#L13) shortens lengthy addresses, which may be appropriate here?

packages/admin-ui/src/pages/Transactions/Transactions.jsx
```diff
@@ -126,203 +153,294 @@ const GET_TRANSACTIONS = gql`
...
-        }`,
+    columns: columns,
+    rowCount: displayData?.[0]?.paginationStats?.totalCount ?? 0,
+    getRowId: it => it.id,
+    data: displayData || [],
+    manualFiltering: true,
+    manualPagination: true,
+    enableSorting: false,
+    onPaginationChange: it => {
+      console.log('PAGINATION', it)
```
siiky: debugging leftovers?

packages/typesafe-db/src/customers.ts
```diff
@@ -122,63 +119,98 @@ function getCustomerList(
...
         ).as('customInfoRequestData'),
       ),
     )
     .orderBy('lastActive', 'desc')
     .execute()
 }
 
-export { getCustomerList }
+function searchCustomers(searchTerm: string, limit: number = 20): Promise<any> {
+  const searchPattern = `%${searchTerm}%`
```
siiky, line 182: Would be better if this used [Kysely's `sql`](https://kysely-org.github.io/kysely-apidoc/interfaces/Sql.html):

```ts
sql`%${searchTerm}%`
```

packages/typesafe-db/src/customers.ts
```diff
@@ -122,63 +119,98 @@ function getCustomerList(
...
+        .leftJoin('editedCustomerData as cstED', 'cstED.customerId', 'cst.id')
+        .select(({ eb, fn }) => [
+          'cst.id',
+          'cst.phone',
+          'cst.email',
+          sql`CONCAT(
+            COALESCE(${selectNewestIdCardData(eb)}->>'firstName', ''),
+            ' ',
+            COALESCE(${selectNewestIdCardData(eb)}->>'lastName', '')
+          )`.as('customerName'),
```
siiky: Maybe we should stop doing this in SQL, we already have the code in the UI ("John Doe" -> "J. Doe"): https://github.com/lamassu/lamassu-server/blob/82abf24c426a76bd54d9a6742a716b178537d6c7/packages/admin-ui/src/utils/customer.js#L14-L24

I don't see a `JOIN()` or `FILTER()` or similar that could help make this correct: https://www.postgresql.org/docs/12/functions-string.html

packages/server/lib/new-admin/services/transactions.js
```diff
@@ -1,230 +1,72 @@
...
-  const packager = _.flow(
-    _.flatten,
-    _.orderBy(_.property('created'), ['desc']),
-    _.map(
-      _.flow(
-        camelize,
-        _.mapKeys(k => (k == 'cashInFee' ? 'fixedFee' : k)),
+  return (
+    Promise.all([
+      getTransactionList(
```
siiky: Why not ditch the `Promise.all()` if there's a single promise?

packages/typesafe-db/src/transactions.ts
```diff
@@ -0,0 +1,373 @@
...
+        'paginationStats',
+      ),
+    )
+    .orderBy('transactions.created', 'desc')
+
+  if (filters.toAddress) {
+    query = query.where(
+      'transactions.toAddress',
+      'like',
+      `%${filters.toAddress}%`,
```
siiky, line 318: similar to the other comment

siiky (approved):

packages/admin-ui/src/pages/Transactions/Transactions.jsx
RafaelTaranto: That's not needed here since the columns are resizable.

packages/typesafe-db/src/customers.ts
RafaelTaranto: Added a trim. We should aim to have a simpler front-end by hiding away implementation details of the data on the server.

PR #1911, opened 2025-07-11
LAM-1478 feat: machine groups

RafaelTaranto:


packages/typesafe-db/src/machines.ts
```diff
@@ -0,0 +1,12 @@
+import db from './db.js'
+
+export async function assignMachinesToGroup(
+  deviceIds: [string],
+  groupId: string,
+) {
+  return db
```
siiky, line 7: since the function is `async` doesn't this have to be awaited on? or remove the `async` qualifier. I think this is what bit me once, using `db.task()`...

packages/typesafe-db/src/machine-groups.ts
```diff
@@ -0,0 +1,34 @@
+import db from './db.js'
+
+export async function createMachineGroup(data: { id: string; name: string }) {
```
siiky, line 3: same as the other comment; if we don't `await` DB operations the functions don't have to be `async`.

packages/server/lib/machine-loader.js
```diff
@@ -26,7 +29,8 @@ const bootingUpStatus = { label: 'Booting up', type: 'warning' }
 const OPERATOR_DATA_DIR = process.env.OPERATOR_DATA_DIR
 
 const MACHINE_WITH_CALCULATED_FIELD_SQL = `
-select d.*, COALESCE(emptybills, 0) + COALESCE(regularbills, 0) as cashbox from devices d
+select d.*, mg.id as machine_group_id, mg.name as machine_group_name, COALESCE(emptybills, 0) + COALESCE(regularbills, 0) as cashbox from devices d
```
siiky, line 32: `d.*` already includes `machine_group_id`..?

PR #1920, opened 2025-07-17
LAM-1488 feat: compliance trigger sets

siiky:


packages/typesafe-db/src/compliance-triggers.ts
```diff
@@ -5,11 +5,15 @@ import { ComplianceTriggers } from './types/types.js'
 
 type ComplianceTriggerInsert = Insertable<ComplianceTriggers>
 
+/*
+ * Deprecated API
+ */
+
```
RafaelTaranto, line 11: Can we remove if deprecated?

packages/typesafe-db/src/compliance-triggers.ts
```diff
@@ -18,3 +22,92 @@ export function saveComplianceTriggers(
...
+export function deleteComplianceTriggerSet(dbOrTx: DBOrTx, id: string) {
+  return inTransaction(dbOrTx, async tx => {
+    await deleteComplianceTriggersBySetId(tx, id)
+    return await tx
+      .deleteFrom('complianceTriggerSets')
+      .where('id', '=', id)
+      .returningAll()
+      .executeTakeFirstOrThrow()
+  })
+}
```
RafaelTaranto: Not needed with cascade.

packages/server/migrations/1752141860742-compliance-triggers-sets.js
```diff
@@ -0,0 +1,46 @@
...
+  name TEXT NOT NULL UNIQUE
+)
+`
+
+const INSERT_DEFAULT =
+  "INSERT INTO compliance_trigger_sets (id, name) VALUES ($1, 'Default')"
+
+const ADD_SET_COLUMN = `
+ALTER TABLE compliance_triggers
+ADD COLUMN compliance_trigger_set UUID REFERENCES compliance_trigger_sets (id)
```
RafaelTaranto: This should be compliance_trigger_set_id and should add the cascade delete.

packages/typesafe-db/src/compliance-triggers.ts
```diff
@@ -18,3 +22,92 @@ export function saveComplianceTriggers(
...
+
+export function deleteComplianceTriggersBySetId(
+  dbOrTx: DBOrTx,
+  complianceTriggerSet: string,
+) {
+  return dbOrTx
+    .deleteFrom('complianceTriggers')
+    .where('complianceTriggerSet', '=', complianceTriggerSet)
+    .execute()
+}
```
RafaelTaranto: Not needed with cascade delete.

packages/typesafe-db/src/compliance-triggers.ts
```diff
@@ -18,3 +22,92 @@ export function saveComplianceTriggers(
...
+
+export function deleteComplianceTriggersByCustomInfoRequestId(
+  dbOrTx: DBOrTx,
+  customInfoRequestId: string,
+) {
+  return dbOrTx
+    .deleteFrom('complianceTriggers')
+    .where('customInfoRequestId', '=', customInfoRequestId)
+    .execute()
+}
```
RafaelTaranto: I think this was the behavior but it's a bit strange, shouldn't we prevent this hidden delete?

packages/server/lib/new-settings-loader.js
```diff
@@ -100,25 +97,14 @@ function showAccounts(schemaVersion) {
...
+      inTransaction(db, async tx => {
         const currentConfig = await _loadConfigTx(tx)
         const newConfig = addTermsHash(_.assign(currentConfig, config))
-        const triggers = newConfig.triggers.map(
-          renameKeys({ requirementType: 'requirement' }),
-        )
         delete newConfig.triggers
-        await saveComplianceTriggers(tx, triggers)
-        await userConfig.insertConfigRow(tx, { config: newConfig })
+        await userConfig.insertConfigRow(tx, { config })
```
RafaelTaranto: why are we not using newConfig?

packages/server/lib/compliance-triggers.js
```diff
@@ -1,56 +1,82 @@
-const _ = require('lodash/fp')
```
RafaelTaranto, line 1: All of the lodash related changes are not null safe anymore, are you sure no usage breaks here?

packages/admin-ui/src/pages/Triggers/TriggerSets.jsx
```diff
@@ -0,0 +1,334 @@
...
+  const [deleteComplianceTriggerSet] = useMutation(
+    DELETE_COMPLIANCE_TRIGGER_SET,
+    {
+      refetchQueries: ['getComplianceTriggerSets'],
+      onError: ({ graphQLErrors }) => {
+        const errorCode = graphQLErrors?.[0]?.extensions?.code
+
+        switch (errorCode) {
+          case 'RESOURCE_NOT_FOUND':
+            setErrorMsg('Compliance trigger set not found')
```
RafaelTaranto: I don't remember seeing this on the service.

RafaelTaranto (changes requested):

packages/typesafe-db/src/compliance-triggers.ts
siiky: It's mentioned in the confirmation screen:

image

packages/server/lib/new-settings-loader.js
siiky: nice catch, must have screwed this up during rebases...

packages/server/lib/compliance-triggers.js
siiky: AFAICS only the `triggers` may be something other than an array and thus cause an error, in which case that's a bug that we need to fix somewhere else (all the uses as of now take the triggers out of the a `settings` object, result of `settingsLoader.load()`). I think it would be worse to have hidden errors.

packages/typesafe-db/src/compliance-triggers.ts
siiky: `saveAllComplianceTriggers` was leftovers, removed. `getAllComplianceTriggers` is used in `settingsLoader.load()` (which still includes all the triggers). We can remove it once the machine groups and compliance trigger sets features work together.

RafaelTaranto (approved):

PR #1929, opened 2025-07-28
LAM-1500 feat: marry machine groups with compliance trigger sets

siiky:


packages/admin-ui/src/pages/Triggers/TriggerSets.jsx
```diff
@@ -191,7 +197,7 @@ const TriggerSets = () => {
           setTriggerSetToDelete(row.original)
           setDeleteDialogOpen(true)
           setErrorMsg('')
-          setActionMenuClose(() => closeMenu)
+          setActionMenuClose(closeMenu)
```
RafaelTaranto, line 200: you can remove all of the state for the action menu close and just close it on click.

packages/server/migrations/1753351586476-machine-groups-with-compliance-trigger-sets.js
```diff
@@ -0,0 +1,12 @@
+const db = require('./db')
+
+const ALTER_TABLE = `
+ALTER TABLE machine_groups
+ADD COLUMN compliance_trigger_set_id UUID
+  REFERENCES compliance_trigger_sets (id)
+  ON DELETE SET NULL
+`
```
RafaelTaranto, line 8: can't we add the compliance_trigger_set to the group if it exists? The migration should leave the server fully functional without manual intervention

RafaelTaranto (changes requested):

PR #1943, opened 2025-08-28
LAM-1505 feat: configurable sweeping

siiky:


packages/admin-ui/src/pages/Wallet/Wizard.jsx
```diff
@@ -48,7 +48,7 @@ const Wizard = ({
 }) => {
   const [{ step, config, accountsToSave }, setState] = useState({
     step: 0,
-    config: { active: true },
```
RafaelTaranto, line 51: why is `active: true` no longer needed?

packages/server/lib/new-admin/config/accounts.js
```diff
@@ -18,6 +18,9 @@ const EMAIL = 'email'
 const ZERO_CONF = 'zeroConf'
 const WALLET_SCORING = 'wallet_scoring'
 const COMPLIANCE = 'compliance'
+const SWEEP_STRATEGY = 'sweepStrategy'
+
+const SWEEPABLE_COINS = ['ETH', 'TRX', 'USDC', 'USDT', 'USDT_TRON']
```
RafaelTaranto: USDC, USDT and USDT_TRON are not sweepable.

RafaelTaranto (changes requested):
Also rebase it.

packages/admin-ui/src/pages/Wallet/Wizard.jsx
siiky: https://github.com/lamassu/lamassu-server/pull/1943/commits/fec20b3b604c79e5fe875c1cf2a0db0deb408582

as far as I could tell, it's unused

packages/server/lib/new-admin/config/accounts.js
siiky: fixed and added `SOL`

RafaelTaranto (approved):

PR #1952, opened 2025-10-15
LAM-1504 feat: add Solana support

siiky:


packages/server/lib/plugins/wallet/solana/base.js
```diff
@@ -0,0 +1,324 @@
...
+const client = {}
+
+// From [SLIP-0010](https://github.com/satoshilabs/slips/blob/master/slip-0010.md#child-key-derivation-ckd-functions):
+// > For ed25519 and curve25519 only hardened key generation from private
+// > parent key to private child key is supported.
+//
+// The payment prefix is used for cashouts, while the default prefix is used
+// for funding and cashins.
+const paymentPrefixPath = "m/44'/501'/0'/0'"
+const defaultPath = "m/44'/501'/1'/0'/0'"
```
RafaelTaranto: There's an extra `0'` on default path here.

packages/server/lib/plugins/wallet/solana/base.js
```diff
@@ -0,0 +1,324 @@
...
+// @returns A signed transaction, ready to send
+// @see https://www.solanakit.com/docs/concepts/transactions
+const createTransaction = async ({
+  sender,
+  toAddress,
+  cryptoAtoms,
+  subtractFees,
+  feeMultiplier,
+}) => {
+  feeMultiplier ??= 1
```
RafaelTaranto: why not do the simpler `feeMultiplier = 1` on the param declaration? It prevents the 'smell' of reassigning params.

packages/server/lib/plugins/wallet/solana/base.js
```diff
@@ -0,0 +1,324 @@
...
+    .times(feeMultiplier)
+    .toFixed(0)
+
+  // Create a mock transaction to estimate compute units limit
+  const transaction = createBaseTransaction({
+    sender,
+    toAddress,
+    cryptoAtoms,
+    latestBlockHash,
+    computeUnitLimit: 450, // typical for 1 signer, local devnet, with CU limit, and priority fee
```
RafaelTaranto: Add a const for this magical number.

packages/server/lib/plugins/wallet/solana/base.js
```diff
@@ -0,0 +1,324 @@
...
+    priorityPrice,
+  })
+  const computeUnitLimit = await client.estimateComputeUnitLimit(transaction)
+
+  // Used for sweeping where cryptoAtoms is the total balance of the sending
+  // account. If fees are greater than available balance, abort the
+  // transaction.
+  if (subtractFees) {
+    const fees = BASE_FEE.plus(
+      BN(computeUnitLimit).times(priorityPrice).div(1000000), // priorityPrice is in micro-Lamports (10^-6)
```
RafaelTaranto: Add a const for micro-lamport scale as well.

RafaelTaranto (changes requested):
This will also need to be added to the sweep strategy list of the previous PR

packages/server/lib/plugins/wallet/solana/base.js
siiky: That's the [full path](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#path-levels) to the default wallet, not the prefix, because we don't use other indexes. The Ethereum wallet could do the same, don't know why it doesn't.

siiky: wtf github...

That's the [full path](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#path-levels) to the default wallet, not the prefix, because that's the only used index. The Ethereum wallet could do the same, I don't know why it doesn't.

RafaelTaranto (approved):

PR #1956, opened 2025-10-22
LAM-1537 feat: custom ticker

siiky:


packages/server/lib/plugins/ticker/custom.js
```diff
@@ -0,0 +1,27 @@
...
+    url: endpoint + '/get-rates',
+    params: {
+      fiat,
+      crypto,
+    },
+    headers: {
+      'x-lamassu-api-key': apiKey,
+    },
+  }).then(({ data, status }) => {
+    if (status !== 200) logger.warn('unexpected status code')
```
RafaelTaranto: isn't this too open? if it's not 200 throw an error, no?

RafaelTaranto (changes requested):

packages/server/lib/plugins/ticker/custom.js
siiky: Axios only resolves for 2xx HTTP codes (200 <= status < 300), for others (4xx, 5xx, ...) it throws an error (see `validateStatus` in the Axios docs). Maybe it should be stricter still and throw an error on 2xx other than 200 too?

RafaelTaranto: Yeah, make it 200 only.

RafaelTaranto (approved):

PR #1959, opened 2025-11-14
LAM-1546 feat: add `walletId` to `cash_out_txs` table

siiky:


siiky:
Not worth keeping separate from #1960 as it turned out not that big.

PR #1960, opened 2025-11-14
LAM-1546 feat: custom wallet

siiky:


packages/server/migrations/1763395869093-crypto-hdindex.js
```diff
@@ -0,0 +1,52 @@
...
+
+      return maxHdIndex
+    })
+    .then(maxHdIndex =>
+      db
+        .runAll([DROP_HDIDX_SEQ, CREATE_HDIDX_SEQ_TABLE])
+        .then(() => insertHDIndexes(maxHdIndex)),
+    )
+    .then(next)
+
```
RafaelTaranto: This can be greatly simplified by just querying this:

```sql
select crypto_code, coalesce(max(hd_index), 0)
from cash_out_txs
group by crypto_code;
```

then insert that into the new table and no js is required for this migration

packages/server/migrations/1763395869093-crypto-hdindex.js
```diff
@@ -0,0 +1,52 @@
+const db = require('./db')
+const { loadConfig } = require('./settings')
+
+const GET_MAX_CRYPT_HDINDEX = `
+SELECT crypto_code, MAX(hd_index) AS hd_index
```
RafaelTaranto:
```suggestion
SELECT crypto_code, COALESCE(MAX(hd_index), 0) AS hd_index
```

packages/server/migrations/1763395869093-crypto-hdindex.js
```diff
@@ -0,0 +1,52 @@
...
+      })
+
+      return maxHdIndex
+    })
+    .then(maxHdIndex =>
+      db
+        .runAll([DROP_HDIDX_SEQ, CREATE_HDIDX_SEQ_TABLE])
+        .then(() => insertHDIndexes(maxHdIndex)),
+    )
+    .then(next)
```
RafaelTaranto: I don't think JS is necessary. You can just ignore whatever coin is configured and just use the query with coalesce to insert straight into the new sequence table.

packages/server/lib/wallet.js
```diff
@@ -126,16 +126,17 @@ function sendCoinsBatch(settings, txs, cryptoCode) {
...
 }
 
-function newAddress(settings, info, tx) {
-  const walletAddressPromise = fetchWallet(settings, info.cryptoCode).then(r =>
-    r.wallet.newAddress(r.account, info, tx),
-  )
+function newAddress(settings, info) {
+  const walletAddressPromise = fetchWallet(settings, info.cryptoCode)
+    .then(r => r.wallet.newAddress(r.account, info))
+    .then(res => (typeof res === 'string' ? { address: res } : res))
```
RafaelTaranto: if we want to return an object, every wallet should return an object. Ideally return the wallet ID for the wallets we can generate the ID from.

packages/server/lib/plugins/wallet/custom-wallet/api.js
```diff
@@ -0,0 +1,97 @@
+const axios = require('axios')
+
+const request = ({ endpoint, apiKey }, { method, path, params }) =>
+  axios({
+    method,
+    url: [endpoint].concat(path).join('/'), // TODO: ensure proper URL
```
RafaelTaranto: TODO left

packages/server/lib/plugins/wallet/custom-wallet/custom-wallet.js
```diff
@@ -0,0 +1,151 @@
...
+    .then(() =>
+      api.transactionSend(account, {
+        crypto: tx.cryptoCode,
+        toAddress: tx.toAddress,
+        amountFullCryptoCoins: convertAmount(
+          tx.cryptoAtoms,
+          tx.cryptoCode,
+          'cryptoAtoms',
+          'fullCoins',
+        ).toFixed(0), // TODO
```
RafaelTaranto: another TODO left

RafaelTaranto (changes requested):

packages/server/migrations/1763395869093-crypto-hdindex.js
siiky: That's true, a lot less code and it's clearer too. Haven't tested that the migration still works.

RafaelTaranto (approved):

